PR #91, opened 2016-11-15
Break up the AWS template into modules

maratoid:
Break up the AWS template into modules. 
Make resourcePrefix a top - level option.
Remove debug code


coffeepac:
LGTM

Issue #118, opened 2016-12-08
etcd nodes get recreated with call of up.sh


coffeepac:
etcd nodes are being destroyed/rebuild each time that up.sh gets called.  This is deleterious behavior for people who are attempting to update part of a cluster config.  Or patch a botched create.

coffeepac:
the user_data field is being regenerated each time.  and its some kind of hash and the hash is changing.  even if the data isn't.  huh.

coffeepac:
the cloud-configs are being regenerated each time.  this should only happen if the the inputs have changed.

coffeepac:
Check if content of file is actually changing.  Look for a field with a timestamp.  

Failing that, figure out why hash is changing in tf state file.

coffeepac:
the SSL certs are changing.  the SSL certs are changing for all of the cluster pools, actually.  waiting for the current up to finish to see if the ASGs get updated or not.

coffeepac:
the SSL certs are not changing.  but 'cat ca.pem | gzip  | base64' is not idempotent.  its the gzip portion, the compression tree is subtly changing each time.  hooray.

coffeepac:
I'm going to remove the compression portion of this and only leave the base64 encoding.  this is an understood format by cloud-init and will be idempotent.

coffeepac:
fixed by https://github.com/samsung-cnct/k2/issues/120

PR #120, opened 2016-12-08
make gz calls idemptotent

coffeepac:
force all calls to gzip to happen an arbitrary, fixed time

### DIFF
--- a/ansible/roles/kraken.config/filter_plugins/gz.py
+++ b/ansible/roles/kraken.config/filter_plugins/gz.py
@@ -1,9 +1,10 @@
 import StringIO
 import gzip
+import time
 
 def gz(content_to_compress):
  	out = StringIO.StringIO()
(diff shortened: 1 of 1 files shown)


ansible/roles/kraken.config/filter_plugins/gz.py
```diff
@@ -1,9 +1,10 @@
 import StringIO
 import gzip
+import time
```
davidewatson: Are you using time?

davidewatson:
LGTM, modulo my one question.

ansible/roles/kraken.config/filter_plugins/gz.py
coffeepac: nope.  thought I was going to need it for setting a consistent time for the gzip but I just hardcoded it instead.  I'll remove it, good catch.

coffeepac:
fixes #118

Issue #95, opened 2016-11-17
More Specific New User How To


mikeln:
There should be an obvious step-by-step first time usage document that should be the first thing listed in the README.

That doc should spell out exactly things that need to be done, values that need to be set to something meaningful for the user, etc... (e.g. Set your cluster name - here, use that same name - here, look in AWS (or whatever) for that name as a prefix ...

coffeepac:
@mikeln is the Getting Started guide in k2cli sufficient?  https://github.com/samsung-cnct/k2cli

Probably need to add aws account requirements, but other then that?

mikeln:
yeah, that looks better.

coffeepac:
thanks!  opened up https://github.com/samsung-cnct/k2cli/issues/26 to specify AWS cred requirements.

Issue #64, opened 2016-10-21
Kraken replacement?


cgswong:
Two very interesting projects here with Kraken and K2 (Kraken 2), with some awesome work being done. It seems that K2 is the replacement for the original Kraken based on the title, correct? If so, then the repo to use and follow would be K2, correct?

I'd recommend a brief snippet or two on both repos on which to use, or maybe it is either depending on use case, just to make things clearer. If one (Kraken?) is deprecated I think that should be stated as well.

Thanks

maratoid:
Thank you for your interest, and that is an excellent point. We'll be addressing this (and https://github.com/samsung-cnct/k2/issues/46) ASAP.

coffeepac:
@cgswong sorry for the delay on this!  we have updated the README for k2 and included the whys and wherefores.

if you have any further questions or comments please open an issue and we'll be much more responsive going forward.

Issue #46, opened 2016-10-07
Document motivation to start K2 project after Kraken


somejfn:
Hi,

It would be great if you could share the motivations behind starting K2 and what lead you to start from scratch after Kraken.  

regards,

Jean-Francois

coffeepac:
@somejfn sorry for the long wait on this.  We have updated the README for this project and included the motivations for the rebuild.  

if you have any further questions please open up an issue.  We will be more responsive going forward.

Issue #69, opened 2016-10-26
ssh to intances


keyolk:
I'm not sure whether it is intended working.
and I use the image "quay.io/samsung_cnct/k2:latest"

there is no ssh client in image,
so I use laptop host's ssh client.

problem is ...
  1) owner of RSA key is root, not not mine.
  2) IdentityFile path in ssh_config is under root's home, not my home.

So I just manually edit the ssh_config under .kraken
and try ssh with root account.

is there other convenient way for it ?

below is my host info

```
Linux keyolk-book 4.4.2-6ph #1 SMP Mon Jun 13 20:01:22 KST 2016 x86_64 GNU/Linux
```

```
Containers: 24
 Running: 2
 Paused: 0
 Stopped: 22
Images: 1
Server Version: 1.12.2
Storage Driver: overlay
 Backing Filesystem: xfs
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: null bridge host overlay
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Security Options: apparmor seccomp
Kernel Version: 4.4.2-6ph
Operating System: Apricity OS
OSType: linux
Architecture: x86_64
CPUs: 4
Total Memory: 15.61 GiB
Name: keyolk-book
ID: AKMT:LO4B:TAXU:M6K3:RYEQ:GE5B:GLZ6:ISBQ:WBN7:FYKF:CMN6:TN2W
Docker Root Dir: /home/keyolk/mount/flash/docker
Debug Mode (client): false
Debug Mode (server): false
Registry: https://index.docker.io/v1/
Insecure Registries:
 127.0.0.0/8
```

coffeepac:
@keyolk this is an issue with using the k2 image directly.  the k2 image doesn't have visibility into where the key lives on the host system.  k2cli will perform this update for you but its not quite ready yet.  

so for now you'll need to either edit the file manually or add '-i <path to your ssh key>' each time you try to ssh to a node

coffeepac:
this will be fixed elsewhere.  not tracking here anymore

PR #130, opened 2016-12-16
README updates

coffeepac:
addresses why k2 and to use k2cli first

Explicit issues addressed:
why k2:  https://github.com/samsung-cnct/k2/issues/46
k2 or kraken?  https://github.com/samsung-cnct/k2/issues/64
explicit user quick start:  https://github.com/samsung-cnct/k2/issues/95 (kind of cheating but it points to k2cli)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@ Deploy a __Kubernetes__ cluster on top of __CoreOS__ using __Terraform__  and __
 [![Docker Repository on Quay](https://quay.io/repository/samsung_cnct/k2/status "Docker Repository on Quay")](https://quay.io/repository/samsung_cnct/k2)
 
 ## What is K2
-K2 is an orchestration and cluster level management system for [Kubernetes](https://kubernetes.io).  Its goal is to, by default, create production scale Kubernetes
-clusters on a range of platforms.  It boasts a large configuration file for extensive tweaking for your environment and use case.  Especially if you are getting 
+K2 is an orchestration and cluster level management system for [Kubernetes](https://kubernetes.io).  By default,  K2 will create a production scale Kubernetes
+cluster on a range of platforms.  It boasts a large configuration file for extensive tweaking for your environment and use case.  Especially if you are getting 
 started and don't need a HA production level cluster right away.
 
 We (Samsung CNCT) built this tool to aid in our own research into high performance and reliability for the Kubernetes control plane.  We realised this would be a useful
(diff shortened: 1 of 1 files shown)


mikeln:
lgtm

README.md
```diff
@@ -42,14 +59,14 @@ You will need the following installed on your machine:
...
-The easiest way to get started with K2 is to use a K2 container image
+The easiest way to get started with K2 directly is to use a K2 container image
 
 `docker pull quay.io/samsung_cnct/k2:latest`
 
 ## Preparing the environment  
   
 ### Initial K2 Directory
-If this is your first time using kraken, use k2 docker image to generate a 'sensible defaults' configuration:
+If this is your first time using kraken, use k2 docker image to generate a 'sensible defaults' configuration (this assumes AWS is the infrastructure provider):
```
leahnp: Should line #69 say, "using K2" instead of "kraken"?

coffeepac: yes.  ein moment.

leahnp:
I think a "Who Should Use K2" or "What K2 Can Provide Your Project" section outlining target demographic and problems K2 specifically solves for users could be really useful. I see section this targeting a developer or project manager researching Kubernetes cluster level management options and allows us to clearly define why K2 is a good option.

coffeepac:
@leahnp I will add another 'graph to the 'What is K2' block.  possibly as a sub header.  thanks.

@spiffxp chasing down case mixups.  
For 'what sort of cluster does the default config represent' question, what is insufficient about the second sentence in the first 'graph of the What is K2 section?

README.md
```diff
@@ -3,7 +3,24 @@ Deploy a __Kubernetes__ cluster on top of __CoreOS__ using __Terraform__  and __
 
 [![Docker Repository on Quay](https://quay.io/repository/samsung_cnct/k2/status "Docker Repository on Quay")](https://quay.io/repository/samsung_cnct/k2)
 
+## What is K2
+K2 is an orchestration and cluster level management system for [Kubernetes](https://kubernetes.io).  Its goal is to, by default, create production scale Kubernetes
+clusters on a range of platforms.  It boasts a large configuration file for extensive tweaking for your environment and use case.  Especially if you are getting 
+started and don't need a HA production level cluster right away.
```
spiffxp: @coffeepac this is the phrase that confused me about what the default cluster represents.  Could we change it to say what the default is, and then the variety of places/ways you could tweak it?

spiffxp: To me, this i_implies_ that the default is not an HA production level cluster, but you could tweak it to be as such.  But what is the default to start with?

spiffxp: I reading the sentence above this that use the words production and default in the same sentence.  Its goal is.  But does it.  Maybe just needs some firming.

coffeepac: Firmed it up.

coffeepac:
@leahnp added a 'What is K2 for' section.  @spiffxp do you mind giving that a read as well?  I think you'll find it interesting.

spiffxp:
I was going to make some mocking comment about how there were 9 graphs of text before the actual code, but then I realized that's what this PR is about.  Adding those 9 graphs.  Because the code after it doesn't explain our intent.  And I can't fault that, so LGTM.

leahnp:
LGTM

maratoid:
LGTM

PR #136, opened 2016-12-25
modify service value template. to use multi level yaml values.

keyolk:
to use multi level yaml values to kraken service.


coffeepac:
ok to test

coffeepac:
LGTM

Issue #159, opened 2017-01-09
k2 e2e fails with kubernetes 1.5.1


spiffxp:
I suspect the tests need to be rewritten to account for the fact that client/server binaries aren't included in the main kubernetes.tar.gz release artifact anymore.

eg: https://pipelet.kubeme.io/job/k2-e2e-aws-k8s-conformance/104/console
```
15:47:55 + go run hack/e2e.go --v --test '--test_args=--ginkgo.v=true --ginkgo.noColor=true --ginkgo.focus=\[Conformance\] --ginkgo.skip=\[Serial\] --e2e-output-dir=/var/jenkins_home/jobs/k2-e2e-aws-k8s-conformance/workspace/output/artifacts --report-dir=/var/jenkins_home/jobs/k2-e2e-aws-k8s-conformance/workspace/output/artifacts' --check_version_skew=false
15:47:56 2017/01/06 23:47:56 e2e.go:722: Running: kubectl version
15:47:56 2017/01/06 23:47:56 e2e.go:724: Step 'kubectl version' finished in 273.874µs
15:47:56 2017/01/06 23:47:56 e2e.go:722: Running: get status
15:47:56 ./hack/e2e-internal/e2e-status.sh: line 28: ./hack/e2e-internal/../../cluster/kube-util.sh: No such file or directory
15:47:56 2017/01/06 23:47:56 e2e.go:724: Step 'get status' finished in 2.90292ms
15:47:56 2017/01/06 23:47:56 e2e.go:180: Something went wrong: encountered 2 errors: [error running kubectl version: fork/exec ./cluster/kubectl.sh: no such file or directory error running get status: exit status 1]
15:47:56 exit status 1
```

spiffxp:
1.5.x e2e tests run now; they're still failing, unclear whether that's flakes or a change in expected behavior

https://pipelet.kubeme.io/job/k2-e2e-aws-k8s-conformance/108/ and https://pipelet.kubeme.io/job/k2-e2e-aws-k8s-conformance/109/ have the same e2e failures, but without a few more runs that's not saying much

spiffxp:
Consistently seeing failures of the following e2e tests
```
Kubernetes e2e suite.[k8s.io] Networking [k8s.io] Granular Checks: Pods should function for node-pod communication: http [Conformance]
Kubernetes e2e suite.[k8s.io] Networking [k8s.io] Granular Checks: Pods should function for node-pod communication: udp [Conformance]
[...]

PR #169, opened 2017-01-25
updates to deal with new conformance tests

coffeepac:
please see: https://github.com/samsung-cnct/k2/issues/159 for a full
description

tl;dr there's a bug that means we need to set the vpc domain to be
'ec2.internal' and not override the hostname.  its a combo between
the aws provider and some new e2e tests.


joejulian:
:+1:

coffeepac:
two upvotes and checks passed.  meeeerging.

PR #170, opened 2017-01-26
Prevent k2 from stomping on user files

### LINKED ISSUE #97: K2 Up Creates Subdirectories That May Cause Problems
When the system comes up, it creates files and subdirs in the mounted /root filesystem.

When this mount is $HOME (as given in the README), you have the potential to have unexpected .dot files added to your account, and potentially changed/deleted.

Anecdotal observations:
files created: .rnd, .srl
directories created: .terraform.d
directory created and deleted: .ansible
directory updated: .kube, .kraken

It would be "safer" to contain any changes to a subdirectory.  Or else warn the user what files/dirs will be modified and how, so they can take action to protect whatever they have before k2 does something to them.

joejulian:
Move most temporary configuration paths out of `$HOME` or moved files we want to keep to the `--output` path.

Tested by setting `HOME` to a read-only path with no access errors.

Fixes #97


coffeepac:
LGTM

Issue #128, opened 2016-12-14
aws provider - credentialsFile doesn't work


coffeepac:
using the credentialsFile for authentication doesn't seem to work.

to repro:  supply credentialsFile and credentialsProfile under authentication and attempt to bring up a cluster.  

I am currently getting 

[...]

PR #178, opened 2017-02-03
Use deterministic pauses rather than fixed-length ones.

joejulian:
Rather than using fixed-length pauses, use loops to test if the thing we're pausing for has completed. This allows each of the former pauses to be shorter under many conditions.


joejulian:
Needs #128. Closing for now.

PR #186, opened 2017-02-09
Use deterministic pauses rather than fixed-length ones.

### LINKED ISSUE #129: Services cleanup pauses for 300 seconds even when there are no services to cleanup
In case of an error collecting running services, or when there are no services running the kill-services task still pauses for 300 seconds. For example:

```
TASK [roles/kraken.services : Collect all services] ****************************
fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["kubectl", "--kubeconfig=/Users/marat/.kraken/zonar-production-cluster/admin.kubeconfig", "get", "services", "--all-namespaces", "-o", "json"], "delta": "0:00:00.173786", "end": "2016-12-14 23:26:54.179977", "failed": true, "rc": 1, "start": "2016-12-14 23:26:54.006191", "stderr": "error: stat /Users/marat/.kraken/zonar-production-cluster/admin.kubeconfig: no such file or directory", "stdout": "", "stdout_lines": [], "warnings": []}
...ignoring

TASK [roles/kraken.services : Register services fact] **************************
fatal: [localhost]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value ([u'services', u'services_only', u'all', u'services', u'all', u'services_only', u'services', u'all', u'services_only', u'services', u'all', u'services_only', u'services', u'all', u'services_only', u'services', u'all', u'services_only', u'services', u'all', u'services_only', u'services', u'all', u'services_only', u'services', u'all', u'services_only', u'services', u'all', u'services_only', u'services', u'all', u'services_only']), and could not be converted to an dict. Error was: No JSON object could be decoded\n\nThe error appears to have been in '/kraken/ansible/roles/kraken.services/tasks/kill-services.yaml': line 9, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Register services fact\n  ^ here\n"}
...ignoring

TASK [roles/kraken.services : Set services info] *******************************
fatal: [localhost]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'added_services_map' is undefined\n\nThe error appears to have been in '/kraken/ansible/roles/kraken.services/tasks/kill-services.yaml': line 15, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Set services info\n  ^ here\n"}
...ignoring

TASK [roles/kraken.services : Clean up services] *******************************
fatal: [localhost]: FAILED! => {"failed": true, "msg": "'the_services' is undefined"}
...ignoring

TASK [roles/kraken.services : Pauase to let services come down] ****************
Pausing for 300 seconds
```

joejulian:
Rather than using fixed-length pauses, use loops to test if the thing we're pausing for has completed. This allows each of the former pauses to be shorter under many conditions.

Fixes #129

### DIFF
--- a/ansible/roles/kraken.services/tasks/kill-services.yaml
+++ b/ansible/roles/kraken.services/tasks/kill-services.yaml
@@ -18,17 +18,6 @@
   when: kraken_action == 'down'
   ignore_errors: yes
 
-- name: Clean up services
-  command: >
-    kubectl --kubeconfig={{ kubeconfig }} delete --namespace {{ item.metadata.namespace }} svc {{ item.metadata.name }}
-  with_items: "{{ the_services }}"
-  when: item.status.loadBalancer.ingress[0].hostname is defined and kraken_action == 'down'
-  ignore_errors: yes
-
-- name: Pauase to let services come down
-  pause: minutes=5
-  when: kraken_action == 'down'
-
 - name: Clean up releases
   command: >
     helm delete --purge {{ item.name }}
@@ -44,9 +33,43 @@
     kubectl --kubeconfig={{ kubeconfig }} delete deployment {{ tiller }} --namespace=kube-system
   when: tiller_present|succeeded and not tiller_present|skipped
 
+- name: Clean up services
+  command: >
+    kubectl --kubeconfig={{ kubeconfig }} delete --namespace {{ item.metadata.namespace }} svc {{ item.metadata.name }}
+  with_items: "{{ the_services }}"
+  when: item.status.loadBalancer.ingress[0].hostname is defined and kraken_action == 'down'
+  ignore_errors: yes
+
 - name: Delete all service namespaces
   command: >
     kubectl --kubeconfig={{ kubeconfig }} delete namespace {{ item }}
   with_items: "{{ cluster_namespaces }}"
   when: cluster_namespaces is defined
   ignore_errors: yes
+
+- name: Get vpc id
+  shell: "terraform state show -state={{ config_base | expanduser }}/{{kraken_config.cluster}}/terraform.tfstate module.vpc.aws_vpc.vpc  | awk '/^id/{print $3}'"
+  register: vpcid
+  when: kraken_action == 'down'
+  changed_when: false
+
+- name: Set vpcid lookup string
+  set_fact:
+    vpc_lookup: "elbs[?vpc_id=='{{ vpcid.stdout }}']"
+  when: kraken_action == 'down'
+  changed_when: false
+
+- name: Wait for ELBs to be deleted
+  action:
+    module: ec2_elb_facts
+    region: "{{ kraken_config.providerConfig.region }}"
+    aws_access_key: "{{ kraken_config.providerConfig.authentication.accessKey or omit}}"
+    aws_secret_key: "{{ kraken_config.providerConfig.authentication.accessSecret or omit }}"
+    profile: "{{ kraken_config.providerConfig.authentication.credentialsProfile or omit }}"
+  register: elb_facts
+  vars:
+    vpc_lookup: "elbs[?vpc_id=='{{ vpcid.stdout }}']"
+  when: kraken_action == 'down'
+  until: "{{ elb_facts|json_query(vpc_lookup)|length }} <= 1"
+  retries: 600
+  delay: 1
--- a/ansible/roles/kraken.services/tasks/kill-services.yaml
+++ b/ansible/roles/kraken.services/tasks/kill-services.yaml
@@ -3,7 +3,7 @@
   command: >
     kubectl --kubeconfig={{ kubeconfig }} get services --all-namespaces -o json
   register: added_services
-  when: kraken_action == 'down'
+  when: kraken_action == 'down' and kubeconfig|is_file
   ignore_errors: yes
 
 - name: Register services fact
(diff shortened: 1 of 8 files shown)


ansible/roles/kraken.services/tasks/kill-services.yaml
```diff
@@ -44,9 +33,43 @@
     kubectl --kubeconfig={{ kubeconfig }} delete deployment {{ tiller }} --namespace=kube-system
   when: tiller_present|success
 
+- name: Clean up services
+  command: >
+    kubectl --kubeconfig={{ kubeconfig }} delete --namespace {{ item.metadata.namespace }} svc {{ item.metadata.name }} --grace-period=600
```
coffeepac: does the grace period introduce a polling loop that waits for the operation to complete?  I see we are still polling for the ELBs to be deleted anyway.  and will ansible run these deletes in parallel or could this end up waiting 600 seconds for each ELB if the AWS API server is slow?

coffeepac:
one comment and that's more clarification for me.  otherwise, LGTM

ansible/roles/kraken.services/tasks/kill-services.yaml
joejulian: It does not. I might as well revert that change. I tried it hoping it would be sufficient.

coffeepac:
retest this please

coffeepac:
this can be merged once the test passes.

joejulian:
Go go Jenkins!

joejulian:
Go go Jenkins!

joejulian:
Go go Jenkins!

ansible/roles/kraken.services/tasks/kill-services.yaml
```diff
@@ -37,16 +26,49 @@
...
+    aws_access_key: "{{ kraken_config.providerConfig.authentication.accessKey or omit}}"
+    aws_secret_key: "{{ kraken_config.providerConfig.authentication.accessSecret or omit }}"
+    profile: "{{ kraken_config.providerConfig.authentication.credentialsProfile or omit }}"
+  register: elb_facts
+  vars:
+    vpc_lookup: "elbs[?vpc_id=='{{ vpcid }}']"
+  when: kraken_action == 'down' and kraken_config.provider == 'aws'
+  until: (elb_facts is none) or (elb_facts|json_query(vpc_lookup) is none) or (elb_facts|json_query(vpc_lookup)|length <= 1)
+  retries: 600
+  delay: 1
```
davidewatson: I think this will cause us to exceed API rate limits. Maybe we should do a quick benchmark of the average ELB deletion time and then make that the delay?

joejulian: > 1000 request per second (rps) with a burst limit of 2000 rps 

Unless there's 1000 k2 doing this from one account, we _should_ be ok.

davidewatson: It's probably more acrobatics than we need, but an geometrically decreasing delay might be one way to minimize the total time we wait. :) I am more concerned with making this robust for users than making it fast for developers.

davidewatson: I believe you are quoting the rate limit for the AWS API Gateway Service, which is different than the rate limit for an AWS account.

http://docs.aws.amazon.com/AWSEC2/latest/APIReference/query-api-troubleshooting.html#api-request-rate

I am having trouble finding documentation on the rate limits for Describe, Modify, and Create actions. You are doing a describe here, so that is safer. Note that we have run into rate limits on our team in the past though, and the documentation specifically recommends an exponential back-off...

joejulian: And ansible doesn't have the facility to do that.

joejulian: The feature has been requested upstream, however. https://github.com/ansible/ansible/issues/20226

joejulian: I argue that we go ahead with the 1 second loop and revisit this if there actually is a problem or if a backoff feature is added to ansible.

davidewatson: So here we are using the ec2_elb_facts ansible module:

https://github.com/wimnat/ansible-modules/blob/master/ec2_elb_facts/ec2_elb_facts.py

This uses boto.ec2.elb which I think implements a binary exponential backoff:

https://github.com/boto/boto/blob/abb38474ee5124bb571da0c42be67cd27c47094f/boto/connection.py

So _maybe_ your change is safe. I know I am being pedantic here, so if others disagree I won't argue further. It's just 1 second really seems excessive when we know ELB operations are much slower.

joejulian: Unfortunately the boto retries are for opening the connection and not the results of the query. 

I'm just suggesting we not pre-optimize. I haven't seen any rate limit errors in all the times I've upped and downed - and that's on us-west which is (apparently) notorious for rate limiting. If it turns out that it's a problem, we can easily change the delay later.

Btw... I've recently been pretty consistently getting 5 retries before it moves on and that's _with_ central-logging added.

coffeepac: this looks fine to me.  API limits (as described here:  https://github.com/kubernetes/kubernetes/issues/39526, seem to be a problem in the thousand plus per hour.  If we want this to have an exponential backoff (seems fine) lets open a fresh ticket and do the work there.

davidewatson: OK, we'll go with it.

davidewatson: LGTM.

PR #182, opened 2017-02-06
Don't use temporary directory for $HOME

joejulian:
Revert `mkdir -d` substitution for $HOME

Document how to map the needed files instead of $HOME
    
Add documentation to the README.md for binding just the files we need
for K2 instead of the entire home directory. This prevents the K2
processes from writing potentially overwriting files in the user's home
directory when running as a container.
    
Fixes #181 and is associated with #97
    
Similar changes to k2cli should be done, https://github.com/samsung-cnct/k2cli/issues/51

PR #187, opened 2017-02-10
remove all refs to naked HOME or '~' from README

### LINKED ISSUE #96: Specifying a /root Mount that Is not $HOME Does Not Work.
In my first start of K2, I wanted to isolate my machine from any files that might be created by the startup.  So I created a k2wrap directory, and specified that as my /root mount.  This work for some items, but there are many assumptions that  there will be other $HOME directories available there.  e.g. .ssh, .aws.  This was not spelled out in the documentation.

Also, there may also be some other hard coded assumptions that ~/ is the mount.

Please document this better, and if it is not allowable to use a mount other than $HOME say that!
I would advocate allowing the /root mount to be wherever the user wants.

coffeepac:
by 'naked HOME or '~'' I mean '~/' alone.  '~/.ssh' is fine as it
clearly states what directory out of the home directory is needed.
This is being done to clarify what directories are needed if a user
doesn't want to use the standard directories

this fixes #96

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ If this is your first time using K2, use the K2 Docker image to generate a 'sens
 With the Docker container:
 
 ```bash
-docker run -v ~:/root --rm=true -it quay.io/samsung_cnct/k2:latest ./up.sh --generate
+docker run -v ~/.kraken:/root/.kraken --rm=true -it quay.io/samsung_cnct/k2:latest ./up.sh --generate
 ```
 
 With the cloned repo:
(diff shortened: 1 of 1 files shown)


joejulian:
I prefer https://github.com/samsung-cnct/k2/pull/182/files#diff-04c6e90faac2675aa89e2176d2eec7d8

joejulian:
Plus, my version more closely emulates what's actually being done in k2cli.

coffeepac:
I forgot you had done this already.  I like your version better as well.

PR #203, opened 2017-02-16
minor typo

gochist:


sostheim:
Good catch, very observant!  LGTM.

PR #204, opened 2017-02-16
Initial tpc provider framework

joejulian:


davidewatson:
This code might be necessary, but not sufficient, to support Joyent. Since you have created a branch, I'm fine with merging it now or waiting until it is more complete. My preference is the later since I am not sure this code will run as is. Thoughts?

joejulian:
Yes, the intention is not to verify functionality but to just get another pair of eyes to look for obvious errors, to ensure that the PR is of reasonable scope, to look at style, and to make sure there's some familiarity with the code as this progresses.

This is a branch because it's expected to be broken until it's ready to be merged in to master.

PR #207, opened 2017-02-17
Check if the kube-networking namespace exists

joejulian:
On a second "up" run, `Ensure the kube-networking namespace exists`
failed with:
> "Error from server: namespaces \"kube-networking\" already exists", "stdout"

This checks to see if the namespace has already been created and skips
creation if it does.


coffeepac:
👍

Issue #205, opened 2017-02-16
k2 should allow a custom version of docker to be installed


davidewatson:
Currently careen generates custom docker builds. k2 should be able to use these artifacts to run a custom version of docker.

coffeepac:
ToReview:  have someone other then @davidewatson use the custom version of docker

davidewatson:
To complete this:

1) Add

```
type: tgz
url: https://s3-us-west-2.amazonaws.com/samsung-cnct/artifacts/docker/v1.12.x/docker-1.12.6%2B7ab89465.tgz
```

to your containerConfig stanza in k2config.yaml and run k2 up.

2) Verify that you can create a kubernetes resource (pause pod will do).

Note that we really should be running conformance and (more importantly) density to test this and we plan to do so automatically once additional CI work is completed.

coffeepac:
@joejulian slow hand raised to review this.

joejulian:
:+1:

Issue #210, opened 2017-02-17
Make containerConfig a property of nodepools instead of kubeConfigs


davidewatson:
Currently containerConfigs are used in two different ways within the k2 configuration file: one way for etcd nodes and one way for all other nodes (master, cluster, and special). For example, for etcd* nodes it is specified like so:

```
  etcd:
    -
      name: etcd
      clientPorts: [2379, 4001]
      clusterToken: espouse-monger-rarely
      nodepool: etcd
      peerPorts: [2380]
      ssl: true
      version: v3.0.3
      containerConfig: dockerconfig
    -
      name: etcdEvents
      clientPorts: [2381]
      clusterToken: animism-training-chastity
      nodepool: etcdEvents
      peerPorts: [2382]
      ssl: true
      version: v3.0.3
      containerConfig: dockerconfig
```

However, all other nodes get their container config from their kubeConfig. For example:

```
  kubeConfig:
    -
      name: krakenKubeConfig
      version: v1.5.2
      hyperkubeLocation: gcr.io/google_containers/hyperkube
      containerConfig: dockerconfig
```

The proposal is to decouple containerConfig from kubeConfig and allow each nodepool, including etcd* nodes, to have a different containerConfig. In addition to being more consistent, it will simplify the ansible rendering of drop-ins because we will not have to have multiple loops to render templates which depend on containerConfig. More broadly, one might argue for all properties to be associated with nodepools.

coffeepac:
I like this.  consistency and simplicity are good.  keep it to just docker configs for now.  we can sort out the rest of the properties later.

👍 

@maratoid @venezia I hate to bug you but I'd like both of you to review this.  We all broadly agreed to this in slack but its real now.

davidewatson:
Note to self: The fact that the k2 config makes the etcd and nodes sections "a lists of dicts" but the master section "a dict" also complicates the required ansible templating.

davidewatson:
Another note: currently the kraken.*.docker roles are currently only used to configure kubernetes (template kubeconfig, apiserver, et al.) Maybe this should renamed, or maybe it should be documented. Either way, it is outside of the scope implementing custom docker.

```
rodin:k2 dwat$ ls ansible/roles/kraken.master/kraken.master.docker/tasks/
main.yml			render-manifests.yaml		render-write-files.yaml
render-kubeconfig.yaml		render-units.yaml		service-account-key.yaml
rodin:k2 dwat$ 
```

[...]

PR #211, opened 2017-02-17
Initial checkin of custom docker configuration option. Addresses #205.

davidewatson:


davidewatson:
The templating of this code is incomplete pending a discussion of #210.

davidewatson:
retest this please

davidewatson:
This will not pass pr checks until https://github.com/samsung-cnct/kraken-ci-jobs/blob/master/k2-configs/default-e2e.yaml is updated. I am waiting to get sign off on the config file change here before making the require ci-jobs changes.

Documentation/kraken-configs/container.md
```diff
@@ -6,9 +6,13 @@
 | Key Name | Required | Type | Description|
 | --- | --- | --- | --- |
 | name | __Required__ | String | Configuration name |
-| runtime | Optional | String | Name of container runtime. At this time only docker is supported |
+| runtime | __Required__ | String | Name of container runtime. At this time only docker is supported |
+| type | Optional | String | Type of docker installation. If not specified distro is assumed. |
+| url | Depends | String | If type is not distro, URL where the installation medium can be found |
 
-# Example
+Type may be one of [distro, tgz]. If type is distro then the distribution provided installation of docker will be used. If type is tgz then url must be a valid url pointing to a gzipped tarball containing a single directory. The contents of the directory must include all binaries needed to install the container runtime. Those binaries will be placed in /opt/cnct/bin and systemd drop-ins will be generated to start the container runtime.
```
venezia: Can we boldface part of the last sentence?

**These binaries will be placed in `/opt/cnct/bin`**

ansible/roles/kraken.cluster_common/templates/docker-wait-for-install.part.jinja2
```diff
@@ -0,0 +1,14 @@
+- name: docker.service
+  drop-ins:
+    - name: wait-docker-install.conf
+      content: |
+        [Unit]
+        After=docker-install.service
+        Requires=docker-install.service
+- name: early-docker.service
```
venezia: Since `early-docker` has been removed since ~Summer 2016, can we remove references to early-docker?  This includes references in the format storage.  (still need the format to happen, just not the before early docker reference)

ansible/roles/kraken.cluster_common/templates/docker-install.part.jinja2
```diff
@@ -0,0 +1,10 @@
+- name: install-docker.service
+  command: start
+  content: |
+    [Unit]
+    Description=Install custom docker
+    ConditionPathExists=!/opt/cnct/bin/dockerd
+    [Service]
+    Type=oneshot
+    ExecStart=/bin/mkdir -p /opt/cnct/bin 
+    ExecStart=/bin/bash -c '/usr/bin/wget {{item.nodepool.containerConfig.url}} -O - | tar -xz --strip-components=1 -C /opt/cnct/bin'
```
venezia: Not that it matters for today, but perhaps we should have a check to see docker is already installed and up to date before fetching it and unpacking it?  Just seems like a waste of bandwidth for every node on every boot.

Also what happens if the URL is dead (sites do go down) - should the node fail here right away?  If it already has a copy of docker should it continue?  Should there be options for this?

davidewatson:
Note that this change must be checked in at the same time as kraken-ci-jobs/#165

ansible/roles/kraken.cluster_common/templates/docker-install.part.jinja2
davidewatson: Good call! I created a script which downloads and verifies a checksum iff docker has not been successfully installed yet.

I did not check if there is a later version of docker however, because it is not clear to me that users should always upgrade, and in any case that is a more general problem affecting k8s and any other component which we install.

ansible/roles/kraken.cluster_common/templates/docker-install.part.jinja2
```diff
@@ -0,0 +1,8 @@
+- name: install-docker.service
+  command: start
+  content: |
+    [Unit]
+    Description=Install custom docker
+    [Service]
+    Type=oneshot
+    ExecStart=/bin/bash <(/usr/bin/curl "https://s3-us-west-2.amazonaws.com/samsung-cnct/artifacts/install-artifact.sh") -s -u {{item.nodepool.containerConfig.url}} -l /var/run/cnct-docker-installed.lock
```
joejulian: :-1: Do not run shell scripts from curl unless you can verify the hash first.

joejulian: Also, this won't work when air gapped.

davidewatson: It may be useful to check the hash. The AWS certificate, or one of our accounts, could be compromised and the install script could be replaced by a malicious one. Note that we can't really include this script in the cloud init because we are already running up against the size limit of what AWS will store for us. Also, wrt to air gapped installations, the url is set in the config and can be changed to a protocol and location available in your environment.

davidewatson:
Given your comment about air-gapped installations, should we change the k2 config from ```     url: "https://s3-us-west-2.amazonaws.com/samsung-cnct/artifacts/docker/v1.12.x/docker-1.12.6%2B7ab89465.tgz"``` to ```      scriptUrl: "https://s3-us-west-2.amazonaws.com/samsung-cnct/artifacts/install-artifact.sh"
     binaryUrl: "https://s3-us-west-2.amazonaws.com/samsung-cnct/artifacts/docker/v1.12.x/docker-1.12.6%2B7ab89465.tgz"``` binaryUrl would still be __required__ when type=tgz but scriptUrl could be optional, defaulting to us if not specified. One thing I don't like about this is that it means the contents of scriptUrl would have to conform to the arguments expected by install-artifact.sh. We improve could the situation by defining installCmd instead of url or *url. Thoughts?

joejulian:
I'd still prefer we have some signature checking but the ssl cert guarantee is good enough for now.

davidewatson:
If were going to allow a custom command to specified (via a installCmd option) then I think the containerConfig type should be "custom" and not "tgz".

davidewatson:
After dancing around the multiple levels of escaping necessary to include a custom command in the k2 config (YAML -> terraform -> cloud-init -> systemd -> bash) I think we should punt on supporting custom commands for now. Instead we should keep the tgz type and introduce a custom type in the future for airgapped solutions.

Documentation/kraken-configs/container.md
```diff
@@ -6,9 +6,13 @@
 | Key Name | Required | Type | Description|
 | --- | --- | --- | --- |
 | name | __Required__ | String | Configuration name |
-| runtime | Optional | String | Name of container runtime. At this time only docker is supported |
+| runtime | __Required__ | String | Name of container runtime. At this time only docker is supported |
+| type | Optional | String | Type of docker installation. If not specified distro is assumed. |
```
coffeepac: I think this is required now?

coffeepac:
One nit:  otherwise, LGTM!

Documentation/kraken-configs/container.md
davidewatson: You're right, fixed. I forgot to change the docs in part because I didn't like the additional requirement. However in view of our Node Mgmt discussions I am less certain.

PR #212, opened 2017-02-17
Work around aws elb EOF fault

joejulian:


ansible/roles/kraken.fabric/kraken.fabric.flannel/tasks/main.yml
```diff
@@ -29,22 +29,20 @@
...
     kubectl --kubeconfig={{ kubeconfig | expanduser }} create namespace kube-networking
+  register: ensure_kube_networking_namespace
+  until: ensure_kube_networking_namespace.rc == 0
   retries: 120
   delay: 1
   when: not kube_networking_namespace.rc == 0
-  ignore_errors: yes
 
-- name: check kube-networking namespace state
+- name: work around elb EOF issue
```
davidewatson: Isn't this a k2 problem and not an ELB problem? If so, maybe this should be called "wait for kube-networking namespace to be available" or similar. /me pedantic

PR #216, opened 2017-02-23
Master nodes are taking up to 4.5 minutes to be listening

joejulian:
Since the ELB health checks expire in 200 seconds, often the api server
is not yet listening. Extend the check interval to 45 seconds to allow a
450 second start time.


coffeepac:
👍

PR #226, opened 2017-02-27
Delete cluster when not fully up

joejulian:
If `up` fails for whatever reason, `down` should still delete the cluster.


coffeepac:
@joejulian I don't see how the original code prevents cluster deleting because of a failed up run.

I do see how its forces the route53 information to be written out so that we can always clean up the route53 entry (note:  does this explode if the route53 entry doesn't get created?).  is the entire fix or am I missing something?

coffeepac:
LGTM

PR #228, opened 2017-02-27
Let terraform remove the dns entries

joejulian:
terraform 0.8.6 seems to be able to remove the dns entries just fine.
Without an upstream issue to track, I think this workaround is no longer
necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,6 @@ You will need the following installed on your machine:
 - Ansible 2.2.x
 - Cloud SDKs
  - aws cli
- - cli53 (https://github.com/barnybug/cli53/releases)
  - gcloud SDK
 - Terraform and providers
  - Terraform 0.7.x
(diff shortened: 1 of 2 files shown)


coffeepac:
was there an upstream issue to track before or was it just broken?

coffeepac:
talked offline and there was no explicit issue upstream we were tracking and nothing created internally for this.

I'd like to see another PR to remove the route53cli package from k2-tools before merging this and then run the tests again.  Just to make sure it all works and we're not carrying an undeeded binary around.

joejulian:
retest this please

Now that amazon isn't broken

coffeepac:
retest this please

coffeepac:
test passed!  I had asked that the upstream k2-tools repo be rebuilt first but seeing as we don't have any versioning in this that was a dumb idea.  ie:  if we let the tools repo build finish before this one, we won't have any valid path for route53 operations.

on that note, Imma merge this PR now.  It already passed the initial test.

PR #242, opened 2017-03-13
Rename .yml extensions to .yaml

joejulian:
Be consistent with file extensions. Use yaml instead of yml.


davidewatson:
If it passes test, LGTM.

PR #244, opened 2017-03-13
Add is_empty test

joejulian:
This test eliminates long cumbersome definition tests like

```
(kraken_config.providerConfig.type is undefined) or (kraken_config.providerConfig.type is none) or (kraken_config.providerConfig.type|trim == '')
```
with

```
kraken_config.providerConfig.type | is_empty
```


coffeepac:
LGTM

PR #264, opened 2017-03-16
A small change to readme for clarity

alejandroEsc:
Added a blurb about what and where YOURCLUSTER means in general, since it was never truly defined and was difficult to understand at first as an outside.

Ideally i do not like having things mentioned outside the order of which it can happen, but I understand when writing it can be difficult to organize, so I wanted to add some clarity to a few places to make sure that order of operations does matter, yeah we can debate that point as obvious, but clarity is preferable than assuming something is obvious.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ should have unique names.
 
 ### Preparing AWS credentials
 
-If you already have configured your machine to be able to use AWS, you can skip this step_
+_If you already have configured your machine to be able to use AWS, you can skip this step_
 
 To configure the environment with your AWS credentials, run one of the following commands:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -111,9 +111,25 @@ This will generate a config.yaml file located at
...
+YOURCLUSTER.yaml
+```
+
+It is particularly useful when trying to create and manage multiple clusters, each of which
+should have unique names.
+
 ### Preparing AWS credentials
 
-_If you already have configured your machine to be able to use AWS, you can skip this step_
+If you already have configured your machine to be able to use AWS, you can skip this step_
```
coffeepac: not sure why we're removing the italics here.  was that intentional?

alejandroEsc: argh, wasnt intentional i believed i made a typo at the moment.

joejulian:
LGTM

PR #266, opened 2017-03-17
Initial checkin of config format changes.

davidewatson:
This checkin is to allow further review to occur publicly instead of slack. A few more changes are necessary before this can be tested in a backwards compatible fashion.


davidewatson:
Closed in lieu of #276.

PR #282, opened 2017-03-22
Add an "until" to test the retry against

joejulian:
I mistakenly thought that a retries clause without an until would retry
until the command succeeded. The ansible documentation is insufficient
on this point.


alejandroEsc:
could @DStorck try this branch and see if that works?

alejandroEsc:
also, @joejulian good catch.

davidewatson:
LGTM.

DStorck:
LGTM

alejandroEsc:
does the check above require  a /lgtm? with a slash?

coffeepac:
@alejandroEsc no, but @DStorck does need to file a github official review.

alejandroEsc:
got it, as opposed to just a comment.

PR #287, opened 2017-03-24
Added kubernetes_minor_version fact

yenicapotediaz:


ansible/roles/kraken.config/tasks/main.yaml
```diff
@@ -32,6 +32,10 @@
       https://github.com/samsung-cnct/k2/issues to request support.
   when: kraken_config.provider not in ['aws', 'gke']
 
+- name: Create k8s minor version fact
+  set_fact:
+    kubernetes_minor_version: "{{ kraken_config.kubeConfig[0].version[0:4] }}"
```
davidewatson: I'm not sure we want to set this as a fact for all roles. The kubernetes version can vary between nodePools, so you can't really assume the first kubeConfig version.

Instead, from the location where you need this fact, I would reference it as ```kraken_config.master.kubeConfig.version[0:4]``` for master and  ```item.nodepool.kubeConfig.version[0:4]``` for nodes.

The k2recon work will cause this syntax to change, but only slightly.

coffeepac:
ok to test

ansible/roles/kraken.config/tasks/main.yaml
coffeepac: will this string slicing method work for a version like 1.10.2 or 10.5.3 or 10.10.4?

yenicapotediaz: It will not work for those version examples. We shall work on that today!

coffeepac:
@yenicapotediaz @guineveresaenger you will need to rebase this PR against master as it won't pass CI until you do.

guineveresaenger:
Found the k8s version for master nodes and assigned it a minor version using some Python string searching. Will research to see if there's a better way to truncate the version, in case the patch gets an extension.

coffeepac:
@guineveresaenger looks like the rebase was successful.  However, the GKE PR tester failed.  You may need to try spinning up a GKE cluster via k2 to check what exactly is going wrong.

guineveresaenger:
I don't think I have any kind of GKE authorizations. What does the tester do?

coffeepac:
You should have GKE privs.  It used to be part of onboarding.  Try typing `gcloud config list` in a terminal and see what you get back.

The GKE test spins up a cluster and spins it back down.  Nothing fancy.  The config is failing to render.

guineveresaenger:
I'm fairly certain GKE wasn't part of our onboarding...
`-bash: gcloud: command not found`

coffeepac:
nuisance, you're right.  okay, lets setup some GKE creds and whatnots tomorrow.

guineveresaenger:
This will inly create kubernetes_minor_version if the provider is aws.

Issue #275, opened 2017-03-20
Migrate etcd, master, node and nodepool stanzas


davidewatson:
# etcd, master, node and nodepool stanzas

This ticket is to remove the old stanzas in the heading from ansible/roles/kraken.config/files/config.yaml  and update the related Ansible tasks to match. When doing this the k2 schema, ansible/roles/kraken.config/files/schema.json , should be updated as well.

davidewatson:
# Notes

The nodepool stanza contains a mishmash of concepts: number of nodes, shared kube/container/os settings, provider specific volume configuration and node specific mount points. Under the k2recon, this information has been moved to the “nodes” (number of nodes and shared configuration)) and “nodeConfig” (provider and node) stanzas. Most of the configuration work should be completed by #269, #270, #271, #272, and #273 .

provider volume and node mountpoint configuration is stored in a new nodeConfig stanza. Is there any benefit to breaking these up a la k8s volume and volumeMounts stanzas?

davidewatson:
Blocked on #276.

davidewatson:
Looking only at etcd, this looks like it will be a fair amount of work. There are > ~80 references. Therefore I am assigning to myself until I either complete the work or figure out how to break it up into more manageable pieces.

davidewatson:
Instead of removing the etc, master, and node roles now, first I am converting all of the templates to use the new item (cluster, node) tuple which is passed to each template task.

```
diff --git a/ansible/roles/kraken.master/kraken.master.docker/tasks/main.yaml b/ansible/roles/kraken.master/kraken.master.docker/tasks/main.yaml
index 0a26658..7cde888 100644
--- a/ansible/roles/kraken.master/kraken.master.docker/tasks/main.yaml
+++ b/ansible/roles/kraken.master/kraken.master.docker/tasks/main.yaml
@@ -1,6 +1,10 @@
 ---
 - include: service-account-key.yaml
 - include: render-manifests.yaml
+  with_subelements:
+    - "{{ kraken_config.clusters }}"
+    - nodes
+  when: item.1.apiServerConfig is defined
 - include: render-kubeconfig.yaml
 - include: render-write-files.yaml
-- include: render-units.yaml
\ No newline at end of file
+- include: render-units.yaml

This could be done more succinctly if roles allowed for with_*, and as of ansible 2.3.x, they do, so I opened a ticket to upgrade ansible in k2-tools.

davidewatson:
So 2 things remain here:

[...]

PR #300, opened 2017-04-02
Migrate etcd, master, node and nodepool stanzas (#275) for k2recon

davidewatson:
work.

### DIFF
--- a/ansible/roles/kraken.master/kraken.master.docker/tasks/main.yaml
+++ b/ansible/roles/kraken.master/kraken.master.docker/tasks/main.yaml
@@ -1,6 +1,40 @@
 ---
 - include: service-account-key.yaml
+  with_subelements:
+    - "{{ kraken_config.clusters }}"
+    - nodePools
+  loop_control:
+    loop_var: cluster_node_tuple
+  when: cluster_node_tuple.1.apiServerConfig is defined
+
 - include: render-manifests.yaml
+  with_subelements:
+    - "{{ kraken_config.clusters }}"
+    - nodePools
+  loop_control:
+    loop_var: cluster_node_tuple
+  when: cluster_node_tuple.1.apiServerConfig is defined
+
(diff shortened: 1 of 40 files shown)


davidewatson:
Note that the master and node roles use different but analogous variable naming conventions (e.g. ```{{ item }}``` vs. ```{{ outer_item }}```). This will be fixed once a one true way is agreed upon. I currently consider the master role way to be canonical (i.e. ```{{ outer_item }}```).

davidewatson:
As part of validation we will also need to ensure that each node has exactly one of etcConfig, (apiServerConfig and kubeConfig) or kubeConfig set. Otherwise the roles may generate files which conflict.

davidewatson:
When reviewing this, please pay particular attention to the master and node roles (the etcd role is rather short and simple anyway). I have unsubmitted code which converts the existing node role code to follow the master role convention whereby we set facts:

```
- set_fact:
    cluster: {{ outer_item.0 }}
    node: {{ outer_item.1 }}
```

And use those through the templates, to aid in readability. I would like agreement on this path forward before I amend this PR any further, mainly because it is time consuming and tedious to refactor. Leaving the PR as is for now, also makes the ramifications of the choice more clear.

ansible/roles/kraken.master/kraken.master.docker/tasks/main.yaml
```diff
@@ -1,6 +1,41 @@
 ---
 - include: service-account-key.yaml
+  with_subelements:
+    - "{{ kraken_config.clusters }}"
+    - nodes
+  loop_control:
+    loop_var: outer_item
```
davidewatson: We really need a better name for this. cluster_node_tuple?

davidewatson:
Also note that I am merging this to the k2recon branch because once we merge to master, all old school configs will break. I am working to add a warning about this fact now.

PR #304, opened 2017-04-03
[Do not merge] K2recon

davidewatson:


davidewatson:
Was only created to create a clean diff.

PR #308, opened 2017-04-04
Attempt to warn users of the impending config file changes.

davidewatson:


davidewatson:
We must tag the current version of k2 before committing this. Since the current version of k2 already has the new and backwards compatible config file format, the advice in the warning will be valid.

davidewatson:
please retest this

davidewatson:
retest this please

PR #311, opened 2017-04-04
Use the actual cluster name in the help text

joejulian:
Also rename `bin/utils.sh` `lib/common.sh`

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -
-#title           :utils.sh
-#description     :utils
+#title           :common.sh
+#description     :common
 #author          :Samsung SDSRA
 #==============================================================================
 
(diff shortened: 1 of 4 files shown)


lib/common.sh
```diff
@@ -15,6 +15,29 @@ KRAKEN_TF_LOG=k2_tf_debug.log
 # set RANDFILE to prevent creation of ${HOME}/.rnd by openssl
 export RANDFILE=$(mktemp)
 
+function parse_yaml {
```
davidewatson: Part of me would rather depend on and use a proper YAML parser (e.g. https://github.com/kislyuk/yq appears to be canonical), but when a YAML file only consists of name value pairs, it is a regular language, so I think the sed/awk implementation is probably safe for current configuration files.

But with new configurations, where there is no cluster, but instead, a list of clusters, are we sure this will still work?

joejulian: Good point. In fact, not only will it not work for a list of clusters, but the entire help text will need changed.

I would prefer merging this for the current configuration and coming back at it when we have a list since we would have to anyway.

davidewatson: My hope is that we will have a list in a weeks time (or two tops) though. Is it really worth it? Perhaps it is. We'll tag k2 0.0 soon, and then anyone who wants to cling to the old style config can benefit for as long they insist.

As an aside, I've used that same code before, mainly for the nostalgia I experience when I see someone write a moderately complicated awk program.

PR #317, opened 2017-04-06
Make readiness a cluster level property instead of a top level

davidewatson:
one. In a future with federated clusters, this value will not
be the same for all clusters.


davidewatson:
please retest this

coffeepac:
retest this please

PR #319, opened 2017-04-06
added upgrade for master nodes

DStorck:
this adds and upgrade.sh which will update master nodes one by one


coffeepac:
@DStorck is this PR covered by #320?

coffeepac:
@DStorck do we still need this PR or is it covered by #320?

DStorck:
@coffeepac I think  #320 is the real one

Issue #269, opened 2017-03-20
Migrate providerConfig


davidewatson:
# Boilerplate

The k2recon project involves creating a versioned schema to validate configuration files. As a result, some of configuration properties have been renamed, and stanzas have been reorganized. For backwards compatibility while we migrate to the new config, the old stanzas have been left in place alongside the new stanzas.

# providerConfig stanza

This ticket is to remove the old stanza, ```"#/deployment/providerConfig"```, from ansible/roles/kraken.config/files/config.yaml  and update the related Ansible tasks to match. When doing this the k2 schema, ansible/roles/kraken.config/files/schema.json , should be updated as well.

davidewatson:
# Notes

If GKE involves significantly more work, then consider opening another ticket.

The providerConfig stanzas are long. @joe suggested that we consider provider.d directory structure and automatically include provider stanzas.

davidewatson:
Blocked on #276.

coffeepac:
in the case of multiple clusters and multiple providers do we want to validate all providers as supported first and then start rendering files?  or is it okay to blindly create clusters until we hit an unsupported provider?

I'm leaning towards the second option being okay for now if we file an issue to fix it as we aren't publicising our ability to support multiple clusters at once but this would be a required feature if we were telling people you can control many clusters from one file.

coffeepac:
I have a hit stopping point for tonight:  
- I have converted the aws tasks
- I am stuck on how to loop over the provider roles.  there is a new to ansible 2.2 module, [include_role](https://docs.ansible.com/ansible/include_role_module.html) which looks like it should do the trick but I'm having issues getting the loop variable to be exported in the included role

coffeepac:
I was having indentation issues last night, got 'em sorted now.  I have migrated the aws provider, tested the broken provider and will now spin up an aws cluster.  while that runs I'll start migrating the gke templates.

coffeepac:
Oh man, porting over the gke provider is huge.  Its a sea of kraken_config vals.  Bleargh.

coffeepac:
got most of the gke provider role converted.  now I need to convert the ssh role as that's also keyed off of provider.

coffeepac:
PR is up and ready to be merged after a successful review.  This should be closed when #332 is done.

coffeepac:
PR is merged, closing.

PR #332, opened 2017-04-12
convert provider stanza to use new style config

coffeepac:
WIP:  need another commit to convert the GKE provider

~~the dry_run role will set a global fact named dry_run to be true.  we are also now defaulting
this global fact to be false.  this allows us to only run part of a role when the dry_run tag
is set.  currently, tags can only be applied at the role level.~~
removed the dry_run role as it didn't work.  I had a typo that was kind of making it work but when I fixed the typo things fell apart.

AWS provider conversion:  this includes converting aws tasks and templates.  this also
includes converting the selector role for providers from a list of dependency roles to using
the new (as of ansible 2.2) preview module (beta in most parlance) named 'include_role' which
allows the developer to use roles like tasks, including the ability to loop over role executions.
which is required for a list of clusters where each cluster can have a different provider.  the
GKE selector has been ported but I don't think anything good can happen if you try to use GKE
with this PR as-is.  I think there is some mutually exclusive config going on.

This handles #269


ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/main.yaml
```diff
@@ -1,16 +1,19 @@
...
+- set_fact:
+    cluster: "{{ cluster_node_tuple }}"
+
 - name: Gather inventory of all cluster nodes
   ec2_remote_facts:
-    region: "{{ kraken_config.providerConfig.region }}"
-    aws_access_key: "{{ kraken_config.providerConfig.authentication.accessKey or omit }}"
-    aws_secret_key: "{{ kraken_config.providerConfig.authentication.accessSecret or omit }}"
-    profile: "{{ kraken_config.providerConfig.authentication.credentialsProfile or omit }}"
+    region: "{{ config.providerConfig.region }}"
```
davidewatson: s/config/cluster/g?

ansible/roles/kraken.provider/kraken.provider.aws/tasks/main.yaml
```diff
@@ -1,4 +1,7 @@
 ---
+- set_fact:
+    cluster: "{{ cluster_node_tuple }}"
```
davidewatson: Or leave it the same as elsewhere for consistency, and I can rename it later?

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/main.yaml
coffeepac: yeah, there are a bunch of these.  My typing skills were a bit muddled this afternoon.  I'll have another PR up in a bit, struggling with keyPair creation now.

ansible/roles/kraken.provider/kraken.provider.aws/tasks/main.yaml
coffeepac: haha, yeah, its neither a tuple nor does it include cluster.  I'm happy to set it to cluster in the top level.  its consistent but dumb on my part.

coffeepac:
@davidewatson there's something wonky with using cluster as loop_variable, I'm not sure what but when I did use cluster as the loop var in the selector roles I got a warning about what seemed to be a scoping issue.  Because include_role is preview I am erring on the side of caution and putting in an extra level of caution.

I can gen the warning tomorrow if you'd like.  Only takes 5 minutes.

Otherwise, this code works now for both aws and gke providers.  I have sample configs for both aws and gke as well I can share for anyone who wants 'em.

Issue #274, opened 2017-03-20
Migrate clusterServices to helmConfig


davidewatson:
# clusterServices stanza

This ticket is to remove the old stanza, ```"#/deployment/clusterServices”```, from ansible/roles/kraken.config/files/config.yaml  and update the dependent Ansible tasks to access ```#/deployment/clusters/{{clusterIndex}}/helmConfig``` instead. When doing this the k2 schema, ansible/roles/kraken.config/files/schema.json , should be updated as well.

davidewatson:
Blocked on #276.

DStorck:
Things are working, but I've had to move a chunk of variables out of default/main.yaml because they are now dynamically created. It has forced an unsettling bit of reduncancy into kill-services and run-services.yaml for variable creation. I'm not loving it but I don't see another way at the moment.

DStorck:
PR is here https://github.com/samsung-cnct/k2/pull/333

coffeepac:
PR is here and currently unmerged:  https://github.com/samsung-cnct/k2/pull/334

guineveresaenger:
PR is closed. Seems done.

PR #333, opened 2017-04-12
migration for kraken.services

DStorck:


ansible/roles/kraken.config/files/config.yaml
```diff
@@ -4,7 +4,7 @@ version: v1alpha
 definitions:
   helmConfigs:
     - &defaultHelm
-      name: defaultHelm
+      name: helm
```
davidewatson: Unlike the defaultEtcd to etcd name change, which was only required if you used a master config file with the k2recon branch, the helm name should probably stay defaultHelm.

ansible/roles/kraken.config/tasks/main.yaml
```diff
@@ -25,6 +25,17 @@
     msg: "Make sure that the 'cluster' is defined in {{ config_file }}"
   when: kraken_config.cluster | is_empty
 
+- name: Warn if still using old style configuration format
```
davidewatson: I am not sure why this change, and others, are appearing in your PR, but we should try to clean them up so that the commit only includes the changes for #274.

ansible/roles/kraken.services/tasks/kill-services.yaml
```diff
@@ -1,4 +1,32 @@
...
+    cluster_namespaces: "{{ cluster_namespaces | unique | reject('search', 'kube-system') | list }}"
+
+- name: See if tiller rc if present
+  shell: >
+    kubectl --kubeconfig={{ kubeconfig }} get deployment {{ tiller }} --namespace=kube-system
+  register: tiller_present
+  when: kubeconfig | is_file
+  ignore_errors: yes
+
+- name: Create Helm home
```
davidewatson: I'm curious under what conditions this directory wouldn't exist. It doesn't hurt anything, but if it's useful we should add a comment as to why. Otherwise we should remove it.

ansible/roles/kraken.services/tasks/run-services.yaml
```diff
@@ -1,4 +1,24 @@
...
+    kubeconfig: "{{ config_base | expanduser }}/{{ cluster.name }}/admin.kubeconfig"
+    helm_home: "{{ config_base | expanduser }}/{{ cluster.name }}/.helm"
+    cluster_namespaces: []
+- set_fact:
+    cluster_namespaces: "{{ cluster_namespaces + [item.namespace | default('kube-system')] }}"
+  with_items: "{{ cluster_services }}"
+- set_fact:
+    cluster_namespaces: "{{ cluster_namespaces | unique | reject('search', 'kube-system') | list }}"
+
+- name: Create Helm home
```
davidewatson: Since you do this above as well, maybe we should only do it once in main.yaml?

DStorck: are you talking about creating the path for helm_home in main.yaml?
I wanted to do that but I don't have access to the helm_home variable

ansible/roles/kraken.config/files/config.yaml
DStorck: no problem

ansible/roles/kraken.services/tasks/kill-services.yaml
DStorck: I can do whatever makes more sense but I don't have the context to make that determination.

ansible/roles/kraken.services/tasks/run-services.yaml
davidewatson: Ah, I see. The helm_home and kubeconfig variables are really just conveniences. If you remove them from the defaults, you should be able to iterate over the tasks in main ```with_items: "{{ kraken_config.clusters }}"``` using their definitions.

DStorck: I can put `create helm home` back in main.yaml if i do something like this: ```- name: Create Helm home
  file: >
    path= "{{ config_base | expanduser }}/{{ kraken_config.cluster }}/.helm" 
    state=directory
``` but i thought we were trying to change `kraken_config.cluster` to `cluster.name` wherever we could for the new config style

PR #334, opened 2017-04-13
migration for clusterServices 

DStorck:


leahnp:
LGTM - only comment is aesthetic - and this might be less specifically for this PR and more for the whole k2recon migration effort - a few times `node` is set as a fact and then not used which raises 2 questions, does it need to be set in the first place and additionally can we just pass the task 'cluster' as a singular variable and not do the whole with_subelements loop? I also might be misunderstanding something - let me know what you think.

davidewatson:
If node is not used, then it generally should not be set, because it is then it means we are doing the same thing for each node pool when it really only needs to be done once per cluster. Patrick's provider ticket shows an example of this.

ansible/roles/kraken.services/tasks/main.yaml
```diff
@@ -1,30 +1,20 @@
...
-  shell: >
-    kubectl --kubeconfig={{ kubeconfig }} get deployment {{ tiller }} --namespace=kube-system
-  register: tiller_present
-  when: kubeconfig | is_file
-  ignore_errors: yes
-
 - include: kill-services.yaml
+  with_subelements:
+    - "{{ kraken_config.clusters }}"
+    - nodePools
```
coffeepac: this should be removed as it is unused.

ansible/roles/kraken.services/tasks/main.yaml
```diff
@@ -1,30 +1,20 @@
...
+  loop_control:
+    loop_var: cluster_node_tuple
+  when: cluster_node_tuple.1.apiServerConfig is defined
```
coffeepac: removing the nodePool level loop and only looping over clusters will let you drop this when clause entirely

ansible/roles/kraken.services/tasks/main.yaml
```diff
@@ -1,30 +1,20 @@
...
+
 - include: run-services.yaml
+  with_subelements:
+    - "{{ kraken_config.clusters }}"
+    - nodePools
+  loop_control:
+    loop_var: cluster_node_tuple
+  when: cluster_node_tuple.1.apiServerConfig is defined and kraken_action == 'up'
```
coffeepac: removing the nodePool looping here will also let you remove the 'is defined' portion of the when clause

ansible/roles/kraken.services/tasks/run-services.yaml
```diff
@@ -1,4 +1,25 @@
 ---
+
+- set_fact:
+    cluster: "{{ cluster_node_tuple.0 }}"
+    node: "{{ cluster_node_tuple.1 }}"
```
coffeepac: not needed and should be removed

ansible/roles/kraken.services/tasks/kill-services.yaml
```diff
@@ -1,4 +1,32 @@
 ---
+  
+- set_fact:
+    cluster: "{{ cluster_node_tuple.0 }}"
+    node: "{{ cluster_node_tuple.1 }}"
```
DStorck: sounds good, I'll axe it

ansible/roles/kraken.services/tasks/kill-services.yaml
```diff
@@ -1,8 +1,8 @@
 ---
-  
+
 - set_fact:
-    cluster: "{{ cluster_node_tuple.0 }}"
-    node: "{{ cluster_node_tuple.1 }}"
+    cluster: "{{ a_cluster }}"
+    # node: "{{ cluster_node_tuple.1 }}"
```
coffeepac: please remove this commented out line

DStorck:
should be good to go

PR #338, opened 2017-04-17
upgrade documentation

DStorck:
updated k2 documentation to include instructions to upgrade master nodes

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,7 +416,11 @@ As mentioned above, before you can upgrade the Kubernetes version, you will firs
 docker run $K2OPTS quay.io/samsung_cnct/k2:latest ./up.sh --config $HOME/.kraken/<your_config>.yaml
 ```
 
-Then run `docker run $K2OPTS quay.io/samsung_cnct/k2:latest ./upgrade.sh --config $HOME/.kraken/foo.yaml`
+Then run:
+
+ ```bash
+docker run $K2OPTS quay.io/samsung_cnct/k2:latest ./upgrade.sh --config $HOME/.kraken/<your_config>.yaml
+```
 
 Refer to [Kuberntes release versioning](https://github.com/kubernetes/kubernetes/blob/release-1.5.4/docs/design/versioning.md#supported-releases) for questions about supported versions
 
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Docker Repository on Quay](https://quay.io/repository/samsung_cnct/k2/status "Docker Repository on Quay")](https://quay.io/repository/samsung_cnct/k2)
 
-Please use [k2cli](https://github.com/samsung-cnct/k2cli), the intended user interface to K2. The 
+Please use [k2cli](https://github.com/samsung-cnct/k2cli), the intended user interface to K2. The
 following instructions are intended for developers working on K2.
 
 ## What is K2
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -409,6 +409,21 @@ docker run $K2OPTS quay.io/samsung_cnct/k2:latest ./up.sh --config $HOME/.kraken
 
 Replace `foo.yaml` with the name of the configuration file you intend to use.
 
+#### Upgrading Kubernetes Version of Master Nodes
+As mentioned above, before you can upgrade the Kubernetes version, you will first need to update your configuration file with the intended version and run:
+
+```bash
+docker run $K2OPTS quay.io/samsung_cnct/k2:latest ./up.sh --config $HOME/.kraken/<your_config>.yaml
```
coffeepac: this needs a warning that if you have updated other options that terraform will run wild.  ie:  if you're using the default config for AWS and ContainerLinux has released a new version of CoreOS then your etcd nodes will get trashed and a change will be staged for every node in the cluster that has an ASG.

On a related node, I think we should build additional guards around this process so we don't accidentally torch things.

README.md
```diff
@@ -409,6 +409,21 @@ docker run $K2OPTS quay.io/samsung_cnct/k2:latest ./up.sh --config $HOME/.kraken
...
+```
+
+Then run:
+
+ ```bash
+docker run $K2OPTS quay.io/samsung_cnct/k2:latest ./upgrade.sh --config $HOME/.kraken/<your_config>.yaml
```
coffeepac: does upgrade.sh only operate on the master nodes right now?

DStorck: yes. upgrade tears down the master nodes one by one and leaves the cluster functional during the process, but doesn't touch the kublelets

README.md
joejulian: > On a related node, I think we should build additional guards around this process so we don't accidentally torch things.

As part of this PR? If not, file an issue please.

coffeepac: additional guards should not be in this PR.  the warning should be.

There is a lineitem in the k2 v1.0 tracking ticket for protecting the etcd nodes, which are the ones that matter for now as we don't have plans for how to upgrade them.

README.md
coffeepac: okay.  solid.

README.md
DStorck: @coffeepac I'm not sure that terraform is being touched at all with upgrade.sh . Are you sure that etcd nodes would get touched in this process?  Upgrade.sh uses a module to delete master nodes specifically and then delete them off AWS. They are recreated by the ASG

DStorck: @coffee are you concerned that terraform would run wild on up.sh?

coffeepac: @DStorck yes, the issues come when calling `cluster up`.  that command does a `terraform apply` and that command will do a replace of etcd nodes if the OS version changed.  note:  This is how I destroyed the first common-tools production cluster.

joejulian: Ah, I see the confusion. Should line 416 be up.sh or upgrade.sh?

DStorck: Okay so this warning should have been there in the first place with the mentioned of running up again, it is not specific to upgrade.

DStorck: @coffeepac There is already a warning section right above these lines starting around line 387. It mentions not changing etcd and specifically says what can be modified. Are there other things I should put in here?

Issue #272, opened 2017-03-20
Migrate containerConfig stanza


davidewatson:
# containerConfig stanza

This ticket is to remove the old stanza, ```"#/deployment/containerConfig”```, from ansible/roles/kraken.config/files/config.yaml  and update the dependent Ansible tasks to access ```#/deployment/clusters/{{clusterIndex}}/nodes/{{nodeIndex}}/containerConfig```. instead. When doing this the k2 schema, ansible/roles/kraken.config/files/schema.json , should be updated as well.

davidewatson:
Blocked on #276.

coffeepac:
going to do the assembler role and re-do part of the provider role so I don't have to have separate loops for each node type (etcd, master and node).

coffeepac:
wait, what is this.  I have to migrate and merge kraken.etcd, kraken.master and kraken.nodes.  how did I end up pulling the two largest sets of changes?

coffeepac:
I have merged the three node roles into a single nodePools role with a selector.  This will require some refactoring soon to speed up this portion of the project as I didn't change how the underlying roles operate and each underlying role iterates through all nodePools in a given cluster.  Its a small thing but its going to bother me.  Not enough to fix it right now but it will.

coffeepac:
ToReview:
- linked PR is merged.

guineveresaenger:
Newbie question: If the PR is merged, does that mean someone should move this ticket to Done? Or is there more review work in this case?

coffeepac:
@guineveresaenger should be closed and moved to done!

PR #343, opened 2017-04-18
Container config recon

coffeepac:
these changes cover the transition from using a container config that hangs off the cluster level to one that hangs off of the nodePool level.  This configuration move challenges how we approach container level configurations in a rather fundamental way.  This should be obvious because I think this PR touches everything in our codebase.  

This approach challenge means that we need to start off by iterating over a list of clusters and passing a list of nodePools down.  it also means that we can no longer attempt to access a global etcd or master configuration because in a single config file there may be several sets of both that have no relation to each other.

Anyway, here's wonderwall.


coffeepac:
This resolves #272

ansible/roles/kraken.cluster_common/tasks/docker-install.yaml
```diff
@@ -1,60 +1,22 @@
 ---
 - name: Determine which nodes will use which container runtime
   set_fact:
-    custom_etcd: "{{ kraken_config.etcd
-                   | selectattr('nodepool.containerConfig.runtime', 'equalto', 'docker')
-                   | selectattr('nodepool.containerConfig.type', 'equalto', 'tgz')
+    custom_nodes: "{{ cluster.nodePools
```
joejulian: cluster.nodePools is a list. docker-install.yaml will need to be looped by nodePools as well as by clusters.

coffeepac: docker-install.yaml is being called once per cluster.

I don't think we need to do anything more as selectattr will operate on a list input.  This code isn't changing actually, in the old version both kraken_config.etcd and kraken_config.node are lists

mikeln:
LGTM

PR #346, opened 2017-04-19
changed order of conditional on clean up services

DStorck:
reversed order of `when` clause to short-circuit on `kraken_action == down `


coffeepac:
LGTM - feel free to self-merge when it passes pipelet test

Issue #283, opened 2017-03-22
Create static definition for kubedns


coffeepac:
due to issues raised in #222 we need to create a pod definition for deploying a DNS solution to our kubernetes clusters.  as a first pass this solution should have the following traits:
- use the newest version of KubeDNS (which is just called DNS now, which is kind of annoying)
- be manually tested against v1.4, v1.5 and v1.6
- be installed via an ansible role before helm is installed

coffeepac:
should also include removing kubeDNS from the default set of services

coffeepac:
can find an example of this in kubernetes the hard way by kelsey hightower

mikeln:
kubedns is now:
```
gcloud alpha container images list --repository=gcr.io/google_containers
:
gcr.io/google_containers/k8s-dns-dnsmasq-amd64
gcr.io/google_containers/k8s-dns-dnsmasq-arm
gcr.io/google_containers/k8s-dns-dnsmasq-arm64
gcr.io/google_containers/k8s-dns-dnsmasq-nanny-amd64
gcr.io/google_containers/k8s-dns-dnsmasq-nanny-arm
gcr.io/google_containers/k8s-dns-dnsmasq-nanny-arm64
gcr.io/google_containers/k8s-dns-dnsmasq-nanny-ppc64le
gcr.io/google_containers/k8s-dns-dnsmasq-nanny-s390x
gcr.io/google_containers/k8s-dns-dnsmasq-ppc64le
gcr.io/google_containers/k8s-dns-dnsmasq-s390x
gcr.io/google_containers/k8s-dns-e2e-amd64
gcr.io/google_containers/k8s-dns-e2e-arm
gcr.io/google_containers/k8s-dns-e2e-arm64
gcr.io/google_containers/k8s-dns-e2e-ppc64le
gcr.io/google_containers/k8s-dns-ginkgo-amd64
gcr.io/google_containers/k8s-dns-ginkgo-arm
gcr.io/google_containers/k8s-dns-ginkgo-arm64
gcr.io/google_containers/k8s-dns-ginkgo-ppc64le
gcr.io/google_containers/k8s-dns-kube-dns-amd64
gcr.io/google_containers/k8s-dns-kube-dns-arm
gcr.io/google_containers/k8s-dns-kube-dns-arm64
gcr.io/google_containers/k8s-dns-kube-dns-ppc64le
gcr.io/google_containers/k8s-dns-kube-dns-s390x
[...]

PR #348, opened 2017-04-19
k2-283 Separate DNS Service from Helm

mikeln:
Separate DNS from Helm Services Deployment.
Need separate template for each kubernetes version,  default template pattered after kubernetes v1.4 template for now.
All infrequently changed variables are defaulted in defaults/main.yaml.
Any of those variables can be overridden by setting in the main config.yaml, but most should never be touched. 

https://github.com/samsung-cnct/k2/issues/283

### DIFF
--- a/Documentation/kraken-configs/clusterservices.md
+++ b/Documentation/kraken-configs/clusterservices.md
@@ -37,13 +37,10 @@ clusterServices:
         url: http://atlas.cnct.io
     services:
       -
-        name: kubedns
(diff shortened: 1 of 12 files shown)


mikeln:
Blocked by: https://github.com/samsung-cnct/kraken-ci-jobs/pull/169  merging.  Otherwise pipelet will never work with the new services layout.

mikeln:
moving to recon branch instead.

Issue #345, opened 2017-04-18
Inject kubernetes minor version into kraken_config rather than set it as a fact


guineveresaenger:
Setting k8s minor version as a fact no longer works when looping over multiple clusters.
The work in this ticket will set it as a part of kraken_config.
Related: #337

guineveresaenger:
Question:
What level in the configuration should kubernetes_minor_version live in? Right below cluster? As in "cluster.kubernetes_minor_version"? Or should each node have its own minor version?
paging @coffeepac for input

coffeepac:
@guineveresaenger its a cluster level property.  if we need to change it later we can change it later.  shouldn't be needed for ages if ever.

guineveresaenger:
^^ PR ready for k8s minor versions fact

coffeepac:
PR merged and being used.  Done.

Issue #337, opened 2017-04-17
Migrate GKE k8s minor version fact to new config file


guineveresaenger:
K2recon branched from master before k8s minor version could be set for gke. This ticket will fix that.

guineveresaenger:
Blocked because the new config doesn't reflect a GKE configuration yet.

guineveresaenger:
PR #349 handles this ticket as well.

coffeepac:
PR Merged.

Issue #230, opened 2017-03-02
Node Management Doc updates


coffeepac:
As part of node management we are going to be creating new processes that our end users will need to follow.  These need to be documented.

coffeepac:
this should include exactly how we are setting the kubernetes_minor_version fact, which is how we decide which version of kubernetes clients to use.
aws:  version of control plane
gke: last nodePool in the list

guineveresaenger:
per chat with @coffeepac this should also warn the user that repeated up and update may break the etcd nodes in case a more recent version of coreOS has been released.

guineveresaenger:
Question: We are still using the version of kubectl that lives in `/usr/bin/kubectl`. I'm assuming this will change - what should the README say in the meantime?

coffeepac:
Make it reflect what we will be doing in a week when the outstanding `k2cli tool` issues get resolved.

guineveresaenger:
PR: https://github.com/samsung-cnct/k2/pull/544

PR #349, opened 2017-04-20
Kubernetes minor versions fact

guineveresaenger:
For AWS this will gather all kubernetes minor versions from all clusters specified in the config file and assign them to a top-level key called kubernetes_minor_version.
You can now look up each cluster's minor version by cluster name as follows:
kubernetes_minor_versions.<clusterName> and it should tell you the minor version.
This partially fulfills #345.
As per #337 the GKE work is still left to do.

### DIFF
--- a/ansible/roles/kraken.config/tasks/main.yaml
+++ b/ansible/roles/kraken.config/tasks/main.yaml
@@ -47,7 +47,7 @@
   with_subelements:
     - "{{ kraken_config.clusters }}"
     - nodePools
-  when: (item.1.apiServerConfig is defined)
+  when: (item.1.apiServerConfig is defined) or (item.0.providerConfig.provider == 'gke')
   vars:
     masterVersion: "{{ item.1.kubeConfig.version }}"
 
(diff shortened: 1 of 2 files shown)


coffeepac:
ok to test

guineveresaenger:
retest this please

guineveresaenger:
@coffeepac is this running a test? also, I added some commits to this existing PR because they were closely related. that should be incorporated automatically, right?

ansible/roles/kraken.config/tasks/main.yaml
```diff
@@ -37,11 +37,26 @@
...
+  set_fact:
+    kubernetes_minor_versions: "{{ kubernetes_minor_versions| default({}) |combine
+      (
+        { item.0.name: masterVersion[0:(masterVersion.find('.', 3 ))] }
+      )
+    }}"
+  with_subelements:
+    - "{{ kraken_config.clusters }}"
+    - nodePools
+  when: (item.1.apiServerConfig is defined) or (item.0.providerConfig.provider == 'gke')
```
coffeepac: I have documented how we are setting this fact:  https://github.com/samsung-cnct/k2/issues/230#issuecomment-296330375

I have a funny feeling this is going to cause some teeth gnashing simply because we've made a decision.  at least this way we'll have a shot of knowing what the funny behavior is coming from.

coffeepac:
LGTM

PR #350, opened 2017-04-20
aws region now set from config

DStorck:
aws region had been set from user's aws cred file, now it grabs the region out of the config


leahnp:
LGTM

Issue #125, opened 2016-12-14
save remote state of config files as a configmapper/etcd/secret?


coffeepac:
k2 should save the stateful file to the hosting provider (or baremetal) so there is a redundant log of where to find state for a specific cluster that isn't on the creating machine.

joejulian:
What about putting this inside the cluster? minio s3 store maybe?

coffeepac:
interesting idea!  if we're using a native cluster resource maybe we could use etcd?  that may be terrible though, I don't know what the size restrictions are on etcd keys.  or size recommendations at least.  I know zk gets cranky if the key data is over 1k.

venezia:
You're safe with etcd.  But why not just (maybe) do it a bit less hacky and say create a k2-management pod that you interact with (and this stores it into etcd)

PR #351, opened 2017-04-20
kube auth

DStorck:
this will mandate users to add a kubeAuth stanza to their configs to be able to bring up a cluster, and remove the creation of kubeAuth info from kraken.config/tasks/setup.yaml

### DIFF
--- a/ansible/roles/kraken.config/files/config.yaml
+++ b/ansible/roles/kraken.config/files/config.yaml
@@ -147,7 +147,7 @@ definitions:
               snapshot_id:
               encrypted: false
     - &defaultAwsEtcdEventsNode
-      name: defaultAwsEtcdEventsNode 
+      name: defaultAwsEtcdEventsNode
       kind: node
       mounts:
         -
@@ -330,6 +330,15 @@ definitions:
       kind: keyPair
       publickeyFile: "$HOME/.ssh/id_rsa.pub"
       privatekeyFile: "$HOME/.ssh/id_rsa"
+  kubeAuth:
+   - &defaultKubeAuth
+      authz: {}
+      authn:
+        basic:
+          -
+            password:
+            user: "admin"
+        default_basic_user: "admin"
 
 # This is the core of the new configuration.
 
@@ -379,6 +388,7 @@ deployment:
           keyPair: *defaultKeyPair
       fabricConfig: *defaultCanalFabric
       helmConfig: *defaultHelm
+      kubeAuth: *defaultKubeAuth
   readiness:
     type: exact
     value: 0
@@ -404,6 +414,7 @@ deployment:
     <<: *defaultCanalFabric
   provider: aws
   providerConfig: *defaultAws
+  kubeAuth: *defaultKubeAuth
   nodepool:
     -
       name: etcd
--- a/ansible/roles/kraken.config/tasks/setup.yaml
+++ b/ansible/roles/kraken.config/tasks/setup.yaml
@@ -14,68 +14,73 @@
   set_fact:
     kraken_config: "{{ kraken_config | combine({'kubernetes_cloudprovider':kraken_config.provider}) }}"
 
-- name: Set empty kubeAuth if not defined
-  set_fact:
-    kraken_config: "{{ kraken_config | combine({'kubeAuth': {}}, recursive=True) }}"
-  when: "cluster.kubeAuth is undefined"
 
-- name: Set default kubernetes basic user if not defined
-  set_fact:
-    kraken_config: "{{ kraken_config | combine(
-    {
-      'kubeAuth': {
-        'authn': {
-          'default_basic_user': 'admin',
-        }
-      }
-    }, recursive=True) }}"
-  when: "cluster.kubeAuth.authn is undefined"
+- name: Verify KubeAuth section exists in config
+  fail:
+    msg: Kube Auth needs to be defined in config
+  when: cluster.kubeAuth not defined
+# - name: Set empty kubeAuth if not defined
+#   set_fact:
+#     kraken_config: "{{ kraken_config | combine({'kubeAuth': {}}, recursive=True) }}"
+#   when: "cluster.kubeAuth is undefined"
 
-- name: Set empty kubernetes authz dic if not defined
-  set_fact:
-    kraken_config: "{{ kraken_config | combine({'kubeAuth': {'authz': { }}}, recursive=True) }}"
-  when: "cluster.kubeAuth.authz is undefined"
+# - name: Set default kubernetes basic user if not defined
+#   set_fact:
+#     kraken_config: "{{ kraken_config | combine(
+#     {
+#       'kubeAuth': {
+#         'authn': {
+#           'default_basic_user': 'admin',
+#         }
+#       }
+#     }, recursive=True) }}"
+#   when: "cluster.kubeAuth.authn is undefined"
+#
+# - name: Set empty kubernetes authz dic if not defined
+#   set_fact:
+#     kraken_config: "{{ kraken_config | combine({'kubeAuth': {'authz': { }}}, recursive=True) }}"
+#   when: "cluster.kubeAuth.authz is undefined"
 
-- name: Generate default kubernetes basic auth if not defined
-  set_fact:
-    kraken_config: "{{ kraken_config | combine(
-    {
-      'kubeAuth': {
-        'authn': {
-          'basic':[
-            {
-              'user': kraken_config.kubeAuth.authn.default_basic_user,
-              'password': lookup('password', config_base + '/' + cluster.name + '/auth/k8spassword-' + kraken_config.kubeAuth.authn.default_basic_user + ' chars=ascii_letters length=11')
-            }
-          ]
-        }
-      }
-    }, recursive=True) }}"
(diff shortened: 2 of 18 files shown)


DStorck:
this PR does not address the set-oidc-info.yaml file

ansible/roles/kraken.config/tasks/setup.yaml
```diff
@@ -12,70 +12,12 @@
...
-  set_fact:
-    kraken_config: "{{ kraken_config | combine(
-    {
-      'kubeAuth': {
-        'authn': {
-          'basic': kube_basic_auths
-        }
-      }
-    }, recursive=True) }}"
+- name: Verify kubeAuth is defined
```
joejulian: Unnecessary because it's now in the default config, it will always exist.

ansible/roles/kraken.config/tasks/setup.yaml
```diff
@@ -91,5 +33,5 @@
 
 - name: Generate random prefix if required
   set_fact:
-    kraken_config: "{{ kraken_config | combine({'resourcePrefix': lookup('password', config_base + '/' + kraken_config.cluster + '/prefix' + ' chars=ascii_letters length=7')}, recursive=True) }}"
+    kraken_config: "{{ kraken_config | combine({'resourcePrefix': lookup('password', config_base + '/' + cluster.name + '/prefix' + ' chars=ascii_letters length=7')}, recursive=True) }}"
```
joejulian: This will replace kraken_config.resourcePrefix for each cluster pass, overwriting it each subsequent time.

ansible/roles/kraken.config/tasks/setup.yaml
```diff
@@ -12,70 +12,12 @@
 
 - name: Set the kubernetes cloud provider to {{ cluster.providerConfig.provider }}
   set_fact:
-    kraken_config: "{{ kraken_config | combine({'kubernetes_cloudprovider':kraken_config.provider}) }}"
+    kraken_config: "{{ kraken_config | combine({'kubernetes_cloudprovider':cluster.providerConfig.provider}) }}"
```
joejulian: This will replace kraken_config.kubernetes_cloudprovider for each cluster, overwriting it with each subsequent cluster.

ansible/roles/kraken.config/tasks/setup.yaml
DStorck: @coffeepac had asked me to put it in there, is there anything in particular we need to validate now?

ansible/roles/kraken.config/tasks/setup.yaml
DStorck: Does it make more sense to change this to be a global variable in kraken_config instead of  a per cluster piece of data?

ansible/roles/kraken.config/tasks/setup.yaml
DStorck: @joejulian Well, it now exists in the default config that will be generated, but if people are using old configs, this will still need a check. In theory down the road once everyone is using the new config it shouldn't be an issue, but it might be during the transition.

ansible/roles/kraken.config/tasks/setup.yaml
DStorck: This is the same type of error we've been running into over and over again with this new configuration, any ideas for how we want to handle it?

ansible/roles/kraken.config/tasks/setup.yaml
DStorck: or are we planning to have the availability of a federated cluster with different clusters running on different cloud providers?

joejulian: I would plan for different cloud providers. Since that info is already under cloud, I'm not even sure why we're doing this. I would rather find the places where this is referenced, and make them just reference the cloud.provider config data directly.

ansible/roles/kraken.config/tasks/setup.yaml
joejulian: On the one hand, we don't keep this data in the config as it is now.

One of our goals is to put the as-configured data into a k/v store on the cluster itself in order to manage that cluster after the initial up from anywhere.

Should we add this to, and export, the merged config or should we continue to just write it to a file? ( @davidewatson @coffeepac @mikeln @venezia )

If we're just writing it to a file, we could just have a top-level dict that has keys for each cluster and values with the prefix. If we're going to write it to the config, it'll need to go into each cluster list entry (which is a lot more complicated).

ansible/roles/kraken.config/tasks/setup.yaml
DStorck: yeah if this info is easily accessible via the config file, it seems like a cleaner way to grab that data

DStorck: I only see it being used here: https://github.com/samsung-cnct/k2/blob/k2recon/ansible/roles/kraken.nodePool/kraken.master.docker/templates/api-server-manifest.yaml.jinja2

ansible/roles/kraken.config/tasks/setup.yaml
mikeln: reading...

mikeln: write to config... it's what we really want.

ansible/roles/kraken.config/tasks/setup.yaml
mikeln: I agree with Joe on this one.

DStorck: Okay I am going to remove this from kraken_config, and have it pull in the cloudprovider directly in the api-server-manifest template, which is the only place i see it being referenced

ansible/roles/kraken.config/tasks/setup.yaml
coffeepac: config is what we want but I'd like to have this ticket closed by lunchish on tuesday.  I'd prefer the following:
 - implement writing to a new top level dict so we can maintain current functionality
 - take a shot at merging this back into the cluster list, failing that, file a new issue to do this as part of #125

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/main.yaml
```diff
@@ -11,7 +11,6 @@
     profile: "{{ cluster.providerConfig.authentication.credentialsProfile or omit }}"
     filters:
       "tag:k2-nodepool": "{{ item.name }}"
-      "tag:Name": "{{ cluster.name }}_{{ kraken_config.resourcePrefix }}_{{ item.name }}*"
```
joejulian: Let's still give it a name tag, but just build the string without the resourcePrefix bit in the middle.

DStorck:
Generate random prefix has been removed as per a slack discussion about its purpose ( check of cluster-name availability / desire to make sure we are not creating clusters with duplicate names...) 
This issue will hopefully be dealt with in https://github.com/samsung-cnct/k2/issues/353

joejulian:
LGTM

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/main.yaml
DStorck: done

ansible/roles/kraken.config/files/config.yaml
```diff
@@ -330,6 +330,15 @@ definitions:
       kind: keyPair
       publickeyFile: "$HOME/.ssh/id_rsa.pub"
       privatekeyFile: "$HOME/.ssh/id_rsa"
+  kubeAuth:
+   - &defaultKubeAuth
+      authz: {}
+      authn:
+        basic:
+          -
+            password:
```
coffeepac: we have removed the parts of the system that will set a default password, should we set it here?

DStorck: The password was being in kraken.config/tasks/main.yaml for reference, but can't be set in the same way with the new config: 
```- name: Generate default kubernetes basic auth if not defined
  set_fact:
    kraken_config: "{{ kraken_config | combine(
    {
      'kubeAuth': {
        'authn': {
          'basic':[
            {
              'user': kraken_config.kubeAuth.authn.default_basic_user,
              'password': lookup('password', config_base + '/' + kraken_config.cluster + '/auth/k8spassword-' + kraken_config.kubeAuth.authn.default_basic_user + ' chars=ascii_letters length=11')
            }
          ]
        }
      }
    }, recursive=True) }}"```

DStorck: I can try Joe's way of looping through and to add this info to kraken_config, or is there a better way to set this password?

joejulian: If we're going to eventually remotely manage this cluster, we will need access to this password. Since I suspect to make this happen we're going to need to write an "as-built" config file to our storage system, this is going to need to be in the config.

That would require looping through to add it.

coffeepac: this is more then I was looking for.  I meant:  should the default config file have a default admin password.  I think it should because otherwise I'm not real sure what will happen.

For now I'd be happy setting it to ChangeMe.

DStorck:
don't merge yet, work to be done to torch all the references to ${var.prefix}

DStorck:
okay, references to prefix are gone, I was able to get a cluster up and down with the current commit.

PR #359, opened 2017-04-26
k2-283 DNS separation

mikeln:
Split DNS into a direct (non-helm service).

### DIFF
--- a/Documentation/kraken-configs/clusterservices.md
+++ b/Documentation/kraken-configs/clusterservices.md
@@ -37,7 +37,7 @@ clusterServices:
         url: http://atlas.cnct.io
     services:
       -
-        name: heapseter
+        name: heapster
         repo: atlas
         chart: heapster
         version: 0.1.0
--- a/Documentation/kraken-configs/clusterservices.md
+++ b/Documentation/kraken-configs/clusterservices.md
@@ -37,13 +37,10 @@ clusterServices:
         url: http://atlas.cnct.io
     services:
       -
-        name: kubedns
+        name: heapseter
         repo: atlas
-        chart: kubedns
+        chart: heapster
         version: 0.1.0
         namespace: kube-system
-        values:
-          cluster_ip: 10.32.0.2
-          dns_domain: krakenCluster.local
 
 ```
--- a/ansible/roles/kraken.services/tasks/kill-dns.yaml
+++ b/ansible/roles/kraken.services/tasks/kill-dns.yaml
@@ -1,6 +1,16 @@
 ---
 - set_fact:
+    cluster: "{{ a_cluster }}"
+
+- set_fact:
     kubeconfig: "{{ config_base | expanduser }}/{{ cluster.name }}/admin.kubeconfig"
+
+- name: Merge User DNS Config with Default DNS Config
+  set_fact:
+    dns_config: "{{ kubedns | combine(cluster.dnsConfig.kubedns) | expand_config }}"
+
+- debug:
+    var: dns_config.namespace
 
 - name: Clean up DNS
   command: >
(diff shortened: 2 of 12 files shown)


Documentation/kraken-configs/clusterservices.md
```diff
@@ -37,13 +37,10 @@ clusterServices:
         url: http://atlas.cnct.io
     services:
       -
-        name: kubedns
+        name: heapseter
```
DStorck: typo

ansible/roles/kraken.services/tasks/main.yaml
```diff
@@ -2,6 +2,28 @@
 - include: generate-oidc-cert.yaml
   when: (kraken_config.kubeAuth.authn.oidc is defined) and (kraken_action == 'up')
 
+- name: Merge User DNS Config with Default DNS Config
+  set_fact:
+    dns_config: "{{ kubedns | combine(kraken_config.clusterServices.kubedns) | expand_config }}"
```
DStorck: kraken_config.clusterServices has changed to cluster.helmConfig and may not be available in the way we want from main.yaml

ansible/roles/kraken.services/tasks/kill-dns.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- set_fact:
+    kubeconfig: "{{ config_base | expanduser }}/{{ cluster.name }}/admin.kubeconfig"
```
DStorck: where is `cluster` being set?

Documentation/kraken-configs/clusterservices.md
mikeln: definitely...

ansible/roles/kraken.services/tasks/main.yaml
mikeln: This whole thing is about pulling DNS out of helm, but understand what you are saying.  I just pulled over my original changes in the location where they were in config.yaml before.  The configs should probably be moved up under `definitions: dnsconfigs:`

ansible/roles/kraken.services/tasks/kill-dns.yaml
mikeln: good question... I threw this together late.  let me re-review what is suppose to happen.

DStorck: if `a_cluster` is being set in main.yaml, you can just set another fact in this file at the top: 
```- set_fact:
    cluster: "{{ a_cluster }}"```

File: ansible/roles/kraken.services/tasks/main.yaml (hunk 2 above)
Comment (mikeln) [inline ansible/roles/kraken.services/tasks/main.yaml]: or actually not.  need to expand/convert `clusters:dns:`  and `clusters:domain` to 
```
clusters:
  dnsconfig:
    cluster_ip: <ip>                           
    dns_domain: <domain>
    namespace: <ns>

```

ansible/roles/kraken.services/tasks/kill-dns.yaml
leahnp: Why are we using the variable `a_cluster` (vs simply `cluster`) in some places?

DStorck: a_cluster is being set in main instead of cluster_node_tuple when we don't need nodepool, then cluster gets set inside each file that needs to access the actual info

mikeln:
NOTE: The documentation files need a total rewrite for recon config changes.  It was easier to just leave them in this PR for now as is.

mikeln:
this will need some additional changes to incorporate the kubectl versioning after https://github.com/samsung-cnct/k2/pull/363 lands.

coffeepac:
LGTM

PR #363, opened 2017-04-26
Kubectl for all versions and all clusters

guineveresaenger:
This sets a kubectl fact in kraken.services/tasks/kill-services and kraken.services/tasks/run-services so that each cluster will use the appropriate version of kubectl.

### DIFF
--- a/ansible/roles/kraken.fabric/kraken.fabric.flannel/tasks/deploy-canal.yaml
+++ b/ansible/roles/kraken.fabric/kraken.fabric.flannel/tasks/deploy-canal.yaml
@@ -1,4 +1,12 @@
 ---
+- name: Look up and set k8s minor version for this cluster
+  set_fact:
+    kubernetes_minor_version: "{{ kubernetes_minor_versions[cluster.name] }}"
+
+- name: Execute appropriate kubectl per minor version
+  set_fact:
+    kubectl: "/opt/cnct/kubernetes/{{ kubernetes_minor_version }}/bin/kubectl"
+
 - name: Ensuring fabric directory exists
   shell: >
     mkdir -p {{ config_base | expanduser }}/{{ cluster.name }}/fabric
@@ -21,14 +29,14 @@
 
 - name: check kube-networking namespace state
   command: >
-    kubectl --kubeconfig={{ kubeconfig | expanduser }} get namespace kube-networking
+    {{ kubectl }} --kubeconfig={{ kubeconfig | expanduser }} get namespace kube-networking
   ignore_errors: yes
   changed_when: false
   register: kube_networking_namespace
 
 - name: Ensure the kube-networking namespace exists
   command: >
-    kubectl --kubeconfig={{ kubeconfig | expanduser }} create namespace kube-networking
+    {{ kubectl }} --kubeconfig={{ kubeconfig | expanduser }} create namespace kube-networking
   register: result
   retries: 120
   until: result|succeeded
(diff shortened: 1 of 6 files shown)


guineveresaenger:
I didn't really see a good way of setting kubectl only once (in main) since we loop over the cluster list individually for both changed files.

mikeln:
So far so good.  But shouldn't ALL reference to `kubectl`  be converted to use `{{ kubectl }}` and the setup.   There are more than just `kraken.services`.

```
kraken.fabric/kraken.fabric.flannel/tasks/deploy-canal.yaml:    kubectl --kubeconfig={{ kubeconfig | expanduser }} get namespace kube-networking
kraken.fabric/kraken.fabric.flannel/tasks/deploy-canal.yaml:    kubectl --kubeconfig={{ kubeconfig | expanduser }} create namespace kube-networking
kraken.fabric/kraken.fabric.flannel/tasks/deploy-canal.yaml:    kubectl --kubeconfig={{ kubeconfig | expanduser }} get namespace kube-networking
kraken.fabric/kraken.fabric.flannel/tasks/deploy-canal.yaml:    kubectl --kubeconfig={{ kubeconfig | expanduser }} apply -f {{ config_base | expanduser }}/{{ cluster.name }}/fabric/config.yaml
kraken.fabric/kraken.fabric.flannel/tasks/deploy-canal.yaml:    kubectl --kubeconfig={{ kubeconfig | expanduser }} apply -f {{ config_base | expanduser }}/{{ cluster.name }}/fabric/canal.yaml
kraken.fabric/kraken.fabric.flannel/templates/canal.yaml.part.jinja2:            # Disable file logging so `kubectl logs` works.
kraken.readiness/tasks/bootstrap-default-rbac-policy.yaml:  shell: "kubectl --kubeconfig={{ kubeconfig | expanduser }} get node --no-headers=true | grep -i schedulingdisabled"
kraken.readiness/tasks/bootstrap-default-rbac-policy.yaml:    kubectl --kubeconfig={{ kubeconfig | expanduser }} delete -f {{ config_base | expanduser }}/{{ cluster.name }}/auth/rbac-default-policy.yaml
kraken.readiness/tasks/bootstrap-default-rbac-policy.yaml:    kubectl --kubeconfig={{ kubeconfig | expanduser }} create -f {{ config_base | expanduser }}/{{ cluster.name }}/auth/rbac-default-policy.yaml
kraken.readiness/tasks/do-wait.yaml:    kubectl --kubeconfig={{ kubeconfig | expanduser }} get nodes --no-headers=true
kraken.readiness/tasks/generate-admin-kubeconfig.yaml:    kubectl config set-cluster {{ cluster.name }}
kraken.readiness/tasks/generate-admin-kubeconfig.yaml:    kubectl config set-credentials {{ cluster.name }}_admin
kraken.readiness/tasks/generate-admin-kubeconfig.yaml:    kubectl config set-context {{ cluster.name }}_context
kraken.readiness/tasks/generate-admin-kubeconfig.yaml:    kubectl config use-context {{ cluster.name }}_context
```

guineveresaenger:
@mikeln I'm about to make a comment on the ticket, which specifically notes kraken.services. But I would also like to find this out.

mikeln:
Well, we are setting the version for each cluster name... so we are always going to have to set the kubectl executable path per cluster.   I.e. it will always be pushed down to the lowest task level.
However, isn't there a way to setup a higher level "subroutine" that we could call to set kubectl fullpath from anywhere... (vs copying the block of code into each low level task)?  (I'm not sure if Ansible has this type of thing?).

guineveresaenger:
@mikeln it does need to be associated with each individual cluster so perhaps I can add it to the loop variable in main. let me see.

coffeepac:
ok to test

guineveresaenger:
As per conversation with @mikeln we should look for a way to refactor having to set minor version in multiple files. But this works for now.

coffeepac:
LGTM - going to fix the few listed additional places with another PR

PR #385, opened 2017-05-03
update defaults for central-logging

DStorck:
the inclusion by default of the central-logging repo means that several specs needed to get updated. We needed more clusterNodes slightly different instance types to handle logging


joejulian:
LGTM but would like @coffeepac to give final approval.

PR #387, opened 2017-05-03
Add ability to use non-HOME as base directory

mikeln:
This adds a `-r` argument to specify the root config directory.  
If it is not specified, then $HOME will be used.


mikeln:
This is for Development Environment only.  It does not impact any running systems.

coffeepac:
LGTM

PR #390, opened 2017-05-05
Checks for availability of helm by version and will create option to override

guineveresaenger:
This is not ready to merge (and apparently I have some rebasing to do...) but here is how I'm currently handling the helm version inside of k2:

- in kraken.config, there is a new file that will set the helm availability to true or false.
- based on this, each cluster gets an option to override.
- both override and availability get saved in a fact and referenced by cluster name (this could get refactores or the logic could live elsewhere; helm_availability as its own fact seems somewhat wordy if we can just check the minor version instead)
- in run_services, we set the helm_command accordingly.
- currently, helm_command is not being set if helm is unvailable for this version and the override is false. I'm trying to see what this will do but aws is not cooperating this afternoon.


coffeepac:
ok to test

ansible/roles/kraken.services/tasks/run-services.yaml
```diff
@@ -34,16 +39,11 @@
...
   when: cluster_namespaces is defined
 
-- name: create helm command string
+- name: create debug helm command string
   set_fact:
-    helm_command: "helm --debug"
+    helm_command: "{{ helm_command }} --debug"
   when: (cluster.helmConfig.debug is defined) and (cluster.helmConfig.debug == True)
 
-- name: create helm command string
```
guineveresaenger: Can someone speak to what cluster.helmConfig.debug does, and why we need it as a condition?

ansible/roles/kraken.config/tasks/helm-versioning.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+- name: Set helm override env variable key to contain the cluster name
+  set_fact:
+    helm_override_key: "HELM_OVERRIDE_{{cluster.name}}"
+
+- name: Get helm override environment variable
+  set_fact:
+    HELM_OVERRIDE: "{{ lookup('env', helm_override_key) }}"
+
+- name: Set helm_available to true for k8s 1.4 and 1.5
```
guineveresaenger: I should add v1.6 to the name of the Task, since 1.6 is now supported.

ansible/roles/kraken.config/tasks/helm-versioning.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+- name: Stop and print out message if helm is not available and HELM_OVERRIDE has not been set
+  fail:
+    msg: "This version of Kubernetes does not currently have helm support. If you would like to continue without using helm, execute following command:
+    export {{ helm_override_key }}=false && ./up.sh
+    If you would like to use the latest version of helm instead, execute:
+    export {{ helm_override_key }}=true && ./up.sh
+    WARNING: You may have to specify your config file path if it is not located in the default path."
+  when: helm_available == false and HELM_OVERRIDE == ""
+
```
guineveresaenger: 1. Does it make sense to capitalize HELM_OVERRIDE since we'll be manually setting an env variable for each cluster not having helm support? It felt nonidiomatic, so for the collection of overrides below I did not capitalize.
2. It is frustrating to attempt to pretty print an ansible error message. It is possible to hack something with `debug` but `fail` parses information differently. I can spend more time on this if deemed necessary.

ansible/roles/kraken.services/tasks/run-services.yaml
```diff
@@ -42,16 +48,19 @@
   set_fact:
     helm_command: "{{ helm_command }} --debug"
   when: (cluster.helmConfig.debug is defined) and (cluster.helmConfig.debug == True)
+  ignore_errors: yes
```
guineveresaenger: I am not certain if `ignore_errors: yes ` is the correct way to handle this. Basically, no helm commands will be executed in case of override being set to false, but I wonder if there should be a more graceful way of handling this, perhaps by adding (yet another) when contition to all helm_command tasks.

ansible/roles/kraken.config/tasks/helm-versioning.yaml
coffeepac: this should be programmatically determined by checking the filesystem.

ansible/roles/kraken.config/tasks/helm-versioning.yaml
```diff
@@ -0,0 +1,50 @@
...
+- name: Set helm_available to true for k8s 1.4 and 1.5
+  set_fact:
+    helm_available: true
+  when: (kubernetes_minor_version == 'v1.4') or (kubernetes_minor_version == 'v1.5') or (kubernetes_minor_version == 'v1.6')
+
+- name: Set helm_available false for unsupported k8s version
+  set_fact:
+    helm_available: false
+  when: kubernetes_minor_version == 'v1.7'
```
coffeepac: this should also be programmatically determined.  both this as the available fact should be based on the current kubernetes_minor_version

ansible/roles/kraken.config/tasks/helm-versioning.yaml
coffeepac: 1. I would say it doesn't make sense to capitalize HELM_OVERRIDE.  Its an ansible fact, not an environment variable
2. only spend the time if you want to.  error messages don't have to be pretty, they only have to be informative

ansible/roles/kraken.services/tasks/kill-services.yaml
```diff
@@ -10,6 +10,13 @@
 - name: Execute appropriate helm per minor version
   set_fact:
     helm_command: "/opt/cnct/kubernetes/{{ kubernetes_minor_version }}/bin/helm"
+  when: helm_availabilities[cluster.name] == true
+
+
+- name: In case of helm not being available, if override was specified, override to k8s 1.5
+  set_fact:
+    helm_command: "/opt/cnct/kubernetes/v.1.5/bin/helm"
```
coffeepac: this should be programmatically determined to be the highest version of helm available.

ansible/roles/kraken.services/tasks/run-services.yaml
```diff
@@ -14,6 +14,12 @@
 - name: Execute appropriate helm per minor version
   set_fact:
     helm_command: "/opt/cnct/kubernetes/{{ kubernetes_minor_version }}/bin/helm"
+  when: helm_availabilities[cluster.name] == true
+
+- name: In case of helm not being available, if override was specified, override to k8s 1.5
+  set_fact:
+    helm_command: "/opt/cnct/kubernetes/v1.5/bin/helm"
```
coffeepac: again, shouldn't be hardcoded.

ansible/roles/kraken.services/tasks/run-services.yaml
coffeepac: these should be skipped with a when. there are many reasons a helm command could fail and we don't want to suppress them.  we don't want to skip them if we know the helm binary doesn't exist.

ansible/roles/kraken.services/tasks/kill-services.yaml
guineveresaenger: Is there an idiomatic way to achieve this? Currently, all I can think of is:
Option 1:
- get the final character from the minor version, transform it into an int, subtract 1, and reappend to monir version (after last char has been removed). This is buggy because what if helm isn't just one minor version behind?
Option 2:
- somehow loop through all minor versions in the file system and determine which is the highest, using math on strings again. *Also* buggy, because just because that subdirectory (k82 minor version) exists, does not mean a helm lives in it.
Option 3:
Something magical I don't know about? Something really obvious I can't seem to think of?

joejulian:
retest this please

ansible/roles/kraken.services/tasks/main.yaml
```diff
@@ -1,4 +1,15 @@
...
+  find:
+    paths: "/opt/cnct/kubernetes"
+    patterns: "helm"
+    recurse: yes
+  register: helm_option_paths
+
+- name: Get latest helm version
+  set_fact:
+    latest_helm: "{{ helm_option_paths | latest_helm }}"
+
```
guineveresaenger: I put this in main to avoid duplication of code. The latest helm is always the latest helm, whether we need it or not. This does mean these tasks will run every single time for every cluster.

ansible/roles/kraken.config/tasks/helm-versioning.yaml
```diff
@@ -0,0 +1,48 @@
...
+    }}"
+  when: helm_override != ""
+
+- name: Keep track of helm availability by cluster name
+  set_fact:
+    helm_availabilities: "{{ helm_availabilities | default({}) | combine
+      (
+        {cluster.name: helm_available }
+      )
+    }}"
```
guineveresaenger: On cluster down, this task will run, which may break this task for unsupported helm on down? This seems suboptimal.

ansible/roles/kraken.services/tasks/run-services.yaml
```diff
@@ -14,6 +14,12 @@
 - name: Execute appropriate helm per minor version
   set_fact:
     helm_command: "/opt/cnct/kubernetes/{{ kubernetes_minor_version }}/bin/helm"
+  when: helm_availabilities[cluster.name] == true
+
+- name: In case of helm not being available, if override was specified, override to override to latest helm
```
coffeepac: nit:  should be 'overide to latest helm'

ansible/roles/kraken.config/tasks/helm-versioning.yaml
coffeepac: nah, that's fine.  helm gets used during cluster down to clean up releases.

ansible/roles/kraken.services/tasks/main.yaml
coffeepac: that's fine.  its not much, maybe file a p3 ticket to extract this into something that runs once total

guineveresaenger: ^^ oh whoops, sorry. I meant every kind of cluster regardless of k8s minor version. Not each cluster. It should run only once per k2, but will run even if we don't need to find the latest helm.

ansible/roles/kraken.services/filter_plugins/latest_helm.py
```diff
@@ -0,0 +1,25 @@
...
+    max_version = versions[0]
+
+    for version in versions:
+        if StrictVersion(max_version) < StrictVersion(version):
+            max_version = version
+
+    return max_version
+
+class FilterModule(object):
+    ''' Unicorn filter that seems to be just a name for now so fuzzy shiny unicorns ahoy '''
```
DStorck: this is a docstring, and is what gets printed out as the __doc__ attribute for the filter. It should be a concise description of what the filter is doing

guineveresaenger: Thank you for catching this.
Where exactly do doc attributes live and what do they do?

PR #396, opened 2017-05-08
Jenkins 2.0

coffeepac:
Add support to K2 for running all CI in a Jenkins-2.0 project and bring all CI dependencies into K2 except for K2-tools


build-scripts/e2etester-docker/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+      dnsutils \
+    && rm -rf /var/lib/apt/lists/*
+
+# All python modules should be installed here if possible
+COPY requirements.txt /opt/kraken-ci/containers/jenkins/
+RUN pip install --upgrade pip \
+  && pip install --requirement /opt/kraken-ci/containers/jenkins/requirements.txt
+
+## Install Google Cloud SDK
+RUN wget https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-${GOOGLE_CLOUD_SDK_VERSION}-linux-x86_64.tar.gz \
```
davidewatson: I realize images may not be built so often that corruption will be a common problem, but I would still feel more comfortable if we included a hash check anytime we download a binary. In this case the untar would probably fail, so the error would be easily discoverable, but why risk it? The additional maintenance required to include the hash each time the binary version changes is small, and occurs infrequently. Also, without hashes this is insecure.

davidewatson:
Your commit messages read like a short story. They're great! :) Generally LGTM, though I dislike that we are still including configuration files in Jenkins instead of generating them using k2(cli). Doing this has historically lead to the tested configuration file becoming out of sync with the current default configuration file.

coffeepac:
@davidewatson config files:  when this PR started there wasn't a way to generate the GKE config out of k2.  We are using `up.sh --generate` for the aws config.

build-scripts/e2etester-docker/Dockerfile
coffeepac: #399 handles this.

Issue #375, opened 2017-04-27
check for unused loop variables in the following files and locations


coffeepac:
ansible/roles/kraken.config/tasks/main.yaml line 57

ansible/roles/kraken.config/tasks/setup.yaml line 4

ansible/roles/kraken.nodePool/kraken.master.docker/tasks/service-account-key.yaml line 4

DStorck:
I'm assuming line 57 was the reference to the include for setup.yaml, recent merges have changed that file a bunch

coffeepac:
work is complete.

PR #419, opened 2017-05-12
removed non-used loops vars

DStorck:
this removes unused variables mentioned in https://github.com/samsung-cnct/k2/issues/375

### DIFF
--- a/ansible/roles/kraken.nodePool/kraken.master.docker/tasks/service-account-key.yaml
+++ b/ansible/roles/kraken.nodePool/kraken.master.docker/tasks/service-account-key.yaml
@@ -1,7 +1,6 @@
 ---
 - set_fact:
     cluster: "{{ cluster_node_tuple.0 }}"
-    node: "{{ cluster_node_tuple.1 }}"
 
 - name: Make sure generated folder for generated key is there
   file:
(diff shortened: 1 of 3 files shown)


ansible/roles/kraken.nodePool/kraken.master.docker/tasks/service-account-key.yaml
```diff
@@ -1,7 +1,6 @@
 ---
 - set_fact:
```
joejulian: Should this be `a_cluster`, too?

coffeepac: shouldn't be.  the rest of the file refers to 'cluster'

'a_cluster' is an oddity I haven't tracked down yet but there is something somewhere that is already using the variable cluster so if you try and use it when looping over cluster ansible throws a duplicate variable message.

coffeepac:
LGTM

coffeepac:
retest this please

coffeepac:
retest this please

coffeepac:
passed pipelet.  good enough for now as I don't expect the new jenkins system to pass this.

Issue #263, opened 2017-03-15
Validate format of properties


davidewatson:
# Task

The k2 schema defines a format for some properties. Some of these formats (e.g. "ip4addr") are part of the JSON Schema standard and some implementations should implement. The existing JSON schema validator should be extended to validate non standard formats (e.g. cidr, semver). Until this is done non standard formats should be ignored by all validators.

coffeepac:
blocked by json schema tickets

PR #422, opened 2017-05-13
all letters need be lowercase because of GKE

coffeepac:
GKE only accepts lower case letters in the cluster names.  AWS is fine either way so we're going to constrain to GKE levels.


davidewatson:
LGTM. Note that we might want to add checks for this in the schema as part of #263.

PR #424, opened 2017-05-14
allow e2e tests to succeed

coffeepac:
the 'tee' portion in a command (and subsequent workarounds) were causing all runs to fail.  I believe the PIPESTATUS workaround is not available in the sh shell.  we no longer need to save off the build-log output file.

removal of this guff has produced at least one e2e successful run.  This PR should not be merged until the new jenkins job passes.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 podTemplate(label: 'k2', containers: [
     containerTemplate(name: 'jnlp', image: 'jenkinsci/jnlp-slave:2.62-alpine', args: '${computer.jnlpmac} ${computer.name}'),
     containerTemplate(name: 'k2-tools', image: 'quay.io/samsung_cnct/k2-tools:latest', ttyEnabled: true, command: 'cat'),
-    containerTemplate(name: 'e2e-tester', image: 'quay.io/samsung_cnct/e2etester:0.1', ttyEnabled: true, command: 'cat'),
+    containerTemplate(name: 'e2e-tester', image: 'quay.io/samsung_cnct/e2etester:0.2', ttyEnabled: true, command: 'cat'),
     containerTemplate(name: 'docker', image: 'docker', command: 'cat', ttyEnabled: true)
   ], volumes: [
     hostPathVolume(hostPath: '/var/run/docker.sock', mountPath: '/var/run/docker.sock'),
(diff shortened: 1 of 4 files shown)


NathanZook:
Not understanding the original motivation for this code, might the pipefail option address the problem?

spiffxp:
Original motivation: `build-log.txt` was an expected artifact for e2e testing if you wanted to submit results for inclusion in test-grid.k8s.io (ref: https://github.com/kubernetes/test-infra/blob/master/docs/federated_testing.md).  I'm not sure if it is today, but it's still produced by google's e2e testing at the very least.

It is expected to contain logs of cluster build and running tests (and maybe cluster teardown, I forget now.)  The `tee` was a way of allowing a user to see the log in progress through the jenkins ui, while allowing the in-progress job (rather than a separately triggered job) to upload the artifact.

The pipestatus workaround was for that the fact that if tests failed, and thus exited with a >0 exit code, I expected tee to propagate that along.  It did not.  So test failures were incorrectly registering as "passing" jenkins jobs

NathanZook:
I believe that the pipefail option (in the bash manual under Pipelines) will address this.  It makes the return of the pipe the rightmost non-zero return code, or zero if all pass.

coffeepac:
ahh, that makes sense @spiffxp I should have guessed that was why it was being created.  However, we don't have any immediate plans for submitting back to the test grid.

@NathanZook sounds like you're volunteering to handle this when/if we get back to it 😄 

as it is, the AWS portion of this job is succeeding occasionally but I'm having issues getting the GKE part to work.

coffeepac:
this branch, but not this PR:  https://common-jenkins.kubeme.io/job/k2-trial/job/remove-tee/

so this does work, lemme see if I can get the PR job to succeed.

spiffxp:
@NathanZook awesome, TIL `set -o pipefail`, that totally would have done the same thing

NathanZook:
I just hate giving up log files...

coffeepac:
the logfile is preserved in the jenkins job history.  this file gets destroyed when the pod exits.

coffeepac:
we now have a green, passing test from the new CI system.  This meets my criteria for merging this.

davidewatson:
Even if we intend to contribute to Federated Testing in the future, this gets us going. LGTM.

PR #431, opened 2017-05-18
Update README.md, mostly to use ${CLUSTER}.

NathanZook:
Replaced YOURCLUSTER, foo, and <your_config> in example command lines with ${CLUSTER}.
Left YOURCLUSTER for examples inside the config file.
Updated helm output and usage, especially for search and install of kubernetes-dashboard.
Added instructions to bring the cluster up prior to the examples for accessing the cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ deployment:
     - name: YOURCLUSTER
 ```
 
-Should should then rename the `config.yaml` file to `YOURCLUSTER.yaml`.  This is best practice.
+You should then rename the `config.yaml` file to `YOURCLUSTER.yaml`.  This is best practice.
 
 **For the rest of the discussion, we will assume that the environmental variable `${CLUSTER}` has been set to the name of your cluster.**
 
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ This will generate a config.yaml file located at
 ```
 ${KRAKEN}/config.yaml
 ```
-In discussions that follow, the variable `YOURCLUSTER` refers to the name you must assign to your cluster at the bottom of the generated `config.yaml` in the deployments section, which once a name has been decided would look like:
+In this section, the variable `YOURCLUSTER` refers to the name you must assign to your cluster at the bottom of the generated `config.yaml` in the deployments section, which once a name has been decided would look like:
 
 ```
 deployment:
@@ -118,11 +118,14 @@ deployment:
     - name: YOURCLUSTER
 ```
 
-Finally, we suggest you rename the `config.yaml` file to something like
+We suggest you rename the `config.yaml` file to something like
 
 ```
 YOURCLUSTER.yaml
 ```
+
+**For the rest of the discussion, we will assume that the environmental variable `${CLUSTER}` has been set to the name of your cluster**
+
 
 It is particularly useful when trying to create and manage multiple clusters, each of which
 **must** have unique names.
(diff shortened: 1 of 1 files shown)


davidewatson:
ok to test

README.md
```diff
@@ -145,47 +148,56 @@ using the local awscli tool:
...
+To bring your cluster up, run:
+```bash
+docker run $K2OPTS quay.io/samsung_cnct/k2:latest ./up.sh --config $HOME/.kraken/${CLUSTER}.yaml
+```
+
+This will take a while, and will generate a lot of output.
+
 ### kubectl
 
-After creating a cluster, to use the kubectl shipped with K2, run a command similar to:
```
joejulian: Although I agree in principal with this change, doing so should also change the syntax of the next command since we are now less clear that we're only referencing the `--kubeconfig` switch and data. Perhaps this needs rephrased completely.

Otherwise this effectively says you can only run "get nodes" which may seem absurd but if we're going for accurate, let's be accurate.

README.md
```diff
@@ -221,8 +233,6 @@ Then the ssh hostnames available will be:
 
 Earlier, you copied a sample cluster configuration over into `~/.kraken`.  Please take a moment to review the sample configuration and make changes if you desire
 
-You may prefer to save it with a name that is consistent with the `cluster` variable in the configuration. In other words, if your `cluster` is `foo`, then perhaps your file should be named `foo.yaml`
```
joejulian: This is a best practice statement and should remain.

coffeepac:
LGTM

leahnp:
LGTM

README.md
```diff
@@ -302,64 +313,64 @@ You can try having helm install a new service, such as the Kubernetes dashboard
...
 The chart has been installed. It will take a moment for AWS ELB DNS to propagate, but you can get the DNS now.
 
 ##### Finding DNS name for Kubernetes Dashboard
 
 ```bash
-docker run $K2OPTS quay.io/samsung_cnct/k2:latest kubectl --kubeconfig ~/.kraken/foo/admin.kubeconfig describe service kubernetes-dashboard --namespace kube-system
+docker run $K2OPTS quay.io/samsung_cnct/k2:latest kubectl --kubeconfig ~/.kraken/${CLUSTER}/admin.kubeconfig describe service kubernetes-dashboard --namespace kube-system
 ```
 
 ```bash
```
joejulian: These details will be different every time. There's no need to change them. If you'd like to add clarification to that effect, that may be useful.

README.md
NathanZook: The problem here is with the structure of the document as a whole.  The issue of the name of the config file was addressed further up in the document.  Since by this point, the use of ${CLUSTER} is established, and continues, I thought it best to remove this statement.

README.md
NathanZook: The main issue is that the example cannot be cut & pasted.  This has been corrected by replacing "foo" with "${CLUSTER}".
What other details will change?

README.md
joejulian: Ah yes, let's make that statement more definitive, stating it as a best practice.

README.md
joejulian: Lines 344 through 354

joejulian: Those are just results and they'll never match verbatim.

joejulian: Just to clarify my thoughts further, I'm not saying "change this back". I'm just suggesting that if you found it to be a confusing portion of the document, perhaps a sentence of clarification may be beneficial.

NathanZook: Yes.  I did the replacement because there was a change in the form of the output.  I expect that most people don't alert when the order of items changes, but I did.  There are in fact several minor formatting changes here.  Certainly, the need to update the expected output here is less than at other locations, but it seemed worthwhile to go ahead & change this as well.

README.md
```diff
@@ -110,23 +110,39 @@ This will generate a config.yaml file located at
...
+In this section, the variable `YOURCLUSTER` refers to the name you must assign to your cluster at the bottom of the generated `config.yaml` in the deployments section, which once a name has been decided would look like:
 
 ```
 deployment:
   clusters:
     - name: YOURCLUSTER
 ```
 
-Finally, we suggest you rename the `config.yaml` file to something like
+Should should then rename the `config.yaml` file to `YOURCLUSTER.yaml`.  This is best practice.
```
joejulian: :point_up:

Issue #268, opened 2017-03-20
K2-Down fails to bring down clusters with ELBs


leahnp:
Currently k2-down is failing when we deploy an ELB with an external IP.

I can spin up a default cluster and bring it down successfully, but when I deploy the above service, k2 down fails at certain tasks (seem to be a bit different each time, see the above comment for a list of common ones I see).

**Bug repro:** 
1. Spin up default k2 cluster
2. Deploy service:  https://gist.github.com/leahnp/7240e1fc4c2a73a75d0403f734e8555a
3. Run k2-down (it should fail at some point)

**More details** on the issue here: https://github.com/samsung-cnct/kraken-ci-jobs/issues/166

DStorck:
I manually deleted all resources off of the AWS console, but still couldn't bring down cluster all the way. It got stuck here and stopped  : 
```
TASK [/kraken/ansible/roles/kraken.provider/kraken.provider.aws : Run terraform destroy] ***
FAILED - RETRYING: TASK: /kraken/ansible/roles/kraken.provider/kraken.provider.aws : Run terraform destroy (10 retries left).
FAILED - RETRYING: TASK: /kraken/ansible/roles/kraken.provider/kraken.provider.aws : Run terraform destroy (9 retries left).
FAILED - RETRYING: TASK: /kraken/ansible/roles/kraken.provider/kraken.provider.aws : Run terraform destroy (8 retries left).```

**Comment by coffeepac** [comment]
@DStorck when you start seeing those lines its important to let them finish as we don't get the STDERR from the ansible run until all 10 tries have failed.  Its Yet Another Dumb Wait but that's where we're stuck.  If you do see this, please wait and dump the error into here.

**Comment by DStorck** [comment]
@coffeepac It is stopping on its own every time.

**Comment by coffeepac** [comment]
Is it stopping and then showing a fresh command line?

**Comment by DStorck** [comment]
Yes. When I scrolled up, this was sitting at the top: 
```
/k2cli cluster down
Using k2 config file: /Users/deirdrestorck/.kraken/config.yaml
Pulling image 'quay.io/samsung_cnct/k2:latest' ███▒▒▒▒▒▒▒
Bringing down cluster 'ds-march20'█▒▒▒▒▒▒▒▒▒ Action timed out!

ERROR bringing down ds-march20
```

guineveresaenger:
Here's what I got today (from a cluster that never ran up correctly). The action stopped on its own, I didn't exit the process myself.

[...]

Issue #372, opened 2017-04-27
K2 Up Services Ignored Errors


mikeln:
The are 2 ignored errors on K2 services up.  They are valid, but there should be a way to do the needed processing without generating errors that can be ignored.

e.g.:

```TASK [roles/kraken.services : See if tiller rc if present] *********************
fatal: [localhost]: FAILED! => {"changed": true, "cmd": "/opt/cnct/kubernetes/v1.5/bin/kubectl --kubeconfig=/Users/guineveresaenger/.kraken/guinrecon/admin.kubeconfig get deployment tiller-deploy --namespace=kube-system", "delta": "0:00:00.857228", "end": "2017-04-27 17:04:17.077705", "failed": true, "rc": 1, "start": "2017-04-27 17:04:16.220477", "stderr": "Error from server (NotFound): deployments.extensions \"tiller-deploy\" not found", "stdout": "", "stdout_lines": [], "warnings": []}
...ignoring
```

```TASK [roles/kraken.services : Clean up services] *******************************
fatal: [localhost]: FAILED! => {"failed": true, "msg": "The conditional check 'item.status.loadBalancer.ingress[0].hostname is defined and kraken_action == 'down'' failed. The error was: error while evaluating conditional (item.status.loadBalancer.ingress[0].hostname is defined and kraken_action == 'down'): 'ansible.vars.unsafe_proxy.AnsibleUnsafeText object' has no attribute 'status'\n\nThe error appears to have been in '/kraken/ansible/roles/kraken.services/tasks/kill-services.yaml': line 72, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Clean up services\n  ^ here\n"}
...ignoring```

**Comment by DStorck** [comment]
okay - the two errors referenced above are now fixed, but I found another one: 
``` 
TASK [/root/k2/ansible/roles/kraken.fabric/kraken.fabric.flannel : check kube-networking namespace state
``` 
that needs to get fixed as well, same idea

DStorck:
work is done in PR https://github.com/samsung-cnct/k2/pull/470, it also handled another ignored failure to check namespace creation of kube-networking

DStorck:
To Review - now that this has been merged, there should be no more `ignoring...` errors on k2 up. This specifically tried to handle : 
- check kube-networking namespace state
- Ensure the kube-networking namespace exists 
- See if tiller rc is present
- Clean up tiller if present

PR #440, opened 2017-05-19
Fix walk of status object when killing services

NathanZook:
https://github.com/samsung-cnct/k2/issues/268

### DIFF
--- a/ansible/roles/kraken.services/tasks/kill-services.yaml
+++ b/ansible/roles/kraken.services/tasks/kill-services.yaml
@@ -61,7 +61,13 @@
 
 - name: Set services info
   set_fact:
-    the_services: "{{ added_services_map.items()[0][1] }}"
+    the_services: "{{ added_services_map['items'] }}"
+  when: kraken_action == 'down'
+  ignore_errors: yes
+
+- name: Set load balanced service data
+  set_fact:
+    load_balanced_services: "{{ the_services|json_query('[?status.loadBalancer.ingress[0].hostname != null].{namespace: metadata.namespace, name: metadata.name}') }}"
   when: kraken_action == 'down'
   ignore_errors: yes
 
@@ -82,9 +88,8 @@
 
 - name: Clean up services
   command: >
-    {{ kubectl }} --kubeconfig={{ kubeconfig }} delete --namespace {{ item.metadata.namespace }} svc {{ item.metadata.name }}
-  with_items: "{{ the_services }}"
-  when: kraken_action == 'down' and item.status.loadBalancer.ingress[0].hostname is defined
+    {{ kubectl }} --kubeconfig={{ kubeconfig }} delete --namespace {{ item.namespace }} svc {{ item.name }}
+  with_items: "{{ load_balanced_services }}"
   ignore_errors: yes
 
 - name: Delete all service namespaces


davidewatson:
ok to test

ansible/roles/kraken.services/tasks/kill-services.yaml
```diff
@@ -61,7 +61,7 @@
 
 - name: Set services info
   set_fact:
-    the_services: "{{ added_services_map.items()[0][1] }}"
+    the_services: "{{ added_services_map.items['items'] }}"
```
joejulian: Wow... that was screwy in the first place. You can simplify that even further by just avoiding the items function (built in dict object function) and use `added_services_map['items']` - though if there is no items key that will fail - currently we ignore failure so that's probably ok.

ansible/roles/kraken.services/tasks/kill-services.yaml
```diff
@@ -84,7 +84,7 @@
   command: >
     {{ kubectl }} --kubeconfig={{ kubeconfig }} delete --namespace {{ item.metadata.namespace }} svc {{ item.metadata.name }}
   with_items: "{{ the_services }}"
-  when: kraken_action == 'down' and item.status.loadBalancer.ingress[0].hostname is defined
+  when: kraken_action == 'down' and item.status.loadBalancer and item.status.loadBalancer.ingress[0].hostname is defined
```
joejulian: This doesn't work as you would expect. You actually have to say `item.status.loadBalancer is defined` here.

ansible/roles/kraken.services/tasks/kill-services.yaml
NathanZook: Uggh.  Yeah, that what what I did in the container.  Will fix.

ansible/roles/kraken.services/tasks/kill-services.yaml
NathanZook: We're currently failing because item.status.loadBalancer is empty, which is falsey.  I'll see if I can figure out if item.status.loadBalancer might not even be defined some of the time.

joejulian: Ah, ok. We have a custom filter I added, "is_empty" that will result in a bool.

NathanZook: I have checked the source.  loadBalancer is being made explicitly an optional part of status, so something more is needed.  Looking into options.

DStorck:
This should also fix part of https://github.com/samsung-cnct/k2/issues/372

ansible/roles/kraken.services/tasks/kill-services.yaml
```diff
@@ -82,9 +88,8 @@
 
 - name: Clean up services
   command: >
-    {{ kubectl }} --kubeconfig={{ kubeconfig }} delete --namespace {{ item.metadata.namespace }} svc {{ item.metadata.name }}
-  with_items: "{{ the_services }}"
-  when: kraken_action == 'down' and item.status.loadBalancer.ingress[0].hostname is defined
+    {{ kubectl }} --kubeconfig={{ kubeconfig }} delete --namespace {{ item.namespace }} svc {{ item.name }}
+  with_items: "{{ load_balanced_services }}"
```
DStorck: this may still want to just run on 'down', otherwise you will delete all services that every time you run up

PR #442, opened 2017-05-22
Change the tiller-install check to test more reliably

joejulian:
For some reason the results show
```
    "status": {
        "conditions": [
            {
                "lastTransitionTime": "2017-05-22T16:57:39Z",
                "lastUpdateTime": "2017-05-22T16:57:39Z",
                "message": "Deployment has minimum availability.",
                "reason": "MinimumReplicasAvailable",
                "status": "True",
                "type": "Available"
            }
```
despite not having an entry for `availableReplicas`.

See if the status is actually correct.


joejulian:
Nope, that doesn't work.

Issue #85, opened 2016-11-10
Add "kubelet cluster update"


maratoid:
Kubelet cluster update should be a new ansible role that is not part of the default set and must be called explicitly and has a selector based on provider.  The ansible role should perform the upgrade/downgrade (should be the same process) tasks as laid out in the following tickets:
- gke: https://github.com/samsung-cnct/k2/issues/217#issuecomment-285467380 
- aws: https://github.com/samsung-cnct/k2/issues/218 (see last comment for additional detail.  if there is confusion, please speak to @coffeepac)
-- this may be completed best by completing #60 first

coffeepac:
this is blocked by #218 #219

coffeepac:
use the information from here:  https://github.com/samsung-cnct/k2/issues/217#issuecomment-285467380

guineveresaenger:
@maratoid @coffeepac @DStorck 

I've been trying to get my head wrapped into this and have a few questions.
- "kubelet cluster update" refers to what, exactly? I'm confused about the use of the work kubelet here.
- In K2, we seem to be handling AWS and GKE entirely differently: GKE update lives in kraken.provider.gke role as a regular Task, whereas we have an entire role called kraken.upgrade that holds an .aws subfolder (but nothing pertaining to gke). This makes sense given the different ways both providers are structured; I just wanted to see if there were any more thoughts and comments on this.
- We have both update.yaml and upgrade.yaml. What is the difference? Also, there seems to be an unresolved merge conflict in upgrade.yaml.

guineveresaenger:
recap from chat: the work for master nodes is done; I am researching the work for non-master and non-etc nodes.

guineveresaenger:
Starting with aws, as it is being handled differently:
- in kraken.upgrade.aws, there's a task for deleting control plane nodes.
- there is a kill node filter plugin that does this. Because it was written with master nodes in mind, it has a builtin wait period.
- it would seem DRY to use that same filter on non control plane nodes; however this would add extra wait time that I'm not sure we need in this case (cluster nodes can all die and be respawned without the cluster breaking)
- it looks like we want a task for updating non control plane nodes that executes in main after the control plain node tasks have run.
Question:
- Do we want to have a "wait to respawn" for cluster nodes? While killing all at once won't break the cluster, could this result in pods breaking or being lost? When I ran this manually a couple months ago, this didn't happen, but they were pretty small, empty pods. Is this a concern?
@coffeepac just tagging you for reminder purposes

[...]

PR #444, opened 2017-05-22
Nodeupgrade

guineveresaenger:
Inviting comments on this for now. Upgrade works for AWS as far as I can tell.

Things to consider:
- does it make sense to have three separate yaml files per nodepool? I like it because although it looks like a lot of duplicated code, it keeps the concerns separate. However, I feel myself wanting to refactor what is being put into the kill_node filter. I'm not sure how to do that.
- I still would like to find a better way of upgrading than requiring an `up.sh` before runnong `upgrade.sh`. Wouldn't it be nice if we only ran necessary tasks? It would also save time. I learned today that this is because we need to update the cloudinit to use the new version. I tried including the assembler task in kraken.upgrade, but while it did update the cloud-init file, the nodes themselves didn't update their versions.
- This upgrade takes quite a long time, about ten minutes per node. That's over an hour, worse than tearing down a cluster and putting up a new one.
- Can I please delete the very confusing update.yaml? It looks like somebody's lost orphan child, so unless it has a purpose I'm not aware of...?


guineveresaenger:
Oh yeah. I haven't finished the GKE work on this yet.

guineveresaenger:
retest this please

NathanZook:
Yeah, I'm nervous about all of the code duplication.  Especially since we are in an exploratory mode in so many ways, this looks like trouble.

guineveresaenger:
(This adresses issues in #85)

ansible/roles/kraken.upgrade/kraken.upgrade.aws/tasks/main.yaml
```diff
@@ -5,4 +5,15 @@
     - "{{ kraken_config.clusters }}"
   loop_control:
     loop_var: a_cluster
-  
\ No newline at end of file
+
+- include: aws-cluster-nodes.yaml
```
coffeepac: this is not going to work.  the names of the nodePools are arbitrary and we will want this to work going forward.  

we made a short term decision to hardcode the master nodePool name as that's a reasonable default most people won't change, but odds are high that people will add additional nodePools as we go forward.

for now I think it would suffice to either upgrade the master nodePool OR all other nodePools.  we can work out how to handle specify individual nodePools later

guineveresaenger: I see. Then my question is (and this will make the files look so much less cluttered): do we really need to separate out control plane and "other" nodes? The main reason for treating control plane nodes differently was to give each node time to respawn, so the remaining nodes can keep the cluster running in the meantime. The kill_node filter takes care of that. Are there other arguments for giving users the capability to upgrade only parts of the cluster, specifically "control plane" vs "everything else"?
We don't have to worry about hard coding anything in that case.

guineveresaenger:
Ouch. It would be great to be able to terminate the ci check if you do something stupid like forget to rebase or accidentally left something commented out...

guineveresaenger:
The test failed on terraform destroy.

guineveresaenger:
retest this please

guineveresaenger:
@coffeepac  do you mind re-testing this for me?

guineveresaenger:
PR: #473

coffeepac:
I am testing this locally now.  Should be done by lunch.

guineveresaenger:
@coffeepac my tests show they passed?

coffeepac:
@guineveresaenger excellent! this is large enough that it needs someone else to operate it to make sure it works on not your machine.  if we had a better testing framework in place for this type of activity (next milestone maybe?) we would write tests to exercise this new functionality automatically.  lacking that, a second human is advised.

guineveresaenger:
@coffeepac that makes so much sense! and makes me feel so much better about changes that, um, don't get used on the automatic tests... hopefully the documentation rewrite is clear.

guineveresaenger:
@coffeepac I think I fixed the documentation to talk about the `--nodepools` flag, and I added a line to the help text that mentions `./upgrade.sh`. Please let me know if that takes care of the explanations part!

guineveresaenger:
@coffeepac this did fail e2e tests as far as I can tell. I'll trigger it one more time. One more dayyyyyyy

guineveresaenger:
As per comments in #85, this is not yet ready to merge. Apologies.

guineveresaenger:
unless we want to merge this, knowing it will work for AWS but not gke.

guineveresaenger:
Should be ready to go now.

Issue #378, opened 2017-05-01
Sanity check needed in deletion of folder on cleanup


venezia:
in `ansible/roles/kraken.clean/tasks/main.yaml` there is a possibility that `a_cluster.name` is not set / empty.  This can result in ansible trying to delete the `.kraken` folder.

While this cannot happen during a docker image run (because the folder is busy as it is mounted) - this could balloon in a different setup.

Please add a sanity check such that _If `a_cluster.name` is empty/not-set, skip step and warn the user_

PR #463, opened 2017-05-30
Don't delete cluster dirs that cover config base.

NathanZook:


ansible/roles/kraken.clean/tasks/main.yaml
```diff
@@ -1,9 +1,20 @@
 ---
+- name: Register safe cluster name directory queries
+  set_fact:
+    cluster_dirs_query: "[*].join('/', ['{{ config_base }}', name])"
+    cluster_name_test: "[?!starts_with('{{config_base | realpath }}', @)]"
```
coffeepac: a comment to say what this is doing, exactly, would be great.

I can't find any documentation on what the leading '?!' does.  not surprising as its normally difficult to search for operators.

coffeepac:
general note:  the pipelet job will never be successful as k2 no longer exists on pipelet.  don't let that be a gate for this PR

NathanZook:
This is to fix bug #378.

PR #466, opened 2017-05-31
removed redundant helm and made into one-liner for ease of cut and paste

DStorck:
for terminal commands with k2, I need to edit this command and replace `root` with `$HOME`, which was difficult when the command was three lines long. I also removed an extra `helm` hiding in there

### DIFF
--- a/ansible/roles/kraken.cluster_common/templates/help.txt.jinja2
+++ b/ansible/roles/kraken.cluster_common/templates/help.txt.jinja2
@@ -9,9 +9,7 @@ To use helm:
 ------------
 
 {% for cluster in kraken_config.clusters %}
-    KUBECONFIG={{ config_base }}/{{ cluster.name }}/admin.kubeconfig helm \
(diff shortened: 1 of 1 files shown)


guineveresaenger:
I'm unclear about how this works - will it merge automatically if I write an "approved" review? And I need to wait for the tests to complete anyway right?

davidewatson:
Merging and approval are different. You might, for example, approve a change and not merge it in order to wait for tests to pass, allow someone else to take a look, etc. If you are going to approve but not merge, you should probably say why, so anyone else that comes across the PR understands what to do next.

guineveresaenger:
Thanks, David - makes sense!

guineveresaenger:
to clarify: looks good to me; waiting for tests to finish.

DStorck:
it seems like all our tests are failing at the moment.  This PR doesn't actually DO anything, it just spits out a command that users can run from the command line, so if it fails, it seems like we need to dig in a bit and figure out what is making things fail.

PR #470, opened 2017-06-01
errors are now ignored on up

DStorck:
Method now checks for an error in stdout instead of checking for a failure, which means no more ignored errors will show up as failures when they were expected to fail. 
This is for issue https://github.com/samsung-cnct/k2/issues/372

### DIFF
--- a/ansible/roles/kraken.fabric/kraken.fabric.flannel/tasks/main.yaml
+++ b/ansible/roles/kraken.fabric/kraken.fabric.flannel/tasks/main.yaml
@@ -41,6 +41,7 @@
   changed_when: false
   register: kube_networking_namespace
   when: "{{ not (dryrun | bool) }}"
+  failed_when: false
 
 - name: Ensure the kube-networking namespace exists
   command: >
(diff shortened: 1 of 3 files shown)


davidewatson:
The continous-intergration/jenkins/pr-head [job](https://common-jenkins.kubeme.io/job/K2/job/PR-470/1/consoleFull) failed:

```
[gke] fatal: [localhost]: FAILED! => {"failed": true, "msg": "The conditional check ''Error' not in tiller_present.stderr and not tiller_present | skipped ' failed. The error was: error while evaluating conditional ('Error' not in tiller_present.stderr and not tiller_present | skipped ): Unable to look up a name or access an attribute in template string ({% if 'Error' not in tiller_present.stderr and not tiller_present | skipped  %} True {% else %} False {% endif %}).\nMake sure your variable name does not contain invalid characters like '-': argument of type 'StrictUndefined' is not iterable\n\nThe error appears to have been in '/home/jenkins/workspace/K2_PR-470-TPDGKSIFVLDIYDAKTAAXP7NPUFNHG43FZPWSZ6NCGYPYR4ZIXWKA/ansible/roles/kraken.services/tasks/kill-services.yaml': line 60, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Clean up tiller if present\n  ^ here\n"}
```

DStorck:
well , that makes sense, I had been testing an aws cluster. Don't know what is different about gke in terms of tiller.

DStorck:
please retest this

PR #471, opened 2017-06-01
Initial schema validator. Also make ids and $refs absolute URIs. Remove malformed or redundant

davidewatson:
arrays, non-existent types, and other general clean up.


davidewatson:
retest this please

schemas/config/v1/nodeConfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "http://judkins.house/apis/k2/v1/nodeConfig.json",
+
+  "description": "Node Configurations",
+  "items": {
+    "anyOf": [
```
DStorck: should this be oneOf?

schemas/config/v1/providerConfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "http://judkins.house/apis/k2/v1/providerConfig.json",
+
+  "description": "Provider Configurations",
+  "anyOf": [
+    { "$ref": "http://judkins.house/apis/k2/v1/awsProviderConfig.json" },
+    { "//$ref": "http://judkins.house/apis/k2/v1/gkeProviderConfig.json" }
```
DStorck: is the `//` before `$ref` a typo?

schemas/config/v1/nodeConfig.json
davidewatson: oneOf [means](https://spacetelescope.github.io/understanding-json-schema/reference/combining.html#oneof) _exactly_ one, which is too restrictive. Consider two different providers which both share the same required properties. Since additionalProperties == true, any document which is valid against one is also valid against the other.

schemas/config/v1/providerConfig.json
davidewatson: Good catch! Fixed.

schemas/config/v1/nodeConfig.json
DStorck: since it was checking whether the node was aws or gke, i thought it would just want one or the other

davidewatson:
CI failed: ```java.io.IOException: Pipe not connected
	at java.io.PipedOutputStream.write(PipedOutputStream.java:140)
	at java.io.OutputStream.write(OutputStream.java:75)```

Trying again: https://common-jenkins.kubeme.io/job/K2/job/PR-471/5/

schemas/config/v1/nodeConfig.json
davidewatson: You're right, we should go back and check which of anyOf can be oneOf. I'd like to open tickets for any additional changes we want to make. As it is, the team keeps on implementing new features faster than I can get this baseline through CI. And since the existing schema has syntax errors, this arguably can't be any worse.

coffeepac:
While there is no single CI run that completed successfully, two in a row completed the GKE stream and the AWS stream.  These streams are independent and both have flaky spots.  

When the current jobs are complete I'm going to merge this.

PR #423, opened 2017-05-13
Update readme versions

davidewatson:
This a proposed solution to the problem observed in k2-tools/#43. Really the links should point to a tagged release to ensure they never break.

PR #472, opened 2017-06-02
Point to k2-tools for the Ansible and Terraform versions.

davidewatson:


davidewatson:
This replaces #423

DStorck:
LGTM

PR #473, opened 2017-06-02
Helmcleanup

guineveresaenger:
After chatting with @davidewatson, we are currently not making changes to the schema, because they would be breaking changes. 
This means we don't currently verify against garbage input on helmOverride in the config file (it should be empty, true, or false) but other than that, this should work better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,7 +424,25 @@ Then run:
 docker run $K2OPTS quay.io/samsung_cnct/k2:latest ./upgrade.sh --config $HOME/.kraken/${CLUSTER}.yaml
 ```
 
-Refer to [Kuberntes release versioning](https://github.com/kubernetes/kubernetes/blob/release-1.5.4/docs/design/versioning.md#supported-releases) for questions about supported versions
+Refer to [Kubernetes release versioning](https://github.com/kubernetes/kubernetes/blob/release-1.5.4/docs/design/versioning.md#supported-releases) for questions about supported versions
+
+### Handling unsupported versions of helm
+Currently, and for the foreseeable future, new helm releases will be shipped after new Kubernetes releases, resulting in helm possibly not being supported for the latest Kubernetes version.
+You have two options.
+
+#### Option 1: Overriding helm in K2 config file
+In the K2 config file, set the cluster level key `helmOverride` to `true` if you wish to use the latest version of helm that is available. Warning: you might encounter bugs.
+Set `helmOverride` to `false` if you would like to run K2 without helm.
+
+#### Option 2: Overriding helm via environment variable
+This will automatically happen if you are trying to run a cluster with a Kubernetes version that does not have helm support, and you did not set helmOverride in the K2 config file.
+K2 will halt and, via fail message, prompt you to set a cluster specific helm override env variable to true or false.
+
+```bash
+export helm_override_<CLUSTER>=<TRUE/FALSE>
+```
+Now, run cluster up again, and K2 will use the override condition you specified.
+
 
 ## Destroying a Kubernetes Cluster
 
--- a/README.md
+++ b/README.md
@@ -431,7 +431,7 @@ Currently, and for the foreseeable future, new helm releases will be shipped aft
 You have two options.
 
 #### Option 1: Overriding helm in K2 config file
-In the K2 config file, set the cluster level key `helmOverride` to `true` if you wish to use the latest version of helm that is available. Warning: you might encounter bugs.
+In the K2 config file, set the cluster level key `helmOverride` to `true` if you wish to use the latest version of helm that is available. Warning: since this would be using a version of helm that does not support your k8s version, this may result in bugs.
 Set `helmOverride` to `false` if you would like to run K2 without helm.
 
 #### Option 2: Overriding helm via environment variable
--- a/README.md
+++ b/README.md
@@ -424,14 +424,12 @@ Then run:
 docker run $K2OPTS quay.io/samsung_cnct/k2:latest ./upgrade.sh --config $HOME/.kraken/${CLUSTER}.yaml
 ```
 
-Refer to [Kubernetes release versioning](https://github.com/kubernetes/kubernetes/blob/release-1.5.4/docs/design/versioning.md#supported-releases) for questions about supported versions
-
 ### Handling unsupported versions of helm
 Currently, and for the foreseeable future, new helm releases will be shipped after new Kubernetes releases, resulting in helm possibly not being supported for the latest Kubernetes version.
 You have two options.
 
 #### Option 1: Overriding helm in K2 config file
-In the K2 config file, set the cluster level key `helmOverride` to `true` if you wish to use the latest version of helm that is available. Warning: since this would be using a version of helm that does not support your k8s version, this may result in bugs.
+In the K2 config file, set the cluster level key `helmOverride` to `true` if you wish to use the latest version of helm that is available. Warning: since this would be using a version of helm that does not support your k8s version, this may result in unexpected behavior.
 Set `helmOverride` to `false` if you would like to run K2 without helm.
 
 #### Option 2: Overriding helm via environment variable
--- a/ansible/roles/kraken.config/tasks/helm-versioning.yaml
+++ b/ansible/roles/kraken.config/tasks/helm-versioning.yaml
@@ -5,14 +5,19 @@
   set_fact:
     kubernetes_minor_version: "{{ kubernetes_minor_versions[cluster.name] }}"
 
-- name: Set helm override from  environment variable
+- name: Look up environment variable for helm override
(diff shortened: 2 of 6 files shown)


ansible/roles/kraken.config/tasks/helm-versioning.yaml
```diff
@@ -28,21 +29,19 @@
...
   set_fact:
-    helm_overrides: "{{ helm_overrides | default({}) | combine
-      (
-        {cluster.name: helm_override }
-      )
-    }}"
+    helm_overrides: "{{ helm_overrides | default({}) | combine(joe) }}"
   when: helm_override != ""
+  vars:
+    joe: "{{ { cluster.name : helm_override } }}"
```
DStorck: joe?

ansible/roles/kraken.config/tasks/helm-versioning.yaml
```diff
@@ -28,21 +29,19 @@
...
 - name: Keep track of helm availability by cluster name
   set_fact:
-    helm_availabilities: "{{ helm_availabilities | default({}) | combine
-      (
-        {cluster.name: helm_available }
-      )
-    }}"
+    helm_availabilities: "{{ helm_availabilities | default({}) | combine(joe) }}"
+  vars:
+    joe: "{{ { cluster.name: helm_available } }}"
```
DStorck: more joe

guineveresaenger: thank you! I need to stop that.

ansible/roles/kraken.services/tasks/kill-services.yaml
```diff
@@ -11,10 +11,9 @@
     helm_command: "/opt/cnct/kubernetes/{{ kubernetes_minor_version }}/bin/helm"
   when: helm_availabilities[cluster.name] == true
 
-- name: In case of helm not being available, if override was specified, override to latest helm
+- name: In case of helm not being available, override to latest helm
   set_fact:
     helm_command: "/opt/cnct/kubernetes/{{ latest_helm }}/bin/helm"
-  when: (helm_availabilities[cluster.name] == false) and (helm_overrides[cluster.name] == true)
```
DStorck: should there be a when clause here to check if helm is not available?

guineveresaenger: I don't think so, unless I'm misunderstanding something. As per @mikeln 's request, helm should always run on down due to if there's any services to kill, we should use any helm version we can to clean them up. I know kill-services runs on both up and down, but it kills services, so. That's why I took out the when statement.

DStorck:
did you also add the helmOverride field to the gke-config.yaml? or should we?

guineveresaenger:
@DStorck when running ./up.sh with gke, it will update automatically on cluster up, when the config file reflects these changes.

ansible/roles/kraken.services/tasks/kill-services.yaml
DStorck: okay, it just seemed strange that the task is labeled, 'in case of ', and then it is going to set a helm_command no matter what , maybe a change to the name of the task would make it clearer what the intention is

DStorck: this will always change the helm version to latest and that might not be necessary if there is already a version set

coffeepac:
My expected precedence is ENV vars outrank config files.  They're overrides.  Did you have a reason to have the config file be the trump location?

guineveresaenger:
@coffeepac David made me aware today that there is a convention for order of precedence that I was previously unfamiliar with.
I originally went with this order because in practice, the env override seems to come from a place of "whoops! we don't have a helm version for you; here are some options, what would you like to do" whereas setting it in the config comes from a place of "I know I don't have a helm version for this, and I'm making advance plans to deal with that". A user could, of course, set the env before running `up` the first time, but that seems less intuitive still; you'd have to have some really explicit documentation.

coffeepac:
@guineveresaenger I would agree will we need clear documentation for this.  The process isn't clean and obvious.

What is the convention that David mentioned?  And what is your plan for this?

guineveresaenger:
@coffeepac David said what you said: ENV vars according to convention outrank config files. I did not know this. :(  He also mentioned a third type of var but I am not sure which it was, or where it ranks.

In general, I would like to go by convention, because at my experience level I probably don't have good enough reasons to break with convention... so I'll change it back.

We use helm no differently for gke than aws clusters, correct? Just checking as per @DStorck's point above.

ansible/roles/kraken.services/tasks/kill-services.yaml
guineveresaenger: Oh, I think I see why you are confused. This will always set to latest helm, despite availability of helm. Thank you for catching that.

ansible/roles/kraken.services/tasks/kill-services.yaml
```diff
@@ -11,10 +11,10 @@
     helm_command: "/opt/cnct/kubernetes/{{ kubernetes_minor_version }}/bin/helm"
   when: helm_availabilities[cluster.name] == true
 
-- name: In case of helm not being available, if override was specified, override to latest helm
+- name: In case of helm not being available, always override to latest helm
   set_fact:
     helm_command: "/opt/cnct/kubernetes/{{ latest_helm }}/bin/helm"
-  when: (helm_availabilities[cluster.name] == false) and (helm_overrides[cluster.name] == true)
+    when: (helm_availabilities[cluster.name] == false)
```
DStorck: shouldn't be indented

README.md
```diff
@@ -424,7 +424,25 @@ Then run:
 docker run $K2OPTS quay.io/samsung_cnct/k2:latest ./upgrade.sh --config $HOME/.kraken/${CLUSTER}.yaml
 ```
 
-Refer to [Kuberntes release versioning](https://github.com/kubernetes/kubernetes/blob/release-1.5.4/docs/design/versioning.md#supported-releases) for questions about supported versions
+Refer to [Kubernetes release versioning](https://github.com/kubernetes/kubernetes/blob/release-1.5.4/docs/design/versioning.md#supported-releases) for questions about supported versions
```
coffeepac: this link doesn't go anywhere, I am not sure where it should go or what it is for.  maybe remove it for now?

README.md
```diff
@@ -424,7 +424,25 @@ Then run:
...
+
+### Handling unsupported versions of helm
+Currently, and for the foreseeable future, new helm releases will be shipped after new Kubernetes releases, resulting in helm possibly not being supported for the latest Kubernetes version.
+You have two options.
+
+#### Option 1: Overriding helm in K2 config file
+In the K2 config file, set the cluster level key `helmOverride` to `true` if you wish to use the latest version of helm that is available. Warning: since this would be using a version of helm that does not support your k8s version, this may result in bugs.
```
coffeepac: '... may result in ~~bugs~~ unexpected behavior'.  

small thing but we don't want to use the word 'bug.'

ansible/roles/kraken.config/tasks/helm-versioning.yaml
```diff
@@ -5,14 +5,12 @@
   set_fact:
     kubernetes_minor_version: "{{ kubernetes_minor_versions[cluster.name] }}"
 
-- name: Set helm override env variable key to contain the cluster name
+- name: Set overall helm_override to to env var if it has been set, config var if not
```
coffeepac: nit:  '... helm_override to ~~to~~ env...'

ansible/roles/kraken.config/tasks/helm-versioning.yaml
```diff
@@ -21,28 +19,25 @@
...
+- name: Stop and print out message if helm is not available and helm_override has not been set in either env or config
   fail:
     msg: "This version of Kubernetes does not currently have helm support. If you would like to continue without using helm, execute following command:
-    export {{ helm_override_key }}=false && ./up.sh
+    export {{ helm_override_key }}=false and then run cluster up again.
     If you would like to use the latest version of helm instead, execute:
-    export {{ helm_override_key }}=true && ./up.sh
-    WARNING: You may have to specify your config file path if it is not located in the default path."
-  when: helm_available == false and helm_override == ""
+    export {{ helm_override_key }}=trueand then run your cluster up again."
```
coffeepac: nit: probably need a space between '=true' and 'and then'

README.md
guineveresaenger: Awwww! but they're so cute!

guineveresaenger:
Done!

coffeepac:
discussion came up to do this with CLI flags.  that will be handled in #489

discussion came up for how to handle dashes and underscores.  we settled on the process laid out in #490  

this PR can go now.

Issue #184, opened 2017-02-08
K2 rbac configuration


coffeepac:
We are currently using the following in our API-server-manifest:

`    - --authorization-rbac-super-user={{ kraken_config.kubeAuth.authz.rbac.super_user }}`

This is being [removed in v1.6](https://kubernetes.io/docs/admin/authorization/).  We need to switch to using the user bootstrapping method

leahnp:
I am not able to bring up a cluster with security enabled, I am running into this error:

 `"DexApp": {"RedirectUri": "http://common-tools.kubeme.io:30080/callback"}}, "version": "0.1.0"}, "msg": "RepresenterError: cannot represent an object: Dex"}`

I believe I might be hitting this bug: https://github.com/openshift/openshift-ansible/issues/3111

But when I try to roll back my ansible version to test it, I hit another big with ELB 'none types'. 

Can we punt this until ansible releases their next stable version or until we are upgrading to K8s 1.6?

coffeepac:
@leahnp its concerning that we currently can't bring up a new kubernetes cluster with our security enabled.  however, this isn't part of our e2e tests or our generated configuration.  I'm okay with punting on this for a bit.  feel free to un-assign and move back to ToDo

coffeepac:
this should be resolved in ansible with the design spec as laid out here: https://github.com/samsung-cnct/k2/issues/222#issuecomment-284583580

there should be folders created for 1.4 and 1.5 to contain the current template.

coffeepac:
DexApp install error fixed with: https://github.com/samsung-cnct/k2-tools/pull/32

DStorck:
When this gets completed, please update kubeauth.md  in k2/documentation/kraken-configs

mikeln:
v1alpha1 --> v1beta1  in 1.6

mikeln:
working through stuff

mikeln:
k2/ansible/roles/kraken.readiness/filter_plugins/expected_masters.py  has issues...
mainly it references `total_masters` when it should be `total_nodes`.

mikeln:
Separating RBAC and Cert Gen out of kraken.readiness role  into its own role.

mikeln:
The 1.6 cluster will come up ok with the default rbac (which is how it should be).
the main issue is getting the correct initial rbac permssions for kubectl so it can install our rbac specifications, and allow it to control the system.

[...]

PR #475, opened 2017-06-06
rbac 16 hotfix

mikeln:
Add missing changes


mikeln:
ok, that github outage messed up the return status from jenkins.   The last run was good...but how to get those results back here?

coffeepac:
I was hoping we would just have to wait for a bit and the results would show up.

mikeln:
additional fix for #184

Issue #262, opened 2017-03-15
Fail depreciated k2 schema


davidewatson:
New:

Schemas are depreciated in whole or not at all. This is a result of the earlier design decision to only version the entire config file and not each individual stanza. This ticket is to add a check to fail the kraken.config role if the schema version claimed by the config file is marked as depreciated. See the comments for more detail.

Old:

# Task

Per #221 , before a property can be removed it must be depreciated first to give the user a chance to stop depending on it. Programmatically this is done by adding the property to the depreciated list within the k2 schema. For example [1] gives a hypothetical schema stanza. Since depreciated is an extension to JSON schema (see [PR #173](https://github.com/json-schema-org/json-schema-spec/pull/173) for the current discussion) no standard validator will print warnings. This ticket is to create an Ansible task which will print warnings as part of the kraken.config role.  

[1]

```yaml
"kubeConfig": {
"title": "A Kubernetes configuration",
"description": "The location and version of a container containing the Kubernetes hyperkube binary.",
"properties": {
"name": {
"default": "defaultKubeConfig",
"description": "Name of the Kubernetes configuration.",
"type": "string"
},
"hyperkubeLocation": {
"default": "gcr.io/google_containers/hyperkube",
"description": "Location of the Kubernetes container.",
"format": "uri",
"type": "string"
},
"version": {
"default": "v1.5.2",
"description": "Version of the hyperkube binary.",
"format": "symver",
"type": "string"
}
},
"required": [
"name"
],
"depreciated": [
"hyperkubeLocation"
],
"type": "object"
},
```

coffeepac:
blocked by #258

coffeepac:
@davidewatson  to fill in new details for this ticket

davidewatson:
The logic to load and validate configuration files is in ansible/roles/kraken.config/tasks/main.yaml. There are four cases:

1) The configuration is unsupported because it was created before k2 had a schema (verified by checking for a version propery in the configuration file).
2) The configuration is unsupported because there are no schema files for the specified version (verified by checking if there is a schema file at schemas/config/{{ kraken_config_user.version }}/config.json).
3) The configuration is supported, but may not be soon (verified by checking if the corresponding schema has a depreciated field set. E.g.

```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "id": "http://judkins.house/apis/k2/v1/config.json",

  "depreciated": true,

  "title": "k2 configuration file",
[...]

PR #476, opened 2017-06-06
check if --force is passed for deprecated versions

DStorck:
this is a check to see if the user has passed the --force option on `up` which will allow them to use their deprecated config, if they don't pass that flag, it fails with a warning. This is for issue https://github.com/samsung-cnct/k2/issues/262


ansible/roles/kraken.config/tasks/main.yaml
```diff
@@ -48,11 +48,25 @@
...
+- name: Set deprecation fact
+  set_fact:
+    deprecated: "{{ validation.result.schema.deprecated | default(false) }}"
+
+- name: Check for deprecated version
+  fail:
+    msg: >-
+      "Configuration file {{ config_file }} uses an API version {{ kraken_config_user.version }}
+      which has been depreciated. It is strongly recommended that you update your configuration file before proceeding.
+      If you would like to proceed anyway, run up again with --force,  but note that support for this API version will be dropped in the near future."
```
davidewatson: LGTM

I know some of our ansible does have unsightly 151 character length lines, but since we are using [YAML block strings](https://stackoverflow.com/a/21699210), maybe this text could smaller, and more uniform? ;)

DStorck: this was the warning you gave me in our chat, i'd be happy to shrink it down , but it displays alright when it fails out. 
```TASK [kraken.config : Check for deprecated version] ******************************************************************************************************************************************************************************************
task path: /root/k2/ansible/roles/kraken.config/tasks/main.yaml:60
fatal: [localhost]: FAILED! => {
    "changed": false,
    "failed": true,
    "msg": "\"Configuration file /root/.kraken/config.yaml uses an API version v1 which has been depreciated. It is strongly recommended that you update your configuration file before proceeding. If you would like to proceed anyway, run up again with --force,  but note that support for this API version will be dropped in the near future.\""
}
```

DStorck: it actually looks normal on my screen, it just goes on and on in github :)

DStorck: So would you like me to edit the comment to make it shorter, or just edit it to make it look better in main.yaml?

DStorck:
retest this please

ansible/roles/kraken.config/tasks/main.yaml
davidewatson: I was only referring to the length of the line in main.yaml.

ansible/roles/kraken.config/tasks/main.yaml
```diff
@@ -48,11 +48,26 @@
...
+  set_fact:
+    deprecated: "{{ validation.result.schema.deprecated | default(false) }}"
+
+- name: Check for deprecated version
+  fail:
+    msg: >-
+      "Configuration file {{ config_file }} uses an API version {{ kraken_config_user.version }}
+      which has been depreciated. It is strongly recommended that you update your configuration
+      file before proceeding.  If you would like to proceed anyway, run up again with --force,
+      but note that support for this API version will be dropped in the near future."
```
joejulian: Minor nit, but not one that should block merging. When you're using `>` you don't need to add quotes (`"`). The following indented lines are already considered to be a string value.

DStorck:
this closes https://github.com/samsung-cnct/k2/issues/262

PR #482, opened 2017-06-07
Use Joe's test build of k2-tools

joejulian:


joejulian:
This is essentially what we should do for k2-tools CI tests

joejulian:
https://common-jenkins.kubeme.io/job/K2/view/change-requests/job/PR-482/17/

Looks like it passes to me. Feel free to double check that I'm testing it appropriately. If so, it's ready to merge.

PR #487, opened 2017-06-08
added length checks

DStorck:
adds a max to clusterName, and a min to ensure that there is something there, as well as a length check for ELBs

### DIFF
--- a/schemas/config/v1/loadBalancerConfig.json
+++ b/schemas/config/v1/loadBalancerConfig.json
@@ -7,7 +7,8 @@
   "properties": {
     "name": {
       "description": "Name of the load balancer configuration",
-      "type": "string"
+      "type": "string",
+      "maxLength": 32
     },
     "type": {
       "description": "Type of load balancer",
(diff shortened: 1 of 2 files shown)


schemas/config/v1/loadBalancerConfig.json
```diff
@@ -7,7 +7,8 @@
   "properties": {
     "name": {
       "description": "Name of the load balancer configuration",
-      "type": "string"
+      "type": "string",
+      "maxLength": 32
```
davidewatson: Is there any reason loadBalancerConfig does not get fancy minLength?

DStorck: I didn't put one in because name is not technically required on the jsonschema. I put one on the cluster name as an added check to make sure the cluster actually had a name.

davidewatson:
Merging, since CI failed due to pre-existing conditions, and validation passed for AWS and GKE...

```
TASK [kraken.config : Validate k2 configuration under schema] ******************
ok: [localhost]
```

Issue #496, opened 2017-06-09
helmOveride not defined


davidewatson:
helmOverride needs a default in ansible, and if not already done, it should be added as an _optional_ property in the schema.

>TASK [kraken.config : Set overall helm_override to env var if it has been set, config var if not] ***
fatal: [localhost]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'dict object' has no attribute 'helmOverride'\n\nThe error appears to have been in '/workspace/ansible/roles/kraken.config/tasks/helm-versioning.yaml': line 8, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Set overall helm_override to env var if it has been set, config var if not\n  ^ here\n"}

coffeepac:
helmOveride should have no default.  it should be marked optional

blackdog0403:
Yes, it should be marked optional but, up.sh doesn't not work without setting up helm_override_xxx env or set key helmOverride in schema at the moment 

It doesn't looks good to me that users cannot run without setup things.

That would better have helmOverride = false default or it needs some excetional process that user don't set value for helmOverride

davidewatson:
Fixed by #497.

PR #497, opened 2017-06-09
Set default for helmOverride (#496).

davidewatson:


DStorck:
Looks good.

davidewatson:
As written, helm_override and cluster.helmOverride are assumed to always be defined, which is a breaking schema change. After a misunderstanding of your comment and a few random commits, I spoke with @guineveresaenger. I think a default "" is correct insofar as it matches the existing logic.

davidewatson:
The beginnings of a schema for helmOverride:

```
    "helmOverride": {
      "description": "Undefined if k2 should halt when there is no supported version of helm for the version of kubernetes specified, true if the latest, and possibly unsupported, version of helm should be used, and false if helm should be skipped entirely.",
      "type": [ "boolean", "string" ]
    }
```

Issue #492, opened 2017-06-09
Organize RBAC Files Under Applicable Role


mikeln:
The RBAC definition file should be placed under the role it applies to.

The "general system" rbac should live under kraken.rbac,  but the canal rbac should live under the correct fabric role directory instead of kraken.rbac (similar to how the DNS rbac is organized).  This also helps with isolation when weave fabric is added.

alejandroEsc:
@mikeln let me know if you need/want help with this one.

mikeln:
PR merged.

PR #499, opened 2017-06-09
492 move fabric rbac to deployment.  dryrun checks

mikeln:
#492 move fabric rbac under the role (vs in common rbac).
add dryrun checks to any ansible task with a kubectl call
fix `when:` warnings for dryrun checks.


blackdog0403:
For example, take them out from main.yaml then add like this on bottom of main.yaml
- include: ensure-kube-networking.yaml
  when: not ( config_dryrun | bool )
- include: deploy-canal-config-daemon.yaml
  when: not ( config_dryrun | bool )

mikeln:
That type of split out may be needed soon...but it would also need to be pushed up to a higher level, i.e. above canal/flannel because weave should also need that namespace.  But, we should worry about that in any reorganization of the fabric tasking.  @alejandroEsc (?)

PR #501, opened 2017-06-09
Dryrun tags bug fixing

blackdog0403:
Should merge after #492 merged into master.

Changing 'dryrun' -> 'config_dryrun' will cause conflicts and bugs


blackdog0403:
I re-change 'config_dryrun' to 'dryrun'. Let's discuss about it then go!

ansible/roles/kraken.fabric/kraken.fabric.flannel/tasks/main.yaml
```diff
@@ -107,7 +107,7 @@
   command: >
     {{ kubectl }} --kubeconfig={{ kubeconfig | expanduser }} apply -f {{ config_base | expanduser }}/{{ cluster.name }}/auth/rbac-canal-policy.yaml
   register: job_rbac_canal_results
-  when: 
+  when:
     - "cluster.kubeAuth.authz.rbac is defined"
     - "(not (dryrun | bool))"
```
joejulian: In most places you don't place quotes around the when clauses. Let's be consistent and not put quotes here.

joejulian:
Well, it does look good, but dryrun does not work, so somehow the logic is failing.

blackdog0403:
@joejulian there is some error on here which is very weird to me. I'm working on it.

```
- name: Wait for api server
  wait_for:
    host: "{{ item | regex_replace('https://','') }}"
    port: 443
    timeout: 600
  with_items: "{{ api_servers }}"
  vars:
    kubefn: "{{ config_base | expanduser }}/{{ cluster.name }}/admin.kubeconfig"
    kubeconfig: "{{ kubefn | is_file | ternary( kubefn, '/dev/null' ) }}"
    api_servers: "{{ lookup('file', kubeconfig) | from_yaml | json_query('clusters[*].cluster.server') }}"
  when: not ( dryrun | bool )
```
when we run this with dryrun admin.kubeconfig file is not generated for intention.
However, even if dryrun tag is set, they check the existency of admin.kubeconfig file

Issue #491, opened 2017-06-09
Add Missing Config Items Required for 1.6 RBAC Enable


mikeln:
Missed in the initial checkin.

Need `rbac` defined
Need `authn.basic..group` set.

```
 - &rbacKubeAuth
      authz:
        rbac: 1
      authn:
        basic:
          -
            password: "BadPassword"
            user: "admin"
            group: "system:masters"
        default_basic_user: "admin"
```

in 1.5:
```
 - &rbacKubeAuth
      authz:
        rbac: 1
        super_user: <some super user value>
      authn:
        basic:
          -
            password: "BadPassword"
            user: "admin"
            group: "system:masters"
        default_basic_user: "admin"
```

alejandroEsc:
BadPassword, so meta. Anyway ill let you know if that stanza is working for me in a few.

mikeln:
merged.

PR #502, opened 2017-06-10
add rbac stanza for use when needed 491

mikeln:
#491 
Add rbacKubeAuth example stanza that can be used when needed.

### DIFF
--- a/ansible/roles/kraken.config/files/config.yaml
+++ b/ansible/roles/kraken.config/files/config.yaml
@@ -371,6 +371,19 @@ definitions:
             password: "ChangeMe"
             user: "admin"
         default_basic_user: "admin"
+   - &rbacKubeAuth
+      authz: 
+        rbac:
+          # super_user is required until kubernetes 1.5 is no longer supported by k2.
+          # It is not used by kubernetes 1.6 or later.
+          super_user: "placeholder"
+      authn:
+        basic:
+          -
+            password: "ChangeMe"
+            user: "admin"
+            group: "system:masters"
+        default_basic_user: "admin"
 
 # This is the core of the new configuration.
 
--- a/ansible/roles/kraken.config/files/gke-config.yaml
+++ b/ansible/roles/kraken.config/files/gke-config.yaml
@@ -134,6 +134,19 @@ definitions:
             password: "ChangeMe"
             user: "admin"
         default_basic_user: "admin"
+   - &rbacKubeAuth
+      authz:
+        rbac:
+          # super_user is required until kubernetes 1.5 is no longer supported by k2.
+          # It is not used by kubernetes 1.6 or later.
+          super_user: "placeholder"
+      authn:
+        basic:
+          -
+            password: "ChangeMe"
+            user: "admin"
+            group: "system:masters"
+        default_basic_user: "admin"
   providerConfigs:
     - &defaultGKE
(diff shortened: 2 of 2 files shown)


mikeln:
authz is required, however, all items in authz are not.  if you are not running rbac, then there is no requirement for any fields, except that the auth stanza exists.

authn.basic does have required fields specified, but I'm not sure that authn is required to actually always have an entry in the base array.

But this brings up a point, the schema requirements are different for each release, and there may or may not be required fields depending on how you are running your system.  e.g.  if you want to rbac in 1.6..then authz.rbac and authn.basic[]password,name,group and authn.default_basic_user are required.  
If you want rbac in 1.5 then the same plus authz.rbac.super_user is required.

davidewatson:
There is no way to enforce different schema depending on the value of a property (in this case the value of the k8s version). There was a proposal to fix this in the [Draft 5](https://github.com/json-schema/json-schema/wiki/%24data-(v5-proposal)) standard. I think for now we need to assume the least common denominator (i.e. required iff required in the intersection of the supported schema).

alejandroEsc:
@davidewatson to that point, about enforcing schemas depending on k8s version. We could make different templates enforcing defaults that fit the min k8s version needs. In this case, having rbac as the default.

ansible/roles/kraken.config/files/gke-config.yaml
```diff
@@ -134,6 +134,19 @@ definitions:
             password: "ChangeMe"
             user: "admin"
         default_basic_user: "admin"
+   - &rbacKubeAuth
+      authz:
+        rbac:
+          # super_user is required until kubernetes 1.5 is no longer supported by k2.
+          #           # It is not used by kubernetes 1.6 or later.
```
joejulian: paste fail

alejandroEsc:
+1 for that commit message

alejandroEsc:
Should auth config entries by subtypes of a general auth type? I ask because it seems, and i could be wrong we do that for  fabrics, to what extent should we be doing this for configurations. Additionally can we organize the schemas to reflect this case, otherwise i can see managing schemas becoming unwieldy

mikeln:
@alejandroEsc schema discussions have been moved into their own issue.  
Reorganization and handing of schema per version etc is something to work toward, but a large future task.

davidewatson:
I am working on a schema to close this out.

ansible/roles/kraken.config/files/config.yaml
```diff
@@ -371,6 +371,19 @@ definitions:
             password: "ChangeMe"
             user: "admin"
         default_basic_user: "admin"
+   - &rbacKubeAuth
+      authz: 
+        rbac:
+          # super_user is required until kubernetes 1.5 is no longer supported by k2.
+          # It is not used by kubernetes 1.6 or later.
+          super_user: "placeholder"
```
davidewatson: In previous versions of k2, the default for super_user was "admin". Will a default of "placeholder" work for k8s <= 1.5, and if not, will the previous default of "admin" work for k8s 1.6, since it is reportedly not used?

mikeln:
wth?  I didn't delete this branch

joejulian:
This has already passed as https://common-jenkins.kubeme.io/job/K2/job/PR-502/14/

PR #504, opened 2017-06-11
increase the memory sizes for the oft-failing images

coffeepac:
attempts at making the k2 build more reliable.


coffeepac:
did not work at all.

PR #505, opened 2017-06-12
Revisions to check for k2 failures and trigger crash app

leahnp:
if ansible failures occur on up or down, send data to crashapp.
Related: 
The crash app: https://github.com/samsung-cnct/k2-crash-application/pull/1
K2-tools integration: https://github.com/samsung-cnct/k2-tools/pull/57

### DIFF
--- a/down.sh
+++ b/down.sh
@@ -11,4 +11,21 @@ set -o pipefail
 my_dir=$(dirname "${BASH_SOURCE}")
 source "${my_dir}/lib/common.sh"
 
-ansible-playbook ${K2_VERBOSE} -i ansible/inventory/localhost ansible/down.yaml --extra-vars "${KRAKEN_EXTRA_VARS}kraken_action=down" --tags "${KRAKEN_TAGS}"
+# file to capture logs for crash app
+log_file=$"/k2-crash-application/logs"
+
+# check if ansible playbook returned error, if so, send to crash app
+function crash_test {
+	RESULT=$?
+	if [ $RESULT -ne 0 ]; then
+		show_post_cluster_error
+		go run /k2-crash-application/crash-app.go $log_file
+	else
+		show_post_cluster
+	fi
+}
+
+# exit trap for crash app
+trap crash_test EXIT
+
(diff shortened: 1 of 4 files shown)


down.sh
```diff
@@ -11,4 +11,21 @@ set -o pipefail
...
-ansible-playbook ${K2_VERBOSE} -i ansible/inventory/localhost ansible/down.yaml --extra-vars "${KRAKEN_EXTRA_VARS}kraken_action=down" --tags "${KRAKEN_TAGS}"
+# file to capture logs for crash app
+log_file=$"/k2-crash-application/logs"
+
+# check if ansible playbook returned error, if so, send to crash app
+function crash_test {
+	RESULT=$?
+	if [ $RESULT -ne 0 ]; then
+		show_post_cluster_error
+		go run /k2-crash-application/crash-app.go $log_file
```
coffeepac: as noted in the k2-tools repo, this should be executing a binary and not running this codefile as a script.  I'm actually not real sure why we have the go tooling installed in k2-tools but including build tools in a released production artifact is a bad idea.  it opens us up for very strange support requests when end-users are able to alter built artifacts.  and yes, we already have this with ansible being interpreted but its best to not increase the scope of this issue.

leahnp:
On standby until I get Jenkins building a binary for the crash-app

down.sh
joejulian: I've removed the go tooling from k2-tools tools.

joejulian:
Added the `blocked` tag pending k2-tools changes. Please remove the tag when samsung-cnct/k2-tools#57 is merged.

leahnp:
Ready, just waiting on  samsung-cnct/k2-tools#57 to be merged.

lib/common.sh
```diff
@@ -88,6 +88,19 @@ function show_update_error {
...
+# if failure, send to crash app
+function crash_test {
+  RESULT=$?
+  if [ $RESULT -ne 0 ]; then
+    /k2-crash-app/k2-crash-application $log_file
+    show_post_cluster_error 
+  else
+    show_post_cluster
+  fi
+  exit 1
```
coffeepac: this should be `exit $RESULT`.  this line means up.sh and down.sh will always exit with an error code of 1.

leahnp: @coffeepac thanks, revision made

PR #506, opened 2017-06-12
Attempt to stabilize CI runs

coffeepac:
turning some knobs to get our failure rate to be based on kubernetes failures or k2 failures and not apparent jenkins networking issues.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,5 @@
 podTemplate(label: 'k2', containers: [
-    containerTemplate(name: 'jnlp', image: 'jenkinsci/jnlp-slave:3.7-1-alpine', args: '${computer.jnlpmac} ${computer.name}'),
+    containerTemplate(name: 'jnlp', image: 'quay.io/samsung_cnct/custom-jnlp:0.1', args: '${computer.jnlpmac} ${computer.name}'),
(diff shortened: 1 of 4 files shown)


coffeepac:
Top level ticket in Jenkins:  https://issues.jenkins-ci.org/browse/JENKINS-40825 which claims will be fixed by: https://issues.jenkins-ci.org/browse/JENKINS-42048 which is fixed here: https://github.com/jenkinsci/kubernetes-plugin/pull/157

The first link also has a potential workaround which I'll try out tomorrow if I have time.  Its a bit late for this now.

Jenkinsfile
```diff
@@ -106,4 +106,28 @@ podTemplate(label: 'k2', containers: [
     }
   }
 
+def ksh(command) {
```
joejulian: `ksh` is confusing since it already exists. Perhaps kubesh?

coffeepac: do mean 'korn shell exists' or 'ksh is already a jenkins command'?

joejulian: korn shell

coffeepac: nuts.  I was hoping `ksh` was a reserved command and that's why my fix didn't work :(

coffeepac:
hurrah:  https://common-jenkins.kubeme.io/job/K2/view/change-requests/job/PR-506/ 

the goal of this PR was not to get a bunch of jobs to succeed at once, it was to get a bunch of jobs to not fail with obvious signs of timeout or jenkins error.  the sample size to investigate is 10 jobs, job 44 -> 53.  Job 48 did fail right away, my guess is the container hadn't spun up in time for the command to be executed on it.  jobs 51, 52 and 53 failed because there were already too many GKE clusters running at once.  

I believe this PR achieves its goal.

PR #510, opened 2017-06-13
Change default region to us-east-1 and update schema with

### LINKED ISSUE #465: change default aws region to us-east-1
default region is currently us-east-2 for keeping CI stuff siloed.  However, this means as people on our team regenerate their configs they will end up in the CI region.

We should:
- change the default to us-east-1 (or us-west-2, either one)
- alter the CI process to use us-east-2 instead by re-writing the config file after generation
  - just a couple of sed one-liners will do

davidewatson:
documentation and defaults. Fixes #465.

### DIFF
--- a/ansible/roles/kraken.config/files/config.yaml
+++ b/ansible/roles/kraken.config/files/config.yaml
@@ -302,19 +302,19 @@ definitions:
       type: cloudinit
       resourcePrefix:
       vpc: 10.0.0.0/16
-      region: us-east-2
+      region: us-east-1
       subnet:
         -
           name: zone-1
-          az: us-east-2a
+          az: us-east-1a
           cidr: 10.0.0.0/18
         -
           name: zone-2
-          az: us-east-2b
+          az: us-east-1b
           cidr: 10.0.64.0/18
         -
           name: zone-3
-          az: us-east-2c
+          az: us-east-1c
           cidr: 10.0.128.0/17
       egressAcl:
         -
--- a/build-scripts/update-generated-config.sh
+++ b/build-scripts/update-generated-config.sh
@@ -7,9 +7,6 @@ set -x
 
 cluster_name=`echo $2 | tr -cd '[[:alnum:]]-' | tr '[:upper:]' '[:lower:]'`
 
-#  old style configs (can be removed after k2recon is merged)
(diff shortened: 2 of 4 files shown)


ansible/roles/kraken.config/files/config.yaml
```diff
@@ -302,19 +302,19 @@ definitions:
       type: cloudinit
       resourcePrefix:
       vpc: 10.0.0.0/16
-      region: us-east-2
+      region: us-east-1
       subnet:
         -
           name: zone-1
-          az: us-east-2a
+          az: us-east-1a
```
coffeepac: us-east-1a is not available to us.  its a legacy only zone.

davidewatson: I learned today that this is no longer true: "To ensure that resources are distributed across the Availability Zones for a region, we independently map Availability Zones to identifiers for each account."

coffeepac: well I'll be.  Just checked the instances from the first run and there are instances in us-east-1.  

sweet.  I can stop moving everything out of us-east-1a when I'm moving a cluster from us-west-2 to us-east-1

build-scripts/update-generated-config.sh
```diff
@@ -7,8 +7,9 @@ set -x
...
-sed -i -e "s/cluster:/cluster: ${cluster_name}/" $1
-
 #  new style config
-sed -i -e "s/- name:[[:space:]]*$/- name: ${cluster_name}/" $1
\ No newline at end of file
+sed -i -e "s/- name:[[:space:]]*$/- name: ${cluster_name}/" $1
+
+# move regions and AZs to us-east-2. note that this is the CNCT CI region for
+# API rate limit purposes.
+sed -i -e "s/us-east-1/us-east-2/g" $1"
```
coffeepac: there is a trailing double quote on the end of this line.  its causing the script to fail in CI:  https://common-jenkins.kubeme.io/job/K2/view/change-requests/job/PR-510/3/execution/node/26/log/

coffeepac:
failed to destroy GKE cluster.  all other steps passed and nodes ran in us-east-2

PR #512, opened 2017-06-13
Define schema for kubeAuth stanza supporting k8s 1.6 (part of #502).

davidewatson:


schemas/config/v1/kubeAuthAuthnEntry.json
```diff
@@ -6,17 +6,19 @@
...
     "basic": {
-      "description": "Users to be created",
+      "description": "List of user/password to be created.",
       "items": { "$ref": "kubeAuthBasicUser.json" },
       "type": "array"
     },
     "default_basic_user": {
-      "description": "Default user for kubeauth file",
+      "description": "Default user for kubeauth file.",
       "type": "string"
```
mikeln: So, this value `default_basic_user` really has to be one of the `basic[].user` values.  Not a blocking issue, but something to start to think about, how to handle that type of requirement.

schemas/config/v1/kubeAuthAuthzEntry.json
```diff
@@ -0,0 +1,28 @@
...
+      "description": "Kubernetes RBAC configuration.",
+      "properties": {
+        "super_user": {
+          "description": "Determines the authorization-rbac-super-user flag used to start the apiserver in versions of kubernetes <= 1.5",
+          "type": "string",
+          "default": "placeholder"
+        }
+      },
+      "required": [
+        "super_user"
```
mikeln: This is only required pre1.6.  I guess it doesn't hurt in 1.6.  
How does required work exactly?   
My concern: if rbac is off, then authz needs to be: `authz: {}`.  Does this required break that?

schemas/config/v1/kubeAuthBasicUser.json
```diff
@@ -6,11 +6,15 @@
...
     },
     "password": {
-      "description": "The password for the user account",
+      "description": "The password for the user account.",
+      "type": "string"
+    },
+    "group": {
+      "description": "The groups for the user account.",
       "type": "string"
     }
```
mikeln: `group` is only needed for 1.6+.   Also, for this entry, `user`,`password` are required for all versions, `group` is also required for 1.6.

schemas/config/v1/kubeAuthAuthnEntry.json
davidewatson: Draft 5 may help with this, but for now we must check in ansible.

schemas/config/v1/kubeAuthBasicUser.json
davidewatson: Without creating a new kind, we can't make group required some of the time. If there end up being many of these corner cases we might want to do something different than check in ansible.

schemas/config/v1/kubeAuthAuthzEntry.json
davidewatson: Good call. I'll make super_user optional.

davidewatson:
Passed cluster creation for both AWS and GKE: https://common-jenkins.kubeme.io/job/K2/job/PR-511/4/console

davidewatson:
AWS cluster create passed: https://common-jenkins.kubeme.io/job/K2/job/PR-512/

PR #518, opened 2017-06-13
Revert "Dryrun tags bug fixing"

joejulian:
Reverts samsung-cnct/k2#501

This never passed CI and fails dryrun test. Should not have been merged into master.


blackdog0403:
OK should do revert and I will rework

PR #522, opened 2017-06-14
made 1.6.4 default

DStorck:
this makes 1.6.4 the default version of kubernetes.  Temporarily we are removing the central logging chart because we have not yet written the rbac specs for fluentd. This will be covered in the 0.3 milestone, and then it can be added back in.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,7 +13,7 @@ podTemplate(label: 'k2', containers: [
 
             stage('checkout') {
                 checkout scm
-            }    
+            }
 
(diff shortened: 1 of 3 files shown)


leahnp:
LGTM

alejandroEsc:
LGTM, going to try the changes on my local config file and see if that works.

DStorck:
Jenkinsfile has been udpated.

alejandroEsc:
FYI, verified changes worked on my latest config file. good work!

Issue #498, opened 2017-06-09
Add a valid_config command


NathanZook:
Anything which reads the config file needs to 1) interpret it the same way, and 2) be certain that the configuration is valid.  ~~3) The validation step must not create a race condition.~~

The valid_config command will resolve the configuration, and will succeed of fail.

davidewatson:
I like it! Can we use our active voices by renaming the option --validate (for k2/k2cli), and validate_config for internal ansible use? Good English makes for good code (maybe).

NathanZook:
Right now, this feels like validate_config.sh

davidewatson:
I'd prefer `k2/up.sh --validate --config dewMagic.yaml` over `k2/hack/validate_config.sh`, but ¯\_(ツ)_/¯.

alejandroEsc:
--validate-config IMHO.

blackdog0403:
—IMHO, “k2/up.sh —validate —config xxx.yaml” looks good to me for consistency : )

NathanZook:
Sticking this in the review pile for Common Tools.

coffeepac:
downgraded to wait for config file package export from k2 to its own repo

coffeepac:
being done as part of a onboarding golang project.

PR #526, opened 2017-06-14
Added computed_kubectl

NathanZook:
Looking for lots of feedback here--especially fitness issues.  (Documentation, code/file arraingement, coding styles.)


NathanZook:
This should close #498

NathanZook:
This was "inspired by" #498, but has clearly diverged.  This is now the guts of the work of https://github.com/samsung-cnct/k2cli/issues/67

ansible/roles/kraken.config/filter_plugins/max_semver.py
```diff
@@ -0,0 +1,14 @@
+import semver
+
+def max_semver(data):
+    v = max([semver.parse(x[1:]) for x in data])
+    string = semver.format_version(v['major'],v['minor'],v['patch'],v['prerelease'],v['build'])
+    return 'v' + string
+
```
mikeln: Nitpick here.  But the variable name 'v' for version, can potentially cause an issue with and future analysis/modifications as the return string also has a text 'v'.   The variable should be given a meaningful name to avoid future confusion..e.g. `cur_version` or similar.

ansible/roles/kraken.config/tasks/main.yaml
```diff
@@ -129,3 +129,16 @@
   loop_control:
     loop_var: a_cluster
   when: "a_cluster.kubeAuth.authn.oidc is defined"
+
+- name: Collect kubectl versions
+  set_fact:
+    query_max_version: "{{ kraken_config | json_query(version_query) | max_semver }}"
```
coffeepac: where is 'version_query' being defined?

max_version.sh
```diff
@@ -0,0 +1,22 @@
...
+QUERY=$1
+OUTFILE=$2
+shift ; shift
+source "${my_dir}/lib/common.sh"
+
+# setup a sigint trap
+trap control_c SIGINT
+
+VARS_EXPORT_PATH=/tmp/k2_vars.yaml
+ansible-playbook ${K2_VERBOSE} -i ansible/inventory/localhost ansible/max_version.yaml --extra-vars "${KRAKEN_EXTRA_VARS}kraken_action=max_version version_query=${QUERY} version_outfile=${OUTFILE}" --tags "${KRAKEN_TAGS}" || ( echo "valid_config failed" && rm ${VARS_EXPORT_PATH} && exit 1 )
```
coffeepac: why is the json_query being passed in?  do we expect it to change over time?

computed_kubectl.sh
```diff
@@ -0,0 +1,14 @@
...
+#description    :Calls a version of kubectl whose version is computed by ansible from the provided config, passing along the remaining arguments
+#author         :Samsung SDSRA
+#====================================================================
+set -o errexit
+set -o nounset
+set -o pipefail
+
+my_dir=$(dirname "${BASH_SOURCE}")
+VERSIONFILE=/tmp/$$.maxver
+${my_dir}/max_version.sh 'clusters[].nodePools[].kubeConfig.version' ${VERSIONFILE} -c $1
```
coffeepac: tacking onto my other comment, I would prefer if the version query was in k2 right by the query itself.  I don't seen an advantage to making someone look in a different file to understand what the query is doing.

ansible/max_version.yaml
```diff
@@ -0,0 +1,8 @@
+---
+- hosts: localhost
+  roles:
+    - {
+        role: 'kraken.config',
+        tags: [ 'config', 'always', 'config_only' ]
```
coffeepac: please remove the tags.  I see no reason for them other then copy/pasta

ansible/roles/kraken.config/tasks/main.yaml
```diff
@@ -129,3 +129,16 @@
...
+    version_out_path: "{{ version_outfile | realpath }}"
+  when: kraken_action == 'max_version'
+
+- name: Write out version
+  template:
+    src: query_max_version.jinja2
+    dest: "{{ version_out_path }}"
+  when: kraken_action == 'max_version'
+  delegate_to: 127.0.0.1
```
coffeepac: I'm confused why this is required here and it is not required anywhere else in the codebase to render a template to a local path.  including the hostfile should make this work.  can you explain a bit why this is required?

max_version.sh
coffeepac: please remove the 'tags' flag.  there is only one role, it will always be called.

coffeepac: this isn't 'valid_config', its 'max_version', please update the error message.

computed_kubectl.sh
```diff
@@ -0,0 +1,14 @@
...
+#title          :computed_kubectl
+#description    :Calls a version of kubectl whose version is computed by ansible from the provided config, passing along the remaining arguments
+#author         :Samsung SDSRA
+#====================================================================
+set -o errexit
+set -o nounset
+set -o pipefail
+
+my_dir=$(dirname "${BASH_SOURCE}")
+VERSIONFILE=/tmp/$$.maxver
```
coffeepac: will this file be deleted automatically when the script exits?  my bash-fu around tmp files is weak

ansible/roles/kraken.config/tasks/main.yaml
```diff
@@ -129,3 +129,16 @@
   loop_control:
     loop_var: a_cluster
   when: "a_cluster.kubeAuth.authn.oidc is defined"
+
+- name: Collect kubectl versions
```
coffeepac: ~~kubectl~~ kubernetes

max_version.sh
```diff
@@ -0,0 +1,22 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# pull in utils
+my_dir=$(dirname "${BASH_SOURCE}")
+
+QUERY=$1
+OUTFILE=$2
+shift ; shift
```
coffeepac: is there a reason for these shifts?  I don't see any more accesses of command line parameters after this

ansible/roles/kraken.config/tasks/main.yaml
NathanZook: Actually, this is not about kubernetes versions any more at all.  I forgot to correct the name.

ansible/roles/kraken.config/tasks/main.yaml
NathanZook: Command line.  You can see the call in max_version.sh

max_version.sh
NathanZook: lib/common.sh does its normal command line manipulations.

computed_kubectl.sh
NathanZook: I was thinking in terms of the container being deleted.  Of course, this does not HAVE to be run in a container.  I will delete the file.

computed_kubectl.sh
NathanZook: The old name in the task was misleading.  The query that Yeni will be running will be different.

max_version.sh
NathanZook: I'm sneaking ahead just a little here.  Yeni will be using this code with a different query.

ansible/max_version.yaml
NathanZook: But...that was done in the best tradition of cargo cultists everywhere!  Not having enough experience with tags to know what they were doing, I left them in.  Wilco.

ansible/roles/kraken.config/filter_plugins/max_semver.py
NathanZook: I also forgot to run pep8 across the file.

NathanZook:
This requires that semver be installed in the quay k2-tools image.  It appears that it is not.

ansible/roles/kraken.config/tasks/main.yaml
NathanZook: Huh? We need to communicate information back to the calling task.  This is on the same host.  The only ways that I have found to do this are local_action and delegate_to.  I was unable to get local_action to work in this case.

ansible/roles/kraken.config/filter_plugins/max_semver.py
mikeln: there is no need to optimize for character space...
and my issue isn't really with the single letter temporary var, it is that it is 'v'.   Which can cause confusion with the quoted 'v' in the return.  An untrained eye future eye, may make assumptions and possibly think the quoted 'v' is wrong or the variable 'v' is wrong or inconsistent...and then make some catastrophic change (I've seen it happen many times).  Rename that var 'v' to something else.

ansible/roles/kraken.config/tasks/main.yaml
coffeepac: this appears to be rendering a template to local disk.   I believe that if you remove this line, it will work as expected, ie:  the template query_max_version.jinja2 will be rendered to {{ version_out_path }}.  

is there something I'm missing?

max_version.sh
coffeepac: okay.

computed_kubectl.sh
coffeepac: what query will Yeni be running?  I'm trying to think it through and I'm not seeing it.

max_version.sh
coffeepac: what does this have to do with a valid config file?

is the tags option going away?

ansible/roles/kraken.config/tasks/main.yaml
coffeepac: hokay, thanks.

ansible/roles/kraken.config/tasks/main.yaml
NathanZook: If I remove delegate_to, then the template will be rendered on some remote server.  I want this to render on localhost.

computed_kubectl.sh
NathanZook: She is getting the latest helm version, which I assume is somewhere in the config object.

ansible/roles/kraken.config/tasks/main.yaml
coffeepac: there is no remote server in this configuration.  the only server is localhost.  should be safe to drop it.

computed_kubectl.sh
coffeepac: helm versions are accessed by the kubernetes version that the helm version is built for.  so to access the appropriate version of helm, the code needs to:
- find the major.minor version of kubernetes being used
- access /opt/cnct/kubernetes/<version>/bin/helm

So it should be the same query.

ansible/roles/kraken.config/tasks/main.yaml
NathanZook: That's...convenient.  And successful.

ansible/roles/kraken.config/tasks/main.yaml
```diff
@@ -129,3 +129,17 @@
   loop_control:
     loop_var: a_cluster
   when: "a_cluster.kubeAuth.authn.oidc is defined"
+
```
joejulian: Do we care if the action is max_version? Can we just do this every time? Or maybe just exclude `'down'`?

I can imagine cases where it may behoove us to use the fact elsewhere.

computed_kubectl.sh
NathanZook: I saw ` ansible/roles/kraken.services/filter_plugins/latest_helm.py`, and expected it to be used somehow.  Will degeneralize.

ansible/roles/kraken.config/tasks/main.yaml
NathanZook: It was necessary to exclude it when the query was being passed in.  As it is no longer being passed in, there is no requirement to exclude it.

NathanZook:
Python semver module is now in the k2-tools image.

coffeepac:
the CI failures look legit.  please take a look at their output.

NathanZook:
Yes, I missed the fact that I was defining two facts there.  The second was really a debug thing, so I've corrected the issue.

NathanZook:
Passed on https://common-jenkins.kubeme.io/job/K2/job/PR-526/17/.  I did a rebase, which kicked off https://common-jenkins.kubeme.io/job/K2/job/PR-526/18

coffeepac:
I accept the passed run/rebase comment.  Onwards.

Issue #541, opened 2017-06-16
Namespace Dance Uneeded.


mikeln:
#537 continued.  

Remove the other 2 places of "Namespace Dancing" in the fabric role.

Add namespace.yaml to canal and weave.  Add step to "apply" that file.  That should create the namespace if it is missing, or do nothing if it is already there.

Issue #540, opened 2017-06-16
Wait for API Server Superfluous


mikeln:
The reorganization of ansible role order needed for RBAC should always guarantee that the kubernetes apis server is available.  The startup has to wait for kubectl to be viable (hence that the api server is available) to install the needed system RBAC before anything else can occur.   

The api server should be available for anything after that step.

The 'Wait for API server' tasks after that step should be able to be removed.

mikeln:
plus all of these steps where broken during k2recon, and not fixed.  The current do nothing.

davidewatson:
A list of existing apiserver waits:
```
rodin:k2 dwat$ grep -ir 'wait for api' ansible
ansible/roles/kraken.rbac/tasks/build-access.yaml:- name: Wait for api server response before continuing
ansible/roles/kraken.readiness/tasks/do-wait.yaml:- name: Wait for api server to become available in case it is not
ansible/roles/kraken.services/tasks/run-dns.yaml:- name: Wait for api server to become available in case it is not already
ansible/roles/kraken.services/tasks/run-services.yaml:- name: Wait for api server to become available in case it's not
rodin:k2 dwat$
```

davidewatson:
For reference, role dependency ordering [documentation](http://docs.ansible.com/ansible/playbooks_roles.html#role-dependencies).

PR #542, opened 2017-06-16
#541 remove namespace dance and extra api up check

mikeln:
#541 and partial #540 

Add namespace.yaml and create using apply.
Remove api server up check


mikeln:
And, yes, I realize this could be refactored to the fabric.selector task instead.  I didn't want to change too much at once. A refactor of what is here would be an easy starting task for someone.

ansible/roles/kraken.fabric/kraken.fabric.weave/templates/namespace.yaml.part.jinja2
```diff
@@ -0,0 +1,4 @@
+piVersion: v1
```
DStorck: more pi!

Issue #508, opened 2017-06-12
Clean up helmOverride


davidewatson:
As mentioned [here](https://github.com/samsung-cnct/k2/pull/497) the schema for helmOverride is not as useful as it could because the type of helmOveride may be a boolean ([22 different ways](http://yaml.org/type/bool.html)) when specified by the config file, a string when specified by an environment variable, or undefined when using an older configuration file.

I believe the root of this problem is that helmOverride has two distinct purposes:

 - Allow potentially incompatible versions of helm to be used anyway
 - Allow helm to be skipped entirely

The latter of these purposes has been requested by other customers, and I think the easiest way to accommodate this behavior is to allow the user to remove the helmConfig line from the cluster definition within the configuration file. For example:

```
deployment:
  clusters:
    - name: dewMaybe
...
      fabricConfig: *kubeVersionedFabric
      kubeAuth: *defaultKubeAuth
      #helmConfig: *defaultHelm
      dnsConfig: *defaultDns
...
```

Since the stanza is commented out, helm should be skipped...

For the former purpose, to proceed or not when there is no supported version of helm for the chosen kubernetes version, a boolean will suffice. If it is true, then use the latest version of helm available, even if it does not support the requested version of k8s. If it is false, then fail if the requested version of kubernetes does not have a supported version of helm. The failure message should indicate that the config file or an environment variable can be specified to proceed with an unsupported version of helm.

In this scheme, to support old versions of the configuration file, by default the helmOverride value should be false.

PR #544, opened 2017-06-19
Documentation for K2 node management and services

guineveresaenger:
This should cover all documentation for K2 in regards to node management. This is meant to cover issue #230 .

We should make sure to document upcoming changes to k2cli as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -428,7 +428,7 @@ You can specify different versions of Kubernetes in each nodepool. This may affe
 docker run $K2OPTS quay.io/samsung_cnct/k2:latest ./update.sh --config $HOME/.kraken/${CLUSTER}.yaml --nodepools clusterNodes,specialNodes
 ```
 
-### Kubernetes versioning for K2 services
+## Kubernetes versioning for K2 services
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -396,56 +396,55 @@ clusters:
 ```
 - etcd settings (beyond machine type)
 
-### Things that can be changed (sometimes with some manual intervention)
+*Warning*
+Repeatedly running `./up.sh` or `./update.sh` can result in etcd nodes being destroyed and regenerated automatically, which will break your cluster's control plane. This is because in the K2 generated config file, the CoreOS version is set to `current`. When CoreOS releases a new version, Terraform will register this change and destroy and re-create etcd nodes. We are working on fixing this, but please be aware that this is a possible issue.
```
davidewatson: This behavior is destined to cause problems. Maybe we should default to the current latest `1353.8.0` and forgo this warning? I had to make a similar change to the original kraken...

davidewatson:
LGTM. We need to structure some of the documentation, and possibly some of the flags (cf. https://github.com/samsung-cnct/k2/issues/508), but those items should not hold up this PR, IMO. The faster we merge code, the fewer merge conflicts we will have...

README.md
guineveresaenger: @davidewatson : @coffeepac and I talked about this on Friday. We have plans for this, though I couldn't point you at the specific issues right now. As far as I understand, quickfix is to place etdc in its own Terraform folder, which will freeze its CoreOS version. Longterm fix is to make sure etcd gets updates safely.

guineveresaenger:
Could someone approve and merge?

README.md
coffeepac: I'll open a fresh issue for this.  Its a good discussion to have.

PR #532, opened 2017-06-15
Add $$target extension used by sphinx-jsonschema to create links.

davidewatson:
Also add a few defaults and modify many descriptions.

Issue #376, opened 2017-05-01
release coordination between k2-tools, k2 and k2cli


coffeepac:
the k2-tools, k2 and k2cli projects should have clear versioning strategies.  this strategy should have the following characteristics:
- versions should be in semver
- versions should be pushed by CNCT operated CI
- nothing should use a :latest tag by default

coffeepac:
There are two flows of releases here:
1. CI flow, ie a new k2-tools build should trigger a k2 build which, if successful, should trigger a k2cli build
2. named release build

The first flow is fairly easy, for every successful PR merged, kick off a downstream build.  This will quickly smoke out issues without waits.  This should be added to the individual job issues.  \<add issues for all three here>  These can all use the :latest tag.

For the second flow I am less sure of the process.  A quick search of "jenkins build tag" shows that building tags is slightly annoying.  It might be easier to push tags from jenkins after its been built (GitPublisher seems to be the key here).  That would suggest the following general workflow for each project:
- a human decides to cut a new release
- click <build now> on the specified build
- this opens a parameterized page that allows for selection of source branch (default to master) and revision to build (default to x.x.x+1, ie increment patch level by one).
- click 'submit'

this will build and push the project to wherever it goes (quay or GitHub).

For k2 and k2cli, this implies there is an additional field on their build pages which is the tag of the upstream source (k2-tools for k2 and k2 for k2cli) to use.  This will default to the latest tag available.  

For k2cli this will mean changing the code to have the k2 default image version be a build-time variable (shouldn't be hard).  
For k2 this will mean changing the Dockerfile to have the FROM value be a parameter (might be tricky, but easy enough to push the Dockerfile through a filter that swaps the image tag from latest to \<whatever>).

This will allow us to rev the version of each component independently and then easily compose them together for releases.  

ToDo:
- discover if this tag default work possible in Jenkins right now
- consider using JJB (or other jenkins job templating engine) to supply the parameterization
- fill in missing issue list above

coffeepac:
Discussion point:  re-write default config file to pin coreOS version as part of release job.  Would this need to be pushed back into the github repo or would it be sufficient to make a note in some kind of auto-generated CHANGELOG?

davidewatson:
Why limit pinned CoreOS versions to released versions? If there is a CI job which tests each release, each passing CoreOS version bump can/should push the latest tested version into the default config.

coffeepac:
[...]

PR #545, opened 2017-06-19
Initial sphinx documentation for jsonschema.

davidewatson:


leahnp:
Is there anyway to preview this?

davidewatson:
This is the infrastructure part of the k2 documentation.

@leahnp a preview is available here: https://davidewatson.github.io/k2/. Sadly, many changes to it are necessary as part of separate PRs (e.g. https://github.com/samsung-cnct/k2/pull/532). The reason for keeping the PRs separate is that this change (to Sphinx with reStructuredText _and_ Markdown) is potentially contentious.

One this is merged, the Source setting for "GitHub Pages" needs to be updated to "docs" here: https://github.com/samsung-cnct/k2/settings

NathanZook:
I'm seeing a lot of overflow of the cells by the terms being defined.  For some reason, the items are list items inside each cell, which causes noisy unordered list activity, and may be part of why the overflow is occurring.  I'm concerned that the interface list is ungrouped and so long.  The fixed width presentation is ugly & fixing that at the table setup level would prevent the overflow issue noted above.  Also, I LOVE our logo!

NathanZook:
The "state" and "events" entries in the first table are missing their data.  The AWS node configuration table contains an invalid entry.  AWS Node Provider Configuration table has more of these.  (And, it is erroneously labeled "AWS Provider Configuration").  There appears to be some sort of systematic breakdown in the presentation.

davidewatson:
The preview will not be accurate until #532 is merged. Items like breaking up the jsonschema into multiple pages are organizational changes I think should be fixed separately, lest the deliverable become too sweeping for a single PR or person to address.

Makefile
```diff
@@ -0,0 +1,29 @@
...
+SPHINX_SOURCE_DIR = ./sphinx
+SPHINX_BUILD_DIR  = ./docs
+
+# Put it first so that "make" without argument is like "make help".
+help:
+	@grep -h "##" $(MAKEFILE_LIST) | fgrep -v fgrep | sed -e 's/\\$$//' | sed -e 's/##//'
+
+.PHONY: help Makefile
+
+gh-pages: ## Build gh-pages documentation, manual push required
```
coffeepac: says manual push, can it be automated or does it require a bunch of human interaction?

sphinx/conf.py
```diff
@@ -0,0 +1,162 @@
...
+author = u'Samsung CNCT'
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+version = u'0.1'
+# The full version, including alpha/beta/rc tags.
+release = u'0.1'
```
coffeepac: how do these values get set?  is it via the Makefile or hand editing?

coffeepac:
Would be nice to rebase this to pick up the CI fix from this morning.  Better chance of going green.

Makefile
davidewatson: It can be (e.g. `git add -u && git commit -m 'Rebuilt GitHub pages documentation using Sphinx' && git push`), but I was worried about this leading to unexpected results. For example, suppose the repo is not clean? Should we check out the commit/release/etc. and build from that? Many people do automate this, so maybe I am wrong-headed here.

sphinx/conf.py
davidewatson: They are currently meaningless. We have a couple of options: http://dreamiurg.net/2011/10/03/using-git-to-get-version-information/ or http://sphinx-git.readthedocs.io/en/latest/ I was thinking about his in the context of https://github.com/samsung-cnct/k2/issues/376 last night. I think we just need to commit to tagging each release a la https://help.github.com/articles/creating-releases/

Makefile
coffeepac: so long as its possible.  lets file an issue for automating this so that the pages are always up-to-date with... actually, should be up to date with latest built artifact or latest commit to master?  those aren't, sadly, the same things right now.

sphinx/conf.py
coffeepac: if we used sphinx-git would you be inclined to use it everywhere?  ie generate CHANGELOGs and whatnot in sphinx?

davidewatson: Ever documentation system I've looked at ([doxygen](http://www.stack.nl/~dimitri/doxygen/), [doca](https://github.com/cloudflare/doca), etc.) has it's warts. Sphinx is used by the Linux kernel, Python, etc., so though I may live to regret it, I think Sphinx is a good start.

PR #550, opened 2017-06-19
Wait for apiserver readiness exactly once (fixes #540).

### LINKED ISSUE #540: Wait for API Server Superfluous

davidewatson:
@mikeln I think the reason the reason you needed to add a sperate readiness check for RBAC is that the rbac role was being run before the readiness role. Note that this change may cause the rbac role to wait longer than it normally would (because readiness usually waits for more than 1 node to be up).

### DIFF
--- a/ansible/roles/kraken.readiness/tasks/do-wait.yaml
+++ b/ansible/roles/kraken.readiness/tasks/do-wait.yaml
@@ -26,7 +26,7 @@
   set_fact:
     api_servers: "{{ lookup('file', kubeconfig) | from_yaml | json_query('clusters[*].cluster.server') }}"
 
-- name: Wait for api server to become available in case it is not
+- name: Wait for api server to become available
   wait_for:
     host: "{{ item | regex_replace('https://','') }}"
     port: 443


davidewatson:
The DO_NOT_MERGE label is (mostly) because I am working on an analogous etcd check. It doesn't technically have to wait if there is agreement.

davidewatson:
Reopening without RBAC changes.

mikeln:
The order of things matters.
The RBAC apiserver wait has to be first, because the default RBAC has to be installed before the nodes can finish coming up and connect.  

That is why it has to be before readiness.  readiness only looks for enough nodes to be running.

ansible/up.yaml
```diff
@@ -30,14 +30,14 @@
...
-      }
-    - {
         role: 'roles/kraken.readiness',
         tags: [ 'readiness', 'services', 'all' ]
       }
     - {
+        role: 'roles/kraken.rbac',
+        tags: [ 'readiness', 'services', 'all', 'dryrun', 'rbac_only' ]
+      }
+    - {
```
mikeln: No, this has to be before readiness

ansible/roles/kraken.rbac/tasks/build-access.yaml
```diff
@@ -6,36 +6,5 @@
...
-
-- name: Wait for api server response before continuing
-  command: >
-    {{ kubectl }} --kubeconfig={{ kubeconfig | expanduser }} get nodes --no-headers=true
-  register: rbac_get_nodes_result
-  until: rbac_get_nodes_result.stdout_lines | length >= 1 | int
-  retries: "{{ repetitions | int }}"
-  delay: "{{ rep_interval | int }}"
-  when: "(not (dryrun | bool))"
-
```
mikeln: No, this has to be here.

PR #552, opened 2017-06-20
K2 image test. DO NOT MERGE

NathanZook:


davidewatson:
This was for testing the k2-tools build only.

PR #557, opened 2017-06-21
Checking clusters status before spin up clusters

blackdog0403:
IMHO, we'd better do conservative approach on **updating cluster**.

**Therefore, when the cluster is already up, let the up.sh command fail**, 
then inform users to use update.sh to update cluster.

As far as I concerned, using config_force option for both checking schema and cluster status looks not clear and adding an another option for this could cause more complexity on ux.


ansible/roles/kraken.config/tasks/check-clusters-status.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - name: Check cluster status when two files exist
+    command: >
+      {{ kubectl }} --kubeconfig={{ kubeconfig }} get nodes --no-headers=true
+    register: job_get_nodes_result
+    when:  kubeconfig is defined and terraform_file.stat.exists == True
+
+  - name: Fail when node count is greater than or equal to 1
+    fail:
+      msg: >-
+        {{cluster.name }} has been already spinned up. Please use 'update.sh' to update cluster(s).
```
davidewatson: Should the message be `up.sh has already been run for {{ cluster.name }}. Please use 'update.sh' to update cluster(s).`?

ansible/roles/kraken.config/tasks/check-clusters-status.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+  - name: Execute appropriate kubectl per minor version
+    set_fact:
+      kubectl: "/opt/cnct/kubernetes/{{ kubernetes_minor_version }}/bin/kubectl"
+
+  - name: Check cluster status when two files exist
+    command: >
+      {{ kubectl }} --kubeconfig={{ kubeconfig }} get nodes --no-headers=true
+    register: job_get_nodes_result
+    when:  kubeconfig is defined and terraform_file.stat.exists == True
```
davidewatson: ~~I think it's possible for `kubeconfig is __undefined__` () and `terraform_file.stat.exists == True`. In this case, do we still want to proceed?~~ I'm wrong. kubeconfig is a fact set above.

How much of `up` has to complete before we prohibit it from running again? We need to make sure this point is sufficiently far along that `update` can carry on. The more I think about it, the more I wonder why we are doing this.

ansible/roles/kraken.config/tasks/check-clusters-status.yaml
```diff
@@ -0,0 +1,34 @@
+---
+  - name: Check status of cluster
+    set_fact:
+      cluster: "{{ a_cluster }}"
+
+  - name: Check that the admin.kubeconfig state file exists
```
davidewatson: This task does not `Check that the admin.kubeconfig state file exists`. It just sets a fact. Is this a mistake?

ansible/roles/kraken.config/tasks/check-clusters-status.yaml
DStorck: Also, there is probably a better way to say 'spinned up'.

ansible/roles/kraken.config/tasks/check-clusters-status.yaml
blackdog0403: That's right, I changed logic just before commit then forgot change comment on it.

ansible/roles/kraken.config/tasks/check-clusters-status.yaml
blackdog0403: Any suggestion? @davidewatson @DStorck 
Maybe "{{cluster.name}} is running at the moment?

ansible/roles/kraken.config/tasks/check-clusters-status.yaml
blackdog0403: @davidewatson IMO, we'd better stop the job as less as it completes, I mean, It becomes time that we could recognize cluster status, just we can stop the job. So I put check-clusters-status.yaml after checking configuration file.

And I need to understand what happen to etcd nodes cluster if we spin up cluster again when it is up. Could @coffeepac explain in detail?

coffeepac: if we run up against etcd nodes that already exist and we specified 'latest' as the CoreOS version then we run the risk of destroying and recreating all etcd nodes with run of up (or update for that matter).  The problem is that specifying 'latest' gets passed into the generated terraform configuration file which means that terraform decides what version to use.  Terraform will also say 'latest' so if a version of CoreOS has been released between runs of 'up' then the current etcd cluster(s) get rebuilt and no data is saved.  This destroys the running kubernetes cluster.

coffeepac: @davidewatson I think the point where we prevent up from running is if admin.kubeconfig exists.  update can't function without it.  

I am a bit torn if we should prevent up from running if the tfstate file exists.  on one hand, if that file exists then we have a chance of doing the etcd mind wipe thing.  on the other hand, if the admin.kubeconfig file doesn't exist, do we have a valid cluster?

coffeepac:
... removing from above thread...

Should we do this?  I think we should as running up.sh against a live cluster can do two negative things:
- destroy your cluster by mauling the etcd nodes
- create longer term oddness by rebuilding the ASG configs and not rebuilding the nodePools those ASGs control

I think preventing, or at least making the user aware, of both of those are important topics.

ansible/roles/kraken.config/tasks/check-clusters-status.yaml
```diff
@@ -0,0 +1,35 @@
+---
+  - name: Check status of cluster
```
coffeepac: this stanza is setting the fact `cluster`, the name only applies to this stanza, not the whole task.

ansible/roles/kraken.config/tasks/check-clusters-status.yaml
joejulian: Yet we still may not have run helm either way. If it's already broken, do we care if it breaks again? I say either just just add a `file:` to the end of kraken.services, or add a kraken.post role that just creates the file as the very last thing, then check to see if that one single file exists in up.sh.

coffeepac: oh, fair point.   that's probably easiest and means we don't need to guess if the network or DNS is up and running.  I vote for @joejulian 's idea of dropping a file after services is done.  I"m not super happy about adding a new state file to the system but it works well.

davidewatson: My curiosity was mostly rhetorical, sorry.

I spent some time this morning attempting to verify that every task which occurs after kubeconfig and tfstate exist will be handled properly by `update`. I realized that even if I succeeded, there would be no guarantee that tomorrow someone won't add another task which must be accounted for. The only full proof way is to ensure that `update` can proceed, is to create a "lock" file at the end of the entire `up`. Which is fair, just not very ansibly.

joejulian: What about pushing some data into a configmap and querying that?

davidewatson: Do we really think that up can't be (easily) made idempotent? I'm OK with the state file course, but honestly wonder if we wouldn't be better off fixing the CoreOS version (since this should be done anyway), and then checking if any work remains.

davidewatson: @joejulian: Or etcd?

coffeepac: idempotency only claims that for a constant input + action the same result will be achieved.  The config file (input) is mutable.  We could make up check to see if stuff is created and if it is run the upgrade process to handle the ASG thing but that seems counter intuitive.

I agree we need to fix the etcd thing.  That's a bad thing.

davidewatson:
@coffeepac: I meant that attempting to determine if it is safe to run `up` by looking for the existence of a couple important files doesn't seem plausible. Half of me wants to fix the known problems and see if we really need to introduce state. The other half understands that fixing this might involve a long tail of PRs, so maybe we'd be better off with a simple state file, at least for now.

NathanZook:
I'm not really sure what idempotency means for a process that can fail part way through.  My instinct is that `up` should be resumable.  Now the problem with this that @coffeepac pointed out is that the desired state can change with time even if the configuration inputs do not.  I would argue that this is the error.  We have a race, and it must be eliminated.

Before initiating any actions to bring up a cluster, `up` needs to write out all of the data necessary to uniquely create the cluster in the desired state.  This should include how the resolutions were obtained.  (ie: coreos:latest -> coreos:1f2e3d4c).  As the process proceeds, items may be checked off as completed.

Up may fail at either end--some error might be reported when attempting to create or configure some resource; or the `up` process might come down with a case of `kill -9`.  In either case, I would argue strongly that a user would expect that `up` function in a resume fashion.  But these two cases have different "expected" results.  In the second case, we would like to just skip to wherever we were and proceed.  In the first case, there is a good chance that the configuration will be deliberately changed in an attempt to correct the error.

My instinct would be to do the resolution, and query the user if there is a difference between part of the previous desired state that has already been implemented and the currently resolved desired state.  Note that whether we implement a checkoff or not, we might query the live cluster to see its state and determine how much of the previously desired state has been realized.

I anticipate from previous systems that the user query would likely to be..."rich".

davidewatson:
https://github.com/samsung-cnct/k2/pull/566 addresses the potential etcd node churn caused by drifting coreOS versions.

Solving the ASG rebuild problem looks to be harder.

```
# Launch configurations for all node pools
{% for node in cluster.nodePools %}
resource "aws_launch_configuration" "{{node.name}}_launch_config" {
...
  iam_instance_profile        = "${aws_iam_instance_profile.kubernetes_node_profile.name}"
  user_data                   = "${file("{{ config_base | expanduser }}/{{cluster.name}}/cloud-config/{{node.name}}.cloud-config.yaml")}"
...
```

Due to user_data, I am not sure terraform can protect us from mutating ASGs.

ansible/roles/kraken.config/tasks/check-clusters-status.yaml
blackdog0403: @coffeepac  ~~Isn't it playbook scope? I referenced k2/ansible/roles/kraken.config/tasks/setup.yaml~~ Oooh, I get it now, it mean 'name'. Removed it.

blackdog0403:
@davidewatson @coffeepac @NathanZook @joejulian I spent time on understanding your opinions and try to figure it out **what should we do on this PR.**.

1. Checking cluster(s) status through admin.kubeconfig and terraform.tfstate is **inappropriate**
-> Adding a 'kraken.post' to after kraken.service just creates a file that indicates 'cluster is a desired state'. We could prevent users to destroy etcd nodes through 'up.sh' at the moment.

2. Seperate 'idempotency' issue ('state' issue) from this PR then add an new issue for it
-> #566  is still on disccussion.

Can #566 solve etcd nodes destroying problem? then we don't need to check cluster status befroe running up.sh.

coffeepac:
this was discussed in standup, but to sum up:  yes to both of your points.  onward sir.

NathanZook:
I think #566 can fix one source etcd cluster destruction.  There or not others exist is a larger question.

blackdog0403:
Actually It was worthing being stuck in this issue. Learned a lot!

ansible/roles/kraken.post/tasks/mark-cluster-status.yaml
```diff
@@ -0,0 +1,13 @@
+---
+  - name: Touch cluster.status.lock file in config_base
+    file:
+      path: "{{ config_base | expanduser }}/cluster.status.lock"
+      state: touch
+      mode: "u=rw,g=r,o=r"
+
+  - name: Set cluster.status.lock as Variables
+    set_fact:
+      cluster_status_lock: "{{ config_base | expanduser }}/cluster.status.lock"
```
davidewatson: This path is a duplicate of the one above.

up.sh
```diff
@@ -11,6 +11,14 @@ set -o pipefail
 my_dir=$(dirname "${BASH_SOURCE}")
 source "${my_dir}/lib/common.sh"
 
+inf "\nChecking cluster status..."
+if [ -f ${KRAKEN_BASE}/cluster.status.lock ]; then
```
davidewatson: By default, `KRAKEN_BASE="${HOME}/.kraken"`. Shouldn't this be checking `{{ config_base | expanduser }}/{{ cluster.name }}` from within ansible?

ansible/roles/kraken.post/tasks/mark-cluster-status.yaml
blackdog0403: Fixed it!

up.sh
blackdog0403: @davidewatson Yes, KRAKEN_BASE is set by --output parameter so that it indicates parent directory of each cluster directory. And this checking logic is still valid for multicluster config.yaml file due to `kraken.post role` makes `cluster.status.lock' file when whole tasks are successful including kraken.services. **At the moment, it regards multiple cluster creating jobs as a unit for idempotency.** 
I think, we'd better seperate each cluster's lock file when other task are prepared for it.

PR #558, opened 2017-06-21
updated k2 to 1.6.6

DStorck:


joejulian:
Added a "blocked" tag as gke does not support 1.6.6 (as of the last jenkins test). Clear the blocked when gke adds 1.6.6 support.

DStorck:
@joejulian That is weird because according to https://cloud.google.com/container-engine/release-notes, 1.6.6 is supported

joejulian:
Maybe re-run the test. Perhaps that support has changed since the test was run.

joejulian:
I dunno what to tell you about those guys.
> ERROR: (gcloud.container.clusters.upgrade) ResponseError: code=400, message=bad desired node version ().

joejulian:
gke passed so I'd say this is no longer blocked.

PR #566, opened 2017-06-22
Fix coreOs version to ensure the tested version is the same one the

davidewatson:
user gets by default.

### DIFF
--- a/ansible/roles/kraken.config/files/config.yaml
+++ b/ansible/roles/kraken.config/files/config.yaml
@@ -167,7 +167,7 @@ definitions:
       name: defaultCoreOs
       kind: os
       type: coreOs
-      version: current
+      version: 1409.2.0
       channel: stable
       rebootStrategy: "off"
   nodeConfigs:


joejulian:
:-1: I realize I'm in the minority here, but I'm really not a fan of adding purely administrative work to development. I'd much rather, especially for testing, that we stay on the latest and leave it to the user to pin a version if they so choose.

NathanZook:
May I have it both ways?  I think for reasons of stability, it makes sense to pin versions for releases.  As @coffeepac pointed out in the discussions of https://github.com/samsung-cnct/k2/pull/557, "latest" tag means that the builds are non-deterministic.  From the standpoint of troubleshooting, we make our lives more complicated when we pass "latest" along.  Certainly, the users need to have the option to override such a thing.  On "edge", we want "latest", if for no other reason than to make our lives easier as a development organization.

joejulian:
I'd be fine pinning it for a tagged release.

alejandroEsc:
I like this personally, and i think its wise to encourage a fixed version regardless. I would go on further and say that latest/current cause an error because of the danger such a choice poses. One question though: Do we test on versions and have an internal 'accepted list'? We could simply warn customers they have selected an unsupported version and let them continue at their own risk.

davidewatson:
This fixes a known problem. Until https://github.com/samsung-cnct/k2/issues/376 is implemented, I don't see any other choice. Sans testing and process, software versions must be pinned for stability. Even with the appropriate CI/CD pipeline, the coreOs version specified for the etcd nodes can not currently be changed without downtime.

We can (and should) dedicate resources to improving our CI to remove or reduce this administrative burden in the future. Remember, there are more users in the world than developers.

venezia:
With the advent of the CoreOS container Linux operator ( https://coreos.com/blog/introducing-container-linux-update-operator ) - there is little reason not to stay up to date with the OS, especially since there may be security implications in not updating. 

Not being compliant with latest would be a problem that needs to be sorted out right away, at least if there is a security risk. So I'd rather that we know asap when our builds aren't going to work with the latest release and always try to pursue the environments being on the latest version. (As well as minimizing accrued technical debt of "fixing" things to be compliant. 

Therefore, latest (stable channel) is probably always what we want, even if it isn't classically deterministic.  It also, like joe mentions, reduces our involvement. We only look into it when it breaks, which shouldn't be often. 

Mike

alejandroEsc:
@venezia in principle i would agree, but we are talking about an OS, should security updates etc.. not be managed by some internal update tool? Almost sounds like we are treating the OS  as an application, but im not sure i like that idea in general.

davidewatson:
How we CI the latest release, and how coreOs is updated are both orthogonal questions. We need to improve our CI process and test operators. Until then, we need something that is stable and works.

@venezia: Note that the current default for coreOs reboot strategy is "off".

coffeepac:
Snipped from slack:  

> there are two things that need to happen:  pinned version for releases (which we don't have), a CI stream for latest that makes sure latest CoreOS doesn't hurt anything

#376 has the discussion for how to manage this within CI and releases going forward.  I believe, unless someone objects, this PR can go forward. 

any last comments or discussions on this?

NathanZook:
Won't this PR fix the coreos version for us?

coffeepac:
Yes.  And until we have a CI process that updates it in some manner we will need to manually update it.

PR #569, opened 2017-06-22
Weave-net yaml problem, addressed because of known and reported apiserver issue 

alejandroEsc:
details: https://github.com/kubernetes/kubernetes/issues/2763

### DIFF
--- a/ansible/roles/kraken.fabric/kraken.fabric.weave/templates/v1.6/weave.yaml.part.jinja2
+++ b/ansible/roles/kraken.fabric/kraken.fabric.weave/templates/v1.6/weave.yaml.part.jinja2
@@ -75,7 +75,7 @@ spec:
             - name: IPALLOC_RANGE
               value: {{fabricConfigVal.options.network.network}}
             - name: CONN_LIMIT
-              value: {{fabricConfigVal.options.network.nodeConnectionLimit}}
+              value: '{{fabricConfigVal.options.network.nodeConnectionLimit}}'
           livenessProbe:
             initialDelaySeconds: 30
             httpGet:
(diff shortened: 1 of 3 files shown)


ansible/roles/kraken.fabric/kraken.fabric.weave/templates/v1.6/weave.yaml.part.jinja2
```diff
@@ -75,7 +75,7 @@ spec:
             - name: IPALLOC_RANGE
               value: {{fabricConfigVal.options.network.network}}
             - name: CONN_LIMIT
-              value: {{fabricConfigVal.options.network.nodeConnectionLimit}}
+              value: '{{fabricConfigVal.options.network.nodeConnectionLimit}}'
```
davidewatson: You are making this a string, but then the schema says it's a number. Which is it?

alejandroEsc: its a number, please look at the issue i mentioned in the description.

alejandroEsc:
jenkins says: 'container not found ("k2-tools")' huh?

coffeepac:
the k2-tools container took too long to fetch so the CI job just tried to execute anyway.  it shouldn't happen but its a bug.  I restarted the job.

ansible/roles/kraken.fabric/kraken.fabric.weave/templates/v1.6/weave.yaml.part.jinja2
coffeepac: okay, so this is a number where it gets used, but to the config file (as specified in the bug) we need to pass a string.  so for K2, this should be a string.

mikeln: I wouldn't really look at this as a "conversion" of type.   

It is really a "limit" on the values are allowable...i.e. a) the value is a number, b) the minimum value is 1, c) default 30.  It is easier to specify these limits as a number, rather than a text regex.

I vote that it is a number in reality, and our schema should be a such.   The required input by the destination is just an artifact of that destination.

davidewatson: Agreed. (FWIW, the only number we quote in our default YAML config is "-1". In this case, it is necessary, because the field is really an enum).

PR #570, opened 2017-06-22
Rename kraken.*.docker roles since they are not specific to docker.

davidewatson:


ansible/roles/kraken.nodePool/kraken.nodePool.selector/tasks/main.yaml
```diff
@@ -3,16 +3,16 @@
...
 
 - name: master nodePools
   include_role:
-    name: '{{ playbook_dir }}/roles/kraken.nodePool/kraken.master.docker'
+    name: '{{ playbook_dir }}/roles/kraken.nodePool/kraken.nodePool.master'
 
 - name: node nodePools
   include_role:
-    name: '{{ playbook_dir }}/roles/kraken.nodePool/kraken.node.docker'
+    name: '{{ playbook_dir }}/roles/kraken.nodePool/kraken.nodePool.node'
```
mikeln: nit pick here... but the role is already `kraken.nodePool`. why have `nodePool` in the sub-role name?  And on that same line, is `kraken` even needed there also?  Couldn't the names just be `master`, `node`, `etcd`?

davidewatson: I thought the same thing. We use this convention elsewhere, presumably to avoid naming collisions with hierarchical roles. Since ansible has a non trivial search order for roles, I think it makes sense, even if it is ugly.

davidewatson: The kraken bit, BTW, is so we can theoretically lift this role out and play it from another project. Or course, our copious use of facts makes that highly unlikely.

joejulian:
This is the third fail in exactly the same spot for the same reason. I think it's legit.

davidewatson:
@joejulian: Thanks for follow up. It turns out, the third time was the charm.

PR #586, opened 2017-06-29
make subnet part of pre-existing VPC config

### LINKED ISSUE #583: BYO VPC change, bring subnets, not route tables
if a user is bringing their own VPC, they should also provide the subnets.  

Steps:
- add the subnets to the existing_vpc config option
- set the aws `providerConfigs.subnet` key to optional
- set the aws `providerConfigs.existing_vpc.route_table_id` to optional (and note as deprecated in comment)
- convert the subnet aws template to be conditional in the manner of the vpc template

coffeepac:
force user to supply subnet when BYOing a VPC.  

set other subnet option to be optional (don't need both)
set route_table_id as optional and set description to be DEPRECATED.

this closes #583

### DIFF
--- a/schemas/config/v1/awsPreexistingSubnetConfig.json
+++ b/schemas/config/v1/awsPreexistingSubnetConfig.json
@@ -1,0 +1,23 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "http://judkins.house/apis/k2/v1/awsPreexistingSubnetConfig.json",
+  "$$target": "awsPreexistingSubnetConfig",
+  "title": "AWS Subnet Identification",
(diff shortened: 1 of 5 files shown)


schemas/config/v1/awsPreexistingSubnetConfig.json
```diff
@@ -0,0 +1,23 @@
...
+    "name": {
+      "description": "The name of the subnet, for use within k2 configurations.",
+      "type": "string"
+    },
+    "id": {
+      "description": "AWS id of the pre-existing subnet.",
+      "type": "string"
+    }
+  },
+  "required": [
```
DStorck: is this saying that "if" you choose to include a VPC, then these things are required?

PR #588, opened 2017-06-29
taints and tolerations added for master nodepool

DStorck:
removes the register-unschedulable for masters.  now static pods are aligned with master nodes because of taints and tolerations

### DIFF
--- a/ansible/roles/kraken.nodePool/kraken.nodePool.master/templates/v1.6/api-server-manifest.yaml.jinja2
+++ b/ansible/roles/kraken.nodePool/kraken.nodePool.master/templates/v1.6/api-server-manifest.yaml.jinja2
@@ -77,7 +77,6 @@ spec:
       path: /usr/share/ca-certificates
     name: ssl-certs-host
   tolerations:
-  - key: dedicatedMaster
+  - key: node-role.kubernetes.io/master
     operator: Equal
-    value: masters
     effect: NoSchedule
--- a/ansible/roles/kraken.nodePool/kraken.nodePool.master/templates/v1.6/api-server-manifest.yaml.jinja2
+++ b/ansible/roles/kraken.nodePool/kraken.nodePool.master/templates/v1.6/api-server-manifest.yaml.jinja2
@@ -76,3 +76,8 @@ spec:
   - hostPath:
       path: /usr/share/ca-certificates
     name: ssl-certs-host
+  tolerations:
+  - key: dedicatedMaster
+    operator: Equal
+    value: masters
+    effect: NoSchedule
--- a/ansible/roles/kraken.nodePool/kraken.nodePool.master/templates/v1.6/controller-manager-manifest.yaml.jinja2
+++ b/ansible/roles/kraken.nodePool/kraken.nodePool.master/templates/v1.6/controller-manager-manifest.yaml.jinja2
@@ -47,3 +47,8 @@ spec:
   - hostPath:
       path: /usr/share/ca-certificates
     name: ssl-certs-host
+  tolerations:
+  - key: dedicatedMaster
+    operator: Equal
+    value: masters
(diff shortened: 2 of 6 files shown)


coffeepac:
@DStorck you'll need to rebase against master to pass CI.

ansible/roles/kraken.nodePool/kraken.nodePool.master/templates/v1.6/api-server-manifest.yaml.jinja2
```diff
@@ -76,3 +76,8 @@ spec:
   - hostPath:
       path: /usr/share/ca-certificates
     name: ssl-certs-host
+  tolerations:
```
venezia: IIRC, taints/tolerances are not applied to static pods.  Therefore this may be nice, but not needed.

ansible/roles/kraken.nodePool/kraken.nodePool.master/templates/v1.6/controller-manager-manifest.yaml.jinja2
```diff
@@ -47,3 +47,8 @@ spec:
   - hostPath:
       path: /usr/share/ca-certificates
     name: ssl-certs-host
+  tolerations:
```
venezia: Same comment as API Server - I don't see similar adjustments for `kube-proxy` - am I missing it or did you not add a similar tolerance?

ansible/roles/kraken.nodePool/kraken.nodePool.master/templates/v1.6/controller-manager-manifest.yaml.jinja2
```diff
@@ -47,3 +47,8 @@ spec:
...
+  - key: dedicatedMaster
```
davidewatson: I think we should follow kubeadm's [example](https://kubernetes.io/docs/setup/independent/create-cluster-kubeadm/#master-isolation) here. This will makes us consistent with more of the kubernetes community and make it easier to transition to kubeadm internally if we decide to do so.

DStorck: So I'm seeing that they use `kubectl taint nodes --all node-role.kubernetes.io/master-`,  but I'm not sure how to get this to do exactly what I want. The [docs](https://kubernetes.io/docs/user-guide/kubectl/v1.7/#taint) say you need a key, value, and effect. The key would be `node-role.kubernetes.io/master-`, and I can put in a value of `NoSchedule`, but I'm not sure how to match kubeadm's example for the value.

DStorck: will do.

ansible/roles/kraken.nodePool/kraken.nodePool.master/templates/v1.6/api-server-manifest.yaml.jinja2
coffeepac: cluster didn't come up very well without the tolerations added.  adding the tolerations made things much smoother.

ansible/roles/kraken.nodePool/kraken.nodePool.master/templates/v1.6/controller-manager-manifest.yaml.jinja2
coffeepac: kube-proxy was added.

coffeepac:
pushing this over the line with @venezia still a blocking review as its late and we need this in.

@venezia feel free to holler at me for any additional work you want done as a part of this. happy to do it but we gotta go.

PR #596, opened 2017-07-05
Enhance command line processing

joejulian:
Just some simple command line option parsing changes to support long options and add some help text.

Also a minor change to allow a manually configured MAPPINGS environment variable to be used to manually add additional `-v source:dest` docker switches.

### DIFF
--- a/hack/dockerdev
+++ b/hack/dockerdev
@@ -14,6 +14,19 @@ realpath() {
     echo ${REALDIR}/${REALFILE}
 }
 
+show_help() {
+    echo " Usage:"
+    echo "     $0 [options]"
+    echo ""
+    echo " Options:"
+    echo "     -c,--config <filename>    config.yaml file to use"
+    echo "                               default: $HOME/.kraken/default.yaml"
+    echo "     -h,--help                 this text"
+    echo "     -i,--image   <imagename>  k2-tools docker image to use"
+    echo "                               default: quay.io/samsung_cnct/k2-tools"
+    echo "     -m,--home <path>          specify the path to be used for \$HOME"
+}
+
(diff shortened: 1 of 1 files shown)


hack/dockerdev
```diff
@@ -23,11 +36,18 @@ EOF
...
 fi
 
-while getopts c:i:r: opt; do
-    case $opt in
-        c) CONFIGFILE="${OPTARG}";;
-        i) IMAGE="${OPTARG}";;
-        r) MYHOME="${OPTARG}";;
+# Parse command line options
+GETOPT=$(getopt -q -o c:fhi:m: --long config:,fresh,help,image:,home: -n $0 -- "$@")
+eval set -- "$GETOPT"
```
mikeln: fresh?   not mentions in the usage or the parsing.

hack/dockerdev
```diff
@@ -48,7 +68,7 @@ if ! [[ -f ${CONFIGFILE} ]]; then
 fi
 
 CONFIGFILE=$(realpath ${CONFIGFILE})
-MAPPINGS="-v ${CONFIGFILE}:${CONFIGFILE}"
+MAPPINGS="$MAPPINGS -v ${CONFIGFILE}:${CONFIGFILE}"
```
mikeln: so you are allowing MAPPINGS brought in from the environment?

joejulian: Yes, it came up a couple weeks ago when someone wanted to map a particular file through to their dev environment. It can be done just by adding dummy entries to the config file, but this seemed less hacky.

hack/dockerdev
joejulian: That's what I get for doing three changes at once, then realizing I should have done them as three separate PRs. :grin:

Issue #576, opened 2017-06-26
allow user to add/delete node pools from config file and have this update a running cluster


coffeepac:
Use Case:  a user would like to add a new nodepool to a running cluster.  They add the nodepool to the config file, run `update` and the new nodepool is created.

Use Case:  a user would like to delete a nodepool from a running cluster.  They remove the nodepool from the config file, run `update` and the nodepool is deleted.

For both use cases, master nodepools should not be allowed to be created or deleted (detect this by checking for apiServer config key).

This may be done except for master nodepool mangling prevention.

guineveresaenger:
Preliminary results:
- Adding a nodepool will happen automatically when running update. The only issue with this is that the update action fails if no nodepools were passed. So the new nodepool will quietly get added whilst running update on a _different_ nodepool. I am not sure what happens if you explicitly pass the new nodepool. 
- Next steps: see if a nodepool can be removed .
- I believe it is possible currently to add master nodepools, although why somebody would want to do so is puzzling. There is definitely some work to be done there.

guineveresaenger:
It turns out nodepools can and will be removed, due to the up action being included in the cluster.
If the nodepool-to-be-removed was passed as an argument to the update command, it will still fail awkwardly:
`TASK [/kraken/ansible/roles/kraken.update/kraken.update.aws : Fail when nodepools passed do not exist] **************************************
failed: [localhost] (item=extraSpecialNodes) => {"failed": true, "item": "extraSpecialNodes", "msg": "You passed a nodepool that does not exist in this cluster. Please make sure to pass --nodepools <comma,separated,list,of,nodepools> to update action."}`

Summary:

The basic functionality for adding and deleting nodepools is there; I recommend some explicit documentation and also perhaps tweaking error/status messages a bit to better reflect that nodepools could have also been added or deleted. But, given that adding and deleting nodepools does not need to trigger the cordon/drain/terminate/regenerate process, this is mostly good to use as is, with the exception that one should not be able to add or remove master nodepools.

coffeepac:
@joejulian is this a feature you need right nowish?

joejulian:
Thanks, @guineveresaenger. 
@coffeepac: It is, yes.

coffeepac:
@guineveresaenger can you get started on this then?  I'll put it on the board.

guineveresaenger:
@joejulian @coffeepac - to clarify:
[...]

PR #600, opened 2017-07-06
Creates flag in K2 for adding and removing nodepools

guineveresaenger:
EDITED TO ADD: This addresses #576 and forces the user to explicitly pass any nodepools they wish to add or remove to the update action.

ToReview (and this should probably be passed into CI job if I’m not mistaken):

Bring up a cluster.

Run `./update.sh` in the following ways and check for appropriate message/behavior:

no arguments
`--addnodepools master` 
`--rmnodepools master`
`--addnodepools <nodepool that already exists>`
`--rmnodepools <bogus nodepool that doesn’t exist>`
`--rmnodepools <nodepool that exists>`, but don’t remove from config file
Remove a nodepool from config file but don’t pass it to `--rmnodepool` flag (give it a nodepool that doesn’t exist to satisfy the command line field)
`--addnodepools <new nodepool>` but don’t add it in config file
Add a nodepool to your config but don’t specify in` --addnodepools flag` (give it a nodepool that already exists, just to satisfy the flag field)

All of the above should fail and do nothing to your cluster.

To actually affect your cluster, run `./update.sh` in these 2 ways:

Add a new nodepool to your config file and then run `--addnodepools <new nodepool>`.
Remove a nodepool from your config file and then run `--rmnodepools <removed nodepool>`.

You should always have to specify which nodepool you are adding or removing.

You should also be able to pass multiple flags and add or remove multiple node pools.

One thing this doesn’t address is the fact that one can still change the configurations of a nodepool (like counts of nodes) in the config file, and K2 will still quietly do it.


coffeepac:
@joejulian since @guineveresaenger is on holiday for three weeks, if any changes are needed here I will do them.  just ping me.

ansible/roles/kraken.update/kraken.update.aws/tasks/main.yaml
```diff
@@ -4,13 +4,14 @@
   set_fact:
     nodepools_to_update: "{{ update_nodepools.split(',') }}"
 
-- name: Fail when no nodepools were passed to update
-  fail:
-    msg: "Argument empty. Please pass --nodepools <comma,separated,list,of,nodepools> to update action."
-  when: update_nodepools == ""
```
joejulian: Please remove instead of commenting.

ansible/roles/kraken.update/kraken.update.gke/tasks/main.yaml
```diff
@@ -4,10 +4,10 @@
   set_fact:
     nodepools_to_update: "{{ update_nodepools.split(',') }}"
 
-- name: Fail when no nodepools were passed to update
-  fail:
-    msg: "Argument empty. Please pass --nodepools <comma,separated,list,of,nodepools> to update action."
-  when: update_nodepools == ""
```
joejulian: same

ansible/roles/kraken.config/tasks/check-update.yaml
```diff
@@ -0,0 +1,123 @@
...
+    {{ kubectl }} --kubeconfig={{ kubeconfig }} get nodes -o jsonpath='{range .items[*]}{.metadata.labels.nodepool}{"\n"}{end}'
+  register: current_nodepools
+
+- name: Create list of current nodepool names
+  set_fact:
+    current_nodepools: "{{ current_nodepools.stdout_lines | unique }}"
+
+- name: Make empty list for non control plane nodepool information
+  set_fact:
+    config_nodepool_data: []
```
davidewatson: The name of this task is incorrect as the variable contains master nodes.

ansible/roles/kraken.config/tasks/check-update.yaml
```diff
@@ -0,0 +1,123 @@
...
+
+- name: Create list of names of nodepools to remove
+  set_fact:
+    nodepools_to_remove: "{{ remove_nodepools.split(',') }}"
+
+- name: Create list of names of nodepools to add
+  set_fact:
+    nodepools_to_add: "{{ add_nodepools.split(',') }}"
+
+- name: Do not allow control plane nodepool to be removed
```
davidewatson: The name of this task is incorrect. It also serves to prevent nodepools from being added.

ansible/roles/kraken.config/tasks/check-update.yaml
coffeepac: good point.  the master nodepool is intended to be excluded.  I will update.

ansible/roles/kraken.update/kraken.update.gke/tasks/main.yaml
coffeepac: removed (upcoming commit)

ansible/roles/kraken.update/kraken.update.aws/tasks/main.yaml
coffeepac: removed (upcoming commit)

ansible/roles/kraken.config/tasks/check-update.yaml
coffeepac: name updated.

davidewatson:
LGTM after the remaining code is removed.

ansible/roles/kraken.config/tasks/check-update.yaml
coffeepac: actually, just checked the upstream issue, and only etcd nodepools are intended to be excluded.  the name is wrong but the logic is correct.

Issue #595, opened 2017-07-05
Check path to k2-crash-app logfile exists


leahnp:
Check path to k2-crash-app logfile exists, in common.sh. For the case that someone is not using k2-tools and does not have that path, so it will not crash on them.

davidewatson:
Note that this can also cause an error when the developer is using an older version of k2-tools:

```
bash-4.3# ./up.sh --config ~/.kraken/dewThinkDifferently.yaml 
WARNING: --output not specified. Using /root/.kraken as location
WARNING: all not specified. Using 'all' as tags
tee: /k2-crash-app/logs: No such file or directory
```

mikeln:
I would recommend adding `if [ -f "$log_file" ];then`  tee'd version `else` non-tee'd version `fi`
in up/down/upgrade

mikeln:
Ugg, but you need the "initial file" there hmm... maybe just execute `touch $log_file` no matter what...won't change the file if it is already there (though it will update the timestamp on it...which shouldn't matter as we will be adding to it right away also), but will create a zero length one if missing.

The return ( `$?` ) should be non zero if the file couldn't be created, in which case you would want to avoid running `|tee`...

leahnp:
To review, check the following PR's address the concerns and are merged:

https://github.com/samsung-cnct/k2/pull/601
https://github.com/samsung-cnct/k2-tools/pull/81

mikeln:
everything is in and is working.

PR #601, opened 2017-07-06
Error handling, separate functions for up/down/update

leahnp:
Added graceful error handling for cases where someone does not have k2-crash-app installed or log file is not in expected location 

Created 3 functions to deal with up/down/update separately, since each one calls different functions depending on success or failure - at this point I think it would be less confusing and less code to just put the k2-crash-app trigger inline in the up.sh/down.sh/update.sh, like I originally had it, but let me know what you think.

### DIFF
--- a/down.sh
+++ b/down.sh
@@ -12,9 +12,16 @@ my_dir=$(dirname "${BASH_SOURCE}")
 source "${my_dir}/lib/common.sh"
 
 # capture logs for crash app
-log_file=$"/k2-crash-app/logs"
+crash_app_dir=$"/usr/bin/k2-crash-app"
+logs=$"logs"
+log_file=$crash_app_dir/$logs
 
 # exit trap for crash app
-trap crash_test EXIT
+trap crash_test_down EXIT
 
-ansible-playbook ${K2_VERBOSE} -i ansible/inventory/localhost ansible/down.yaml --extra-vars "${KRAKEN_EXTRA_VARS}kraken_action=down" --tags "${KRAKEN_TAGS}" | tee $log_file
+# check crash application directory exists
+if [ -d "$crash_app_dir" ]; then
+	ansible-playbook ${K2_VERBOSE} -i ansible/inventory/localhost ansible/down.yaml --extra-vars "${KRAKEN_EXTRA_VARS}kraken_action=down" --tags "${KRAKEN_TAGS}" | tee $log_file
+else 
+	ansible-playbook ${K2_VERBOSE} -i ansible/inventory/localhost ansible/down.yaml --extra-vars "${KRAKEN_EXTRA_VARS}kraken_action=down" --tags "${KRAKEN_TAGS}"
+fi
--- a/down.sh
+++ b/down.sh
@@ -3,7 +3,7 @@
 #description     :bring down a kraken cluster
 #author          :Samsung SDSRA
 #==============================================================================
-set -o errexit
+# set -o errexit
 set -o nounset
 set -o pipefail
 
@@ -12,16 +12,14 @@ my_dir=$(dirname "${BASH_SOURCE}")
 source "${my_dir}/lib/common.sh"
 
 # capture logs for crash app
-crash_app_dir=$"/usr/bin/k2-crash-app"
-logs=$"logs"
-log_file=$crash_app_dir/$logs
+LOG_FILE=$"/tmp/crash-logs"
 
 # exit trap for crash app
 trap crash_test_down EXIT
 
-# check crash application directory exists
-if [ -d "$crash_app_dir" ]; then
-	ansible-playbook ${K2_VERBOSE} -i ansible/inventory/localhost ansible/down.yaml --extra-vars "${KRAKEN_EXTRA_VARS}kraken_action=down" --tags "${KRAKEN_TAGS}" | tee $log_file
-else 
+K2_CRASH_APP=$(which k2-crash-application)  
+if [ $? -ne 0 ];then  
 	ansible-playbook ${K2_VERBOSE} -i ansible/inventory/localhost ansible/down.yaml --extra-vars "${KRAKEN_EXTRA_VARS}kraken_action=down" --tags "${KRAKEN_TAGS}"
+else
+	ansible-playbook ${K2_VERBOSE} -i ansible/inventory/localhost ansible/down.yaml --extra-vars "${KRAKEN_EXTRA_VARS}kraken_action=down" --tags "${KRAKEN_TAGS}" | tee $LOG_FILE
 fi
--- a/down.sh
+++ b/down.sh
@@ -3,6 +3,8 @@
 #description     :bring down a kraken cluster
 #author          :Samsung SDSRA
 #==============================================================================
+# k2-crash-app is taking over the EXIT in common.sh functions called in the trap
+# leaving this commented out in case we missed an edge case
 # set -o errexit
 set -o nounset
 set -o pipefail
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -94,7 +94,7 @@ function crash_test_up {
   RESULT=$?
   if [ $RESULT -ne 0 ]; then
     if [ -f "$K2_CRASH_APP" ]; then
-      /usr/bin/k2-crash-application $LOG_FILE
+      ${K2_CRASH_APP} ${LOG_FILE}
     else 
       echo "k2-crash-application not found, to capture the data from k2 failures, please install"
     fi
@@ -111,7 +111,7 @@ function crash_test_down {
   RESULT=$?
   if [ $RESULT -ne 0 ]; then
     if [ -f "$K2_CRASH_APP" ]; then
-      /usr/bin/k2-crash-application $LOG_FILE
+      ${K2_CRASH_APP} ${LOG_FILE}
     else 
       echo "k2-crash-application not found, to capture the data from k2 failures, please install"
     fi
@@ -125,7 +125,7 @@ function crash_test_update {
   RESULT=$?
   if [ $RESULT -ne 0 ]; then
     if [ -f "$K2_CRASH_APP" ]; then
-      /usr/bin/k2-crash-application $LOG_FILE
+      ${K2_CRASH_APP} ${LOG_FILE}
     else 
       echo "k2-crash-application not found, to capture the data from k2 failures, please install"
     fi
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -88,18 +88,61 @@ function show_update_error {
   exit 1
 }
 
-# check if ansible return failure
+# check if ansible return failure on up
 # if failure, send to crash app
-function crash_test {
+function crash_test_up {
   RESULT=$?
   if [ $RESULT -ne 0 ]; then
-    /k2-crash-app/k2-crash-application $log_file
-    show_post_cluster_error 
+    # check the logs file exists
+    if [ -f "$log_file" ]; then
+      /usr/bin/k2-crash-app/k2-crash-application $log_file
+      show_post_cluster_error 
+    else 
+      echo "k2-crash log file not found...ignoring"
+      exit $RESULT
+    fi
   else
     show_post_cluster
   fi
   exit $RESULT
 }
+
+# check if ansible return failure on down
+# if failure, send to crash app
+function crash_test_down {
+  RESULT=$?
+  if [ $RESULT -ne 0 ]; then
+    # check the logs file exists
+    if [ -f "$log_file" ]; then
+      /usr/bin/k2-crash-app/k2-crash-application $log_file
+    else 
+      echo "k2-crash log file not found...ignoring"
+      exit $RESULT
+    fi
+  fi
+  exit $RESULT
+}
+
+# check if ansible return failure on update
+# if failure, send to crash app
+function crash_test_update {
+  RESULT=$?
+  if [ $RESULT -ne 0 ]; then
+    # check the logs file exists
+    if [ -f "$log_file" ]; then
+      /usr/bin/k2-crash-app/k2-crash-application $log_file
+      show_update_error
+    else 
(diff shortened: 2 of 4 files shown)


leahnp:
This is tied to issue: https://github.com/samsung-cnct/k2/issues/595

down.sh
```diff
@@ -12,9 +12,16 @@ my_dir=$(dirname "${BASH_SOURCE}")
 source "${my_dir}/lib/common.sh"
 
 # capture logs for crash app
-log_file=$"/k2-crash-app/logs"
+crash_app_dir=$"/usr/bin/k2-crash-app"
+logs=$"logs"
+log_file=$crash_app_dir/$logs
```
mikeln: You should not store logs in /usr/bin.   The executable is fine in that location, but not the log data...by unix/linux standards.
The log data should go in another appropriate location...e.g.  `/var/log`.   
However, if the intention is for that file to exist longer than the container, it would need to go onto a mounted filesystem..as user could specify that.  So I would recommend `/var/log/k2-crash-app` as the default location.

down.sh
```diff
@@ -12,9 +12,16 @@ my_dir=$(dirname "${BASH_SOURCE}")
...
 
 # exit trap for crash app
-trap crash_test EXIT
+trap crash_test_down EXIT
 
-ansible-playbook ${K2_VERBOSE} -i ansible/inventory/localhost ansible/down.yaml --extra-vars "${KRAKEN_EXTRA_VARS}kraken_action=down" --tags "${KRAKEN_TAGS}" | tee $log_file
+# check crash application directory exists
+if [ -d "$crash_app_dir" ]; then
```
mikeln: So, you would want to check the log directory  (which should not be the app dir location).

lib/common.sh
```diff
@@ -88,19 +88,62 @@ function show_update_error {
...
 # if failure, send to crash app
-function crash_test {
+function crash_test_up {
   RESULT=$?
   if [ $RESULT -ne 0 ]; then
-    /k2-crash-app/k2-crash-application $log_file
-    show_post_cluster_error 
+    # check the logs file exists
+    if [ -f "$log_file" ]; then
+      /usr/bin/k2-crash-app/k2-crash-application $log_file
```
mikeln: Try not to use the absolute path of the application.  Since it is in `/usr/bin` which should be in the $PATH of the env, it would be more flexible to `which k2-crash-application` to get the location.  
Also, this is common to all these routines, it would be wise to make a common function (or even globally assigned var) for K2_CRASH_APP, and use that variable instead of hard coded paths.  
So either `K2_CRASH_APP=$(which k2-crash-application)` or `K2_CRASH_APP=/usr/bin/k2-crash-application`.  

However, in the case where the user is running a container that is missing that app,  you would still want to detect that situation.

lib/common.sh
```diff
@@ -88,19 +88,62 @@ function show_update_error {
...
+      exit $RESULT
+    fi
+  else
+    show_update
+  fi
+  exit $RESULT
+}
+
+
+
```
mikeln: same comments to each of these functions.

up.sh
```diff
@@ -15,9 +15,16 @@ source "${my_dir}/lib/common.sh"
 trap control_c SIGINT
 
 # capture logs for crash app
-log_file=$"/k2-crash-app/logs"
+crash_app_dir=$"/usr/bin/k2-crash-app"
+logs=$"logs"
+log_file=$crash_app_dir/$logs
```
mikeln: same comments as above on down.sh

update.sh
```diff
@@ -19,11 +19,18 @@ source "${my_dir}/lib/common.sh"
 trap control_c SIGINT
 
 # capture logs for crash app
-log_file=$"/k2-crash-app/logs"
+crash_app_dir=$"/usr/bin/k2-crash-app"
+logs=$"logs"
+log_file=$crash_app_dir/$logs
 
 # exit trap for crash app
```
mikeln: same comments as in down.sh

down.sh
leahnp: The intention is not for the file to last longer than the container, the logs are held for the duration of up/down/update -> if there is a failure -> they are sent to elasticsearch -> container is killed

mikeln: then `/var/log` is the correct choice.

lib/common.sh
leahnp: As I understand the `which` command, it only can provide a path for applications in your PATH - since I don't add k2-crash-application to the path, `which k2-crash-application` returns nothing when ran in k2-tools...so I don't understand the benefit of that command in this context.

down.sh
leahnp: I think maybe the name "logs" is confusing, they aren't really logs for a program...they are just temp data that will always die with the container. So maybe it's better to put them in a tmp directory?

mikeln: sure, that works also.

lib/common.sh
mikeln: /usr/bin is usually in $PATH so it will find k2-crash-application.

mikeln: There are 2 intents here:
1) Use a variable so you only specify the location in one spot. 
2) Use which to find it...mainly to determine if it is even in the image.  (e.g. If I run `which` on a non-upgraded container...it will return an error indication, so we know that the crash app isn't present).  [I realize this will become a moot point in the near future as everyone migrates to the new(er) k2-tools container...but it is still a good shell programming practice].

leahnp: I figured out why `which` was not working for me...it was a directory not a file.

davidewatson: To address 1. above, it suffices to set a variable `k2-crash-application=/usr/bin/k2-crash-app/k2-crash-application` once, and then use it thereafter. For 2. you could put the variable declaration in an if statement to avoid the error on exit behavior of `set -o errexit`. For example,

```
if k2-crash-application=$(which k2-crash-application); then
  :
else
  echo "k2-crash-application not found, ignoring..."
fi
```

davidewatson: I am unsure how great that my suggestion is though. You would still have to avoid calling `${k2-crash-application}` forever after, or call a wrapper which ignores errors. Part of me thinks you are on your own if you do not use a support k2-tools image. ¯\_(ツ)_/¯

mikeln:
This one will not pass without: https://github.com/samsung-cnct/k2/pull/601 
and that one will not pass without this one merged.

leahnp:
@mikeln just made another commit that should address all the concerns.

lib/common.sh
```diff
@@ -93,12 +93,11 @@ function show_update_error {
...
-    if [ -f "$log_file" ]; then
-      /usr/bin/k2-crash-app/k2-crash-application $log_file
+    if [ -f "$K2_CRASH_APP" ]; then
+      /usr/bin/k2-crash-application $LOG_FILE
       show_post_cluster_error 
     else 
-      echo "k2-crash log file not found...ignoring"
+      echo "k2-crash-application not found, to capture the data from k2 failures, please install"
       exit $RESULT
     fi
```
mikeln: I think we want to display: `show_post_cluster_error` in both cases.

lib/common.sh
```diff
@@ -93,12 +93,11 @@ function show_update_error {
...
-    # check the logs file exists
-    if [ -f "$log_file" ]; then
-      /usr/bin/k2-crash-app/k2-crash-application $log_file
+    if [ -f "$K2_CRASH_APP" ]; then
+      /usr/bin/k2-crash-application $LOG_FILE
       show_post_cluster_error 
     else 
-      echo "k2-crash log file not found...ignoring"
+      echo "k2-crash-application not found, to capture the data from k2 failures, please install"
       exit $RESULT
```
mikeln: This `exit` is not needed.  The script will exit properly with the final `exit`.  It is generally best to have 1 exit point if possible.  plus that plays into the next comment..

lib/common.sh
```diff
@@ -93,12 +93,11 @@ function show_update_error {
 function crash_test_up {
   RESULT=$?
   if [ $RESULT -ne 0 ]; then
-    # check the logs file exists
-    if [ -f "$log_file" ]; then
-      /usr/bin/k2-crash-app/k2-crash-application $log_file
+    if [ -f "$K2_CRASH_APP" ]; then
+      /usr/bin/k2-crash-application $LOG_FILE
       show_post_cluster_error
```
mikeln: Don't need this here... move it after the `fi`.  I think we want to display `show_post_cluster_error` in every case.

lib/common.sh
```diff
@@ -128,12 +126,11 @@ function crash_test_down {
 function crash_test_update {
   RESULT=$?
   if [ $RESULT -ne 0 ]; then
-    # check the logs file exists
-    if [ -f "$log_file" ]; then
-      /usr/bin/k2-crash-app/k2-crash-application $log_file
+    if [ -f "$K2_CRASH_APP" ]; then
+      /usr/bin/k2-crash-application $LOG_FILE
       show_update_error
```
mikeln: same comment as in up().   I think we want `show_update_error` in every case.  Move this to after the `fi` for this.

lib/common.sh
```diff
@@ -128,12 +126,11 @@ function crash_test_down {
...
     else 
-      echo "k2-crash log file not found...ignoring"
+      echo "k2-crash-application not found, to capture the data from k2 failures, please install"
       exit $RESULT
```
mikeln: remove this exit.  relay on the final exit.

lib/common.sh
```diff
@@ -88,19 +88,59 @@ function show_update_error {
...
+# check if ansible return failure on down
+# if failure, send to crash app
+function crash_test_down {
+  RESULT=$?
+  if [ $RESULT -ne 0 ]; then
+    if [ -f "$K2_CRASH_APP" ]; then
+      /usr/bin/k2-crash-application $LOG_FILE
+    else 
+      echo "k2-crash-application not found, to capture the data from k2 failures, please install"
+      exit $RESULT
```
mikeln: remove this exit... the final exit is enough.

up.sh
```diff
@@ -3,7 +3,7 @@
 #description     :bring up a kraken cluster
 #author          :Samsung SDSRA
 #==============================================================================
-set -o errexit
+# set -o errexit
 set -o nounset
```
leahnp: Whatever ansible returns is trapped and passed to the crash_test_x functions in common.sh - as far as I can tell, they are handled the same as pre-crash-reporter integration.

lib/common.sh
leahnp: amended

lib/common.sh
leahnp: amended

lib/common.sh
leahnp: amended

lib/common.sh
leahnp: amended

lib/common.sh
leahnp: amended

lib/common.sh
leahnp: amended

lib/common.sh
```diff
@@ -88,19 +88,56 @@ function show_update_error {
...
+  exit $RESULT
+}
+
+# check if ansible return failure on update
+# if failure, send to crash app
+function crash_test_update {
+  RESULT=$?
+  if [ $RESULT -ne 0 ]; then
+    if [ -f "$K2_CRASH_APP" ]; then
+      /usr/bin/k2-crash-application $LOG_FILE
```
mikeln: sorry missed this:  should be: `${K2_CRASH_APP} ${LOG_FILE}`  
we do not want the hard coded paths.

lib/common.sh
```diff
@@ -88,19 +88,56 @@ function show_update_error {
...
   exit $RESULT
 }
 
+# check if ansible return failure on down
+# if failure, send to crash app
+function crash_test_down {
+  RESULT=$?
+  if [ $RESULT -ne 0 ]; then
+    if [ -f "$K2_CRASH_APP" ]; then
+      /usr/bin/k2-crash-application $LOG_FILE
```
mikeln: sorry missed this: should be: ${K2_CRASH_APP} ${LOG_FILE}
we do not want the hard coded paths.

lib/common.sh
```diff
@@ -88,19 +88,56 @@ function show_update_error {
...
-# check if ansible return failure
+# check if ansible return failure on up
 # if failure, send to crash app
-function crash_test {
+function crash_test_up {
   RESULT=$?
   if [ $RESULT -ne 0 ]; then
-    /k2-crash-app/k2-crash-application $log_file
+    if [ -f "$K2_CRASH_APP" ]; then
+      /usr/bin/k2-crash-application $LOG_FILE
```
mikeln: sorry missed this: should be: ${K2_CRASH_APP} ${LOG_FILE}
we do not want the hard coded paths.

up.sh
mikeln: Perhaps add a comment above this line, explaining that we are taking over EXIT processing.  But leaving this here in case we have missed a corner case...

leahnp: Sure thing

lib/common.sh
leahnp: got it

lib/common.sh
leahnp: got it

lib/common.sh
leahnp: got it

leahnp:
Thanks Mikel - am I allowed to merge this myself?

mikeln:
Sorry, no.  I'll merge the 2 once this one goes through (as it should now that all this handling of missing files is in).

leahnp:
Got it, thanks @mikeln for your help with this one

mikeln:
Well, CI is having issues today.  However, I just saw that this worked, as in when the anible error occurred, the check for crash-app worked correctly and the whole thing didn't just abort out like it had been.  That is at least 1 good test for this change.

PR #602, opened 2017-07-06
Fix for 1999 version of getopt on macs

joejulian:


guineveresaenger:
@mikeln I never ran into trouble but I know some folks did using the dockerdev.

joejulian:
I just wanted to add long options. My ocd was bothering me.

blackdog0403:
LGTM

PR #603, opened 2017-07-06
Remove calico fabric type since it is unused.

davidewatson:
Note that calico is used as part of our canal fabric type configuration, which is part of the flannel role. Sigh.


alejandroEsc:
isnt calico a frabric on its own right?

davidewatson:
It is. And canal is a combination of calico and flannel. The naming situation needs to be cleaned up. This is the first step in that direction.

alejandroEsc:
Then why eliminate it? as an option?

davidewatson:
I actually think leaving it does hurt things. I had to spend five minutes rediscovering what calico as a configuration option does. This was compounded by the fact that even explaining what flannel as a configuration option does it annoying. ;)

But I relent.

alejandroEsc:
@davidewatson i think it would be worth opening a ticket that mentions what you want to address.

coffeepac:
@mikeln I think having a non implemented network fabric is a confusing user experience.  We have canal and weave types implemented so we can show we are multi (cni) network capable.  Do you have a further reason to keep this no op around?   We can always bring it back if we want to support calico on its own.

mikeln:
e2e test failures do not appear to be fabric related.  merging.

PR #605, opened 2017-07-06
Update default k8s version to 1.6.7

davidewatson:


davidewatson:
Fixes issues/#23

davidewatson:
CI was failing due to
> "ERROR: (gcloud.container.clusters.upgrade) ResponseError: code=400, message=bad desired node version ()."

Reverting change for GKE and opening a ticket to follow up...

PR #610, opened 2017-07-10
added user option for taints

DStorck:


joejulian:
I see that this is for aws only. Is this not something that could be provider agnostic? When we add more providers, it would be best if we didn't have to have provider-unique configs.

DStorck:
The work in theory is provider agnostic - I will make sure it is working as intended, but I do need to add in a gke-node schema and update the gke docs. Thanks

DStorck:
blocked by GKE bug, unable to test at the moment

coffeepac:
latest run for CI failed but the GKE stages passed.  removing blocked label.

schemas/config/v1/gkeNodeConfig.json
```diff
@@ -15,7 +15,14 @@
       "enum": [ "node" ],
       "type": "string"
     },
-    "providerConfig": { "$ref": "gkeNodeProviderConfig.json" }
+    "providerConfig": {
+      "$ref": "gkeNodeProviderConfig.json"
+    },
+    "taints": {
+      "description": "List of AWS taints to associate with the kubernetes node",
```
davidewatson: This is for GKE.

schemas/config/v1/kubeNodeTaint.json
```diff
@@ -0,0 +1,30 @@
...
+  "properties": {
+    "key": {
+      "description": "Taint key which must match toleration",
+      "type": "string"
+    },
+    "value": {
+      "description": "Taint value",
+      "type": "string"
+    },
+    "effect": {
```
DStorck: yeah i think enum is a good call

DStorck:
**Update** - Conclusion is that for the moment taints cannot be added at cluster creation.  Users wishing to add taints to GKE nodes can do so manually with `kubectl taint nodes <node> <key>=<value>:<effect>`. @joejulian

coffeepac:
We could have K2 add the taint as part of the cluster creation in a follow-on task.  We really need the taint to be added at cluster creation for AWS because we need to keep the control plane nodes clean

DStorck:
@coffeepac is the follow-on task you're referring to a way to get taints on to gke? Taints are currently being added at aws cluster creation for master nodes with https://github.com/samsung-cnct/k2/pull/588.

joejulian:
My thought is that it would be similar to, and before, kraken.services.

coffeepac:
@DStorck what joe said.  Another ansible task that calls kubectl to apply the taints.

DStorck:
@joejulian @coffeepac Sorry if I wasn't clear, I was asking if the follow-on task that Patrick mentioned was exclusive to gke since this is already handled for aws.

coffeepac:
@DStorck taints should be added as soon as possible, so for AWS it should be at process startup.  For GKE its as soon as possible.  this doesn't have to be the same for all providers, especially since GKE and AWS are so different.

coffeepac:
rebased to pickup some CI improvement.

PR #614, opened 2017-07-11
do not use $0 for call to mktemp

coffeepac:
$0 will contain the entire path being called which can inject
additional path segments into the file name.  this causes
problems.

the desire was to have the base name that mktemp creates be the
script name so its hardcoded now


coffeepac:
@mikeln passed CI, finally.

Issue #574, opened 2017-06-26
k2 etcd cluster names and tokens may not be unique


davidewatson:
Each etcd instance is started with a cluster token set within `ansible/roles/kraken.nodePool/kraken.etcd.docker/templates/etcd.units.etcd.part.jinja2 `:

```
     --discovery-srv {{cluster_node_tuple.1.name}}.{{cluster_node_tuple.0.name}}.internal \
...
     --initial-cluster-token {{cluster_node_tuple.1.etcdConfig.clusterToken}} \
```

Since the clusterToken is fixed, two k2 clusters with the same name could attempt to form an etcd cluster together. Note that this might not affect our CI, because, at least for the k2 PR job, the job number is included in the cluster name (e.g. `sed -i -e s/- name:[[:space:]]*$/- name: master-138/ cluster/aws/config.yaml`).

For more information on etcd discovery, see the [Clustering Guide](https://coreos.com/etcd/docs/latest/v2/clustering.html#dns-discovery) and the [Discovery Service Protocol](https://coreos.com/etcd/docs/latest/v2/discovery_protocol.html).

This is related to https://github.com/samsung-cnct/k2/issues/353 and https://github.com/samsung-cnct/k2/pull/351 (which introduced the problem). It may or may not be fixed by the solution to https://github.com/samsung-cnct/k2/issues/559.

joejulian:
Two k2 clusters will be in two independent networks. There should be no way for the two etcd clusters to discover each other.

coffeepac:
This will be a possible problem after we make it allowable to use an existing VPC.  That work is slated for the current sprint.

coffeepac:
blocked by a k2-tools issue RE: bunch of stuff happens twice in the dockerfile build and its weird.

coffeepac:
@davidewatson are you working on this?  if so, please assign it to yourself.  if not I'll chuck it back to ToDo

Issue #353, opened 2017-04-24
Create a uuid when generating config files


joejulian:
Create a uuid for a generated config file.

Check that uuid against the uuid stored in etcd or whatever we store it in to ensure the config file we're using matches the cluster we're attempting to manage.

Depends on #125

davidewatson:
Without this, our current use of the `ec2_remote_facts` module to generate the `ssh_config` can get confused if there is more than one cluster in the same region with the same name. The result will be an `ssh_config` with instances from multiple clusters.

PR #622, opened 2017-07-13
Split terraform actions into two phases, one for etcd, and one for everything else

davidewatson:


ansible/roles/kraken.provider/kraken.provider.aws/tasks/main.yaml
```diff
@@ -1,19 +1,39 @@
...
+- stat:
+    path: "{{ config_base | expanduser }}/{{ cluster.name }}/etcd.status.lock"
+  register: etcd_status_lock_file
+
+- name: Create etcd nodes
+  include: terraform-apply-etcd.yaml
+  when:
+    - not etcd_status_lock_file.stat.exists
+    - kraken_action == 'up'
+    - not (dryrun | bool)
```
coffeepac: can you add a 'terraform-plan-etcd.yaml' to ensure that the templated are being rendered out in a fully sane way? much like there is an aws-action and aws-action-dryrun

ansible/roles/kraken.provider/kraken.provider.aws/tasks/wait-for-etcd.yaml
```diff
@@ -0,0 +1,26 @@
...
+    ssh -o StrictHostKeyChecking=no
+        -o UserKnownHostsFile=/dev/null
+        -F {{ ssh_config }} {{ item.name }}-1
+        etcdctl --ca-file={{ etcd_cafile }}
+                --cert-file={{ etcd_certfile }}
+                --key-file={{ etcd_keyfile }}
+                -- endpoints={{ etcd_loopback_endpoints }}
+                   cluster-health
+  register: etcd_result
+  until: etcd_result | succeeded
```
coffeepac: will this catch a split brain cluster?  (might be out of scope, but asking)

ansible/roles/kraken.provider/kraken.provider.aws/tasks/wait-for-etcd.yaml
```diff
@@ -0,0 +1,26 @@
+---
+- set_fact:
+    ssh_config: "{{ config_base | expanduser }}/{{ cluster.name }}/ssh_config"
+    etcd_loopback_endpoints: "{% for port in item.etcdConfig.clientPorts %}{% if item.etcdConfig.ssl == true %}https{% else %}http{% endif %}://127.0.0.1:{{ port }}{% ',' if not loop.last %}{% endfor %}"
+    #etcd_loopback_endpoints: "{% set comma = joiner(",") %}{% for port in etcd.etcdConfig.clientPorts %}{% if etcd.etcdConfig.ssl == true %}https{% else %}http{% endif %}://127.0.0.1:{{ port }}{{ comma() }}{% endfor %}"
```
coffeepac: is this needed?

ansible/roles/kraken.provider/kraken.provider.aws/tasks/terraform-apply-etcd.yaml
```diff
@@ -0,0 +1,16 @@
+---
+- name: Get terraform modules
+  command: >
+    terraform get -no-color -update=true
+  args:
+    chdir: "{{ config_base | expanduser }}/{{ cluster.name }}"
+
+- name: Apply terraform configuration to current state of etcd
+  command: >
+    terraform apply -target module.etcd -input=false -no-color
```
coffeepac: to be clear, this `apply` will only target the etcd modules but the more general `apply` will target all modules including the etcd one, correct?  but hitting etcd twice doesn't matter, just some extra processing?

ansible/roles/kraken.provider/kraken.provider.aws/tasks/wait-for-etcd.yaml
davidewatson: It's in scope under https://github.com/samsung-cnct/k2/issues/574 If you join using a uuid (https://github.com/samsung-cnct/k2/issues/353), and start all subsequent new/recovered etcd instances correctly, then I don't think a split brain is possible.

We should probably add a debug statement to log etcd health before we proceed.

ansible/roles/kraken.provider/kraken.provider.aws/tasks/wait-for-etcd.yaml
davidewatson: No. I was just comparing the template with someone else's.

ansible/roles/kraken.provider/kraken.provider.aws/tasks/main.yaml
davidewatson: I was planning to make terraform plan always run and log the result. Is there any reason not to do this for the other terraform invocations? Even if `dryrun` is not set, I think we should capture the plan.

We can assume it is the same plan between runs only if the templating does not change.

coffeepac: I'm fine with running plan in both dryrun and not-dryrun.  that's a good idea.

ansible/roles/kraken.provider/kraken.provider.aws/tasks/wait-for-etcd.yaml
coffeepac: I was more referring to the occasional split brain we get when spinning up etcd with K2 right now.  I'm still not sure of the cause of this but I believe its a "bad time" involving SRV records and slow image downloads.

if we are moving to a less flaky etcd creating process then consider this ignorable.

ansible/roles/kraken.provider/kraken.provider.aws/tasks/wait-for-etcd.yaml
```diff
@@ -0,0 +1,26 @@
+---
+- set_fact:
+    ssh_config: "{{ config_base | expanduser }}/{{ cluster.name }}/ssh_config"
+    # !? Reformat this line using block strings
+    etcd_loopback_endpoints: "{% for port in item.etcdConfig.clientPorts %}{% if item.etcdConfig.ssl == true %}https{% else %}http{% endif %}://127.0.0.1:{{ port }}{% ',' if not loop.last %}{% endfor %}"
```
davidewatson: This line is incorrect. SInce we are checking over loopback, the list of endpoints should be one.

ansible/roles/kraken.provider/kraken.provider.aws/tasks/wait-for-etcd.yaml
```diff
@@ -0,0 +1,25 @@
+---
+- set_fact:
+    ssh_config: "{{ config_base | expanduser }}/{{ cluster.name }}/ssh_config"
+    etcd_loopback_endpoints: "{% for port in item.etcdConfig.clientPorts %}{% if item.etcdConfig.ssl == true %}https{% else %}http{% endif %}://127.0.0.1:{{ port }}{% ',' if not loop.last %}{% endfor %}"
```
davidewatson: This template is incorrect:

```
TASK [/workspace/ansible/roles/kraken.provider/kraken.provider.aws : set_fact] ********************************************************************************
fatal: [localhost]: FAILED! => {"failed": true, "msg": "template error while templating string: tag name expected. String: {% for port in item.etcdConfig.clientPorts %}{% if item.etcdConfig.ssl == true %}{% raw %}https{% endraw %}{% else %}{% raw %}http{% endraw %}{% endif %}{% raw %}://127.0.0.1:{% endraw %}{{ port }}{% ',' if not loop.last %}{% endfor %}"}
```

Issue #394, opened 2017-05-08
clean up k2 Jenkinsfile


coffeepac:
the current k2 Jenkinsfile uses an e2etester docker image that is far too large, isn't checking for correct fork (ie samsung_cnct vs coffeepac) when pushing a new image, has no concept of versioning and probably more.  

This issue should be used to tackle the topics in the previous paragraph (except versioning) and should be used to collect further desired improvements.  These improvements should then be converted into issues.  When the first three issues are cleaned up this issues should be closed and any further CI work should be in the normal flow of work.

coffeepac:
not going to be using jenkins going forward, closing.

PR #632, opened 2017-07-17
Simplify Jenkins stages

joejulian:


Jenkinsfile
```diff
@@ -11,96 +11,88 @@ podTemplate(label: 'k2', containers: [
...
-
-            parallel (
-                aws: {
-                    stage('aws config generation') {
-                        kubesh './up.sh --generate cluster/aws/config.yaml'
-                    }
+            // Unit tests
+            stage('Test: Unit tests') {
+                kubesh 'true' // Add unit test call here
+            }
```
coffeepac: ansible doesn't have a unit testing framework.  I don't want to write one.  Is there another option here?

Jenkinsfile
```diff
@@ -11,96 +11,88 @@ podTemplate(label: 'k2', containers: [
...
+                        stage('aws config generation') {
+                            kubesh './up.sh --generate cluster/aws/config.yaml'
+                        }
 
-                    try {
-                        stage('create k2 cluster') {
-                            kubesh 'PWD=`pwd` && ./up.sh --config $PWD/cluster/aws/config.yaml --output $PWD/cluster/aws/'
+                        stage('update generated aws config') {
+                            kubesh "build-scripts/update-generated-config.sh cluster/aws/config.yaml ${env.JOB_BASE_NAME}-${env.BUILD_ID}"
                         }
```
coffeepac: shouldn't these be in the 'Configure' stage before the parallel with the GKE stuff?

Jenkinsfile
```diff
@@ -11,96 +11,88 @@ podTemplate(label: 'k2', containers: [
...
-            stage('docker build') {
+            stage('Build') {
                 kubesh 'docker build -t quay.io/samsung_cnct/k2:latest docker/'
             }
 
             //only push from master.   assume we are on samsung-cnct fork
             //  ToDo:  check for correct fork
-            stage('docker push') {
-                if (env.BRANCH_NAME == "master") {
+            if (env.BRANCH_NAME == "master" && env.GIT_URL ==~ '/samsung_cnct/') {
```
coffeepac: nice!  this closes part of #394.

Jenkinsfile
joejulian: I have some ideas. Nothing that's far enough along to document as an issue yet.

Jenkinsfile
joejulian: Yep. Just wanted to get something running before I watched GoT.

joejulian:
Good enough for this time around.

Jenkinsfile
```diff
@@ -1,7 +1,16 @@
+// Configuration variables
+repo_org="samsung_cnct"
+cloud_test_timeout=8 // Should be about 5 min
+e2e_test_timeout=18  // Should be about 15 min
+cleanup_timeout=9    // Should be about 6 min
```
coffeepac: I don't really like having a timeout on cleanup.  I want cleanup to run for as long as it wants so long as its still trying to delete stuff.

Jenkinsfile
```diff
@@ -1,7 +1,16 @@
+// Configuration variables
+repo_org="samsung_cnct"
+cloud_test_timeout=8 // Should be about 5 min
```
coffeepac: this needs to be more like 50 minutes.  if we are spinning up the non-default GKE version then it normally takes 43-47 minutes to spin up the cluster.  its not a failure but it is a nuisance.  we dont' want to cause CI failures unless there is actually a failure.

Jenkinsfile
```diff
@@ -1,7 +1,16 @@
...
+e2e_test_timeout=18  // Should be about 15 min
```
coffeepac: if we term the e2e tests early, we are, I think, going to lose any information on what was taking so long.  so I'm not sure we want this timeout either.

Jenkinsfile
```diff
@@ -1,7 +1,16 @@
...
+cleanup_timeout=9    // Should be about 6 min
+kubernetes_version="v1.6.7"
+e2etester_version="0.2"
```
coffeepac: we are moving to 1.7 soon, should probably have a version for the custom-jnlp image as well.

Jenkinsfile
joejulian: Ugh. That's horrible. Ok though.

Jenkinsfile
joejulian: I'd be willing to make it longer, but I don't think we want anything to run forever.

Jenkinsfile
joejulian: Again... we'd had occasion where it ran forever. What I'd like is for a cleanup failure (including timeout) to notify slack so we would know that, 1) manual cleaning needs done, and 2) that something is seriously wrong with cleanup and needs fixed.

Jenkinsfile
coffeepac: I've never seen e2e run forever.  I have seen it run for 45 minutes.  It failed but still it ran.

Jenkinsfile
coffeepac: Right, some destroys have gone on forever.  For the slack Jenkins plugin can we have specific phases send custom messages (I would be okay with putting this on the slack notifier issue)

joejulian: Right, down ran forever before. Point is, at some point we can be sure it failed. Considering the flakiness of the E2E tests, I'd argue for it failing sooner rather than later so we can start it again.

joejulian: And I just responded to the wrong thread, but you get where I'm going.

coffeepac: do we want to let E2E finish its known failed run so we can get the test failure?  I do, sometimes, actually read through the logs.

joejulian: Even reading through the logs, I have to search carefully to find out why it failed. I just don't think that the E2E is stable enough to give it that much leash, yet.

coffeepac: if you click on the job number the published junit test makes the error(s) clear.

however, I won't argue to hard on this.  we can have the timeout for now but this bears looking into again in a month or so.

coffeepac:
Last job failed but job-1 succeeded.  Onward.

PR #661, opened 2017-07-24
helm: login to app registry

### LINKED ISSUE #657: Install private charts from app registry
The new helm chart registry supports private applications. Support them being installed via k2.

joejulian:
Add the ability to use private helm app registries

Fixes #657

### DIFF
--- a/schemas/config/v1/helmRegistry.json
+++ b/schemas/config/v1/helmRegistry.json
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
-  "id": "http://judkins.house/apis/k2/v1/helmRegistries.json",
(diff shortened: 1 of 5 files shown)


coffeepac:
Fixes #657

schemas/config/v1/helmRegistry.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "http://judkins.house/apis/k2/v1/helmRegistry.json",
+  "$$target": "helmRegistries.json",
```
coffeepac: I think this should be `helmRegistry.json` or the link will be broken when the docs get generated

joejulian: I think you're right.

Issue #607, opened 2017-07-07
Support k8s 1.7.x


davidewatson:
k8s v.1.7.0 ([CHANGELOG](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md#v170)) has been released. k2 needs to be updated to support 1.7.x and 1.4.x support should be removed. k2-tools will also need to be updated.

davidewatson:
RBAC changes will be needed for any resources using API version: `rbac.authorization.k8s.io/v1alpha1` In particular, the fabric daemonsets are affected.

@DStorck: New in 1.7, "The DaemonSet controller will evict running Pods that do not tolerate the NoSchedule taint if the taint is added to a Node." FYI

alejandroEsc:
Ill help with the fabrics side of this ticket if its ok with everyone?

coffeepac:
we need to do this nowish.  I will up the priority level and drag to ToDo after a discussion in Post.

davidewatson:
From slack: "helm doesn't work with 1.7 stateful sets"...

mikeln:
RBAC: system:node is deprated (which is used by kubelet).   https://kubernetes.io/docs/admin/admission-controllers/#noderestriction   is the recommendation.

coffeepac:
note: GKE supports 1.7 but the default is still 1.6.4 so leave that until this page: https://cloud.google.com/container-engine/release-notes updates (will be caught during weekly version check most likely)

samba:
A few findings in my work on this so far...

- [Helm support for 1.7 is expected in Helm's 2.6.0][1]
- In generating the `admin.kubeconfig`, we're adopting the username `admin` from the basic auth components, but the `{{ cluster.name }}_admin` is configured as the super user for RBAC. This prevents the cluster from spinning up because the user `admin` cannot query nodes, etc. *I have a fix for this already pending.*
- Currently getting a lot of `FailedSync` status from kubelet when attempting to deploy services (without helm) on 1.7... investigating.

Given our current reliance Helm, I'd guess that we shouldn't expect to merge support for 1.7 into K2 until Helm's 2.6.0 is available; correct? Thoughts?

[1]: https://github.com/kubernetes/helm/releases

mikeln:
Just remembered,  the DNS services are setup with specifc template directories.  unknown version default to the 1.4 setup.   Need to copy the v1.6 template dir to v1.7, and update the defaults/main.yaml for 1.7 versions.

[...]

Issue #667, opened 2017-07-25
authn change breaks existing configs


joejulian:
With an authn like:
```
        basic:
        - user: admin
          group: system:masters
          password: supersecret
```
and no "cert" section, up fails as kubectl does not have the permissions to `get nodes`.

mikeln:
that was well tested and did work.  
are you running 1.6?

joejulian:
Yes.

Renaming "basic" to "cert" fixed it for this config file.

mikeln:
basic should/did still work....checking now.

mikeln:
Just ran off an older branch (friday) and it worked fine with basic: only.
Still attempting to get master/latest to run.

mikeln:
ok, finally.   Ran master/latest  with this auth config worked fine.... :
```
 &rbacKubeAuth
      authz:
        rbac:
          # super_user is required until kubernetes 1.5 is no longer supported by k2.
          # It is not used by kubernetes 1.6 or later.
          super_user: "placeholder"
      authn:
        basic:
          -
            user: "admin"
            password: "bogusthing"
            group: "system:masters"
        default_basic_user: "admin"
```

mikeln:
Reran, with current, and previous, version of calicoNode to see if it had and CNI impacts.
```
calicoNode:
            #version: v1.2.1
            version: v1.1.3
            location: quay.io/calico/node
```
(we haven't moved our default up to 1.2.1 yet).
Didn't matter though.

mikeln:
ok, using the "real" config.   the group: system:masters wasn't quoted.
Yaml can have issues with non-quoted colons.
quoting seems to have fixed the issue.

joejulian:
closing as invalid

PR #669, opened 2017-07-26
Support k8s 1.7.x

samba:
Relates to samsung-cnct/k2#607

This is partially working. For me it correctly brings up a 1.7.1 cluster, and I can deploy stateless Nginx to it. Known issues described below. 

Feel free to suggests improvements if anything stands out.

### DIFF
--- a/ansible/roles/kraken.config/files/config.yaml
+++ b/ansible/roles/kraken.config/files/config.yaml
@@ -66,7 +66,7 @@ definitions:
             version: v1.8.3
             location: quay.io/calico/cni
           calicoNode:
-            version: v1.1.3
+            version: v1.2.1
             location: quay.io/calico/node
           flannel:
             version: v0.7.1
@@ -86,6 +86,7 @@ definitions:
         default: *defaultCanalFabric
         versions:
           v1.6: *defaultCanalFabric16
+          v1.7: *defaultCanalFabric16
     - &defaultWeaveFabric
       name: defaultWeaveFabric
       kind: fabric
@@ -124,7 +125,7 @@ definitions:
         default: *defaultWeaveFabric
         versions:
           v1.6: *defaultWeaveFabric16
-
+          v1.7: *defaultWeaveFabric16
   kvStoreConfigs:
     - &defaultEtcd
       name: etcd
@@ -159,7 +160,7 @@ definitions:
     - &defaultKube
       name: defaultKube
       kind: kubernetes
-      version: v1.6.7
+      version: v1.7.1
       hyperkubeLocation: gcr.io/google_containers/hyperkube
 
   containerConfigs:
--- a/ansible/roles/kraken.config/files/config.yaml
+++ b/ansible/roles/kraken.config/files/config.yaml
@@ -195,7 +195,7 @@ definitions:
       providerConfig:
         enablePublicIPs: true
         provider: aws
-        type: t2.small
+        type: t2.medium
         subnet: ["zone-1", "zone-2", "zone-3"]
         tags:
           -
@@ -235,7 +235,7 @@ definitions:
       providerConfig:
         enablePublicIPs: true
         provider: aws
-        type: t2.small
+        type: t2.medium
         subnet: ["zone-1", "zone-2", "zone-3"]
         tags:
           -
--- a/ansible/roles/kraken.fabric/kraken.fabric.flannel/templates/v1.7/canal.yaml.part.jinja2
+++ b/ansible/roles/kraken.fabric/kraken.fabric.flannel/templates/v1.7/canal.yaml.part.jinja2
@@ -1,0 +1,175 @@
+---
+#--------------------------
+# SOURCE OF THIS FILE: https://github.com/projectcalico/canal/blob/master/k8s-install/1.6/canal.yaml
+# 5/19/2017 mikeln - for samsung-cnct
+#--------------------------
+# This manifest installs the calico/node container, as well
+# as the Calico CNI plugins and network config on
+# each master and worker node in a Kubernetes cluster.
+#
+kind: DaemonSet
+apiVersion: extensions/v1beta1
+metadata:
+  name: canal
+  namespace: kube-networking
+  labels:
+    k8s-app: canal
+spec:
+  selector:
+    matchLabels:
+      k8s-app: canal
+  template:
+    metadata:
+      labels:
+        k8s-app: canal
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ''
+    spec:
+      hostNetwork: true
+      serviceAccountName: canal
+      tolerations:
+        # Allow the pod to run on the master.  This is required for
+        # the master to communicate with pods.
+        - key: node-role.kubernetes.io/master
+          effect: NoSchedule
+        # Mark the pod as a critical add-on for rescheduling.
+        - key: "CriticalAddonsOnly"
+          operator: "Exists"
+        - key: node-role.kubernetes.io/master
+          operator: Equal
+          effect: NoSchedule
+        - operator: Exists
+      containers:
+        # Runs calico/node container on each Kubernetes node.  This
+        # container programs network policy and routes on each
+        # host.
+        # image: quay.io/calico/node:v1.1.3
+        - name: calico-node
+          image: {{fabricConfigVal.options.containers.calicoNode.location}}:{{fabricConfigVal.options.containers.calicoNode.version}}
+          env:
+            # Use Kubernetes API as the backing datastore.
+            - name: DATASTORE_TYPE
+              value: "kubernetes"
+            # Enable felix logging.
+            - name: FELIX_LOGSEVERITYSYS
+              value: "info"
+            # Period, in seconds, at which felix re-applies all iptables state
+            - name: FELIX_IPTABLESREFRESHINTERVAL
+              value: "60"
+            # Disable IPV6 support in Felix.
+            - name: FELIX_IPV6SUPPORT
+              value: "false"
+            # Don't enable BGP.
+            - name: CALICO_NETWORKING_BACKEND
+              value: "none"
+            # Disable file logging so `kubectl logs` works.
+            - name: CALICO_DISABLE_FILE_LOGGING
+              value: "true"
+            - name: WAIT_FOR_DATASTORE
+              value: "true"
+            # No IP address needed.
+            - name: IP
+              value: ""
+            - name: HOSTNAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+            # Set Felix endpoint to host default action to ACCEPT.
+            - name: FELIX_DEFAULTENDPOINTTOHOSTACTION
+              value: "ACCEPT"
+          securityContext:
+            privileged: true
+          resources:
+            requests:
+              cpu: 250m
+          volumeMounts:
+            - mountPath: /lib/modules
+              name: lib-modules
+              readOnly: true
+            - mountPath: /var/run/calico
+              name: var-run-calico
+              readOnly: false
+        # This container installs the Calico CNI binaries
+        # and CNI network config file on each node.
+        # image: quay.io/calico/cni:v1.8.3
+        - name: install-cni
+          image: {{fabricConfigVal.options.containers.calicoCni.location}}:{{fabricConfigVal.options.containers.calicoCni.version}}
+          command: ["/install-cni.sh"]
+          env:
+            # The CNI network config to install on each node.
+            - name: CNI_NETWORK_CONFIG
+              valueFrom:
+                configMapKeyRef:
+                  name: canal-config
+                  key: cni_network_config
+            - name: KUBERNETES_NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+          volumeMounts:
+            - mountPath: /host/opt/cni/bin
+              name: cni-bin-dir
+            - mountPath: /host/etc/cni/net.d
+              name: cni-net-dir
+        # This container runs flannel using the kube-subnet-mgr backend
+        # for allocating subnets.
+        # image: quay.io/coreos/flannel:v0.7.1
+        - name: kube-flannel
+          image: {{fabricConfigVal.options.containers.flannel.location}}:{{fabricConfigVal.options.containers.flannel.version}}
+          command: [ "/opt/bin/flanneld", "--ip-masq", "--kube-subnet-mgr" ]
+          securityContext:
+            privileged: true
+          env:
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: FLANNELD_IFACE
+              valueFrom:
+                configMapKeyRef:
+                  name: canal-config
+                  key: canal_iface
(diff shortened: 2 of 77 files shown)


samba:
This revision does not currently:

- [ ] Deploy the "vanilla" Kubernetes dashboard
- [ ] Install Helm chart for either GitLab or MongoDB w/ stateful sets 

There appears to be some issue with the way storage classes (e.g. the default) are configured in the K2 cluster, so the Helm charts I've attempted will:

- Deploy a `PersistentVolumeClaim`, which gets stuck in `Pending` state, because:
- No persistent volumes can be provisioned/claimed/etc.

I've attempted all this *even after* deploying the below storage class, to no avail. 

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: standard
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
    volume.alpha.kubernetes.io/storage-class: "default"
provisioner: kubernetes.io/aws-ebs
parameters:
  type: gp2
```

ansible/roles/kraken.config/files/config.yaml
```diff
@@ -61,6 +61,12 @@ definitions:
       type: canal
       options:
         containers:
+          kubePolicyController:
+            version: v0.5.1
+            location: calico/kube-policy-controller
+          etcd:
+            version: v3.0.9
+            location: quay.io/coreos/etcd
```
mikeln: This is the wrong setup.  from 1.6 on we use the non-etcd (self contained) setup.  I thought this was documented in the 1.6 template file. (which is baed on this setup: https://github.com/projectcalico/canal/blob/master/k8s-install/1.6/canal.yaml).  Earlier version used this external etcd setup.

Do not add these lines.

ansible/roles/kraken.config/files/config.yaml
```diff
@@ -77,6 +83,34 @@ definitions:
...
+        containers:
+          kubePolicyController:
+            version: v0.6.0
+            location: calico/kube-policy-controller
+          etcd:
+            version: v3.0.9
+            location: quay.io/coreos/etcd
+          calicoCni:
+            version: v1.9.1
+            location: quay.io/calico/cni
```
mikeln: same comment here, as above.  this should be exactly the same as the current 1.6.

ansible/roles/kraken.fabric/kraken.fabric.flannel/templates/v1.7/canal.yaml.part.jinja2
```diff
@@ -0,0 +1,175 @@
...
+          configMap:
+            name: canal-config
+
+---
+
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: canal
+  namespace: kube-networking
```
mikeln: this is correct.  notice there is no policyController or etcd backing store.

ansible/roles/kraken.nodePool/kraken.nodePool.master/templates/v1.7/api-server-manifest.yaml.jinja2
```diff
@@ -0,0 +1,82 @@
...
+    - --etcd-certfile=/etc/etcd/ssl/client.pem
+    - --etcd-keyfile=/etc/etcd/ssl/client-key.pem
+    - --allow-privileged=true
+    - --service-cluster-ip-range={{ cluster.network }}
+    - --secure-port=443
+    - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,ResourceQuota
+    - --tls-cert-file=/etc/kubernetes/ssl/apiserver.pem
+    - --tls-private-key-file=/etc/kubernetes/ssl/apiserver-key.pem
+    - --client-ca-file=/etc/kubernetes/ssl/ca.pem
+    - --basic-auth-file=/etc/kubernetes/basicauth.csv
```
mikeln: You need to rebase your branch.  
There are recent changes that enable/disable basicauth that you are missing here.

ansible/roles/kraken.rbac/templates/rbac-default-policy.yaml.jinja2
```diff
@@ -121,6 +121,18 @@ roleRef:
...
+- kind: User
+  name: {{ cluster.name }}_admin
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
+  apiGroup: ""
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
```
mikeln: why is this needed?  This is redefining a built-in role.  
our default_basic_user:  has to be in the system:masters group.  There should not need to be a user of this type defined in the default rbac.   This comes back to my question, what is actually wrong with the default user being named `admin` :  `system:masters`?

ansible/roles/kraken.services/templates/rbac-dns-policy.yaml.jinja2
```diff
@@ -1,7 +1,7 @@
 ---
 # kubedns Roles
 kind: ClusterRole
-apiVersion: rbac.authorization.k8s.io/v1alpha1
+apiVersion: rbac.authorization.k8s.io/v1beta1
```
mikeln: this is correct only if pre v1.6 can't get to this file.   (which I think is true, as each release has it's own subdir).

ansible/roles/kraken.services/templates/v1.7/dns.yaml.jinja2
```diff
@@ -0,0 +1,199 @@
...
+          operator: "Exists"
+      volumes:
+        - name: kube-dns-config
+          configMap:
+            name: {{dns_config.k8sapp}}
+            optional: true
+      containers:
+        - name: kubedns
+          image: {{dns_config.kubedns_image_v16}}
+          resources:
```
mikeln: needs to be kubedns_image_v17

ansible/roles/kraken.services/templates/v1.7/dns.yaml.jinja2
```diff
@@ -0,0 +1,199 @@
...
+            protocol: TCP
+          - containerPort: 10055
+            name: metrics
+            protocol: TCP
+          volumeMounts:
+          - name: kube-dns-config
+            mountPath: /kube-dns-config
+        - name: dnsmasq
+          image: {{dns_config.dnsmasq_image_v16}}
+          livenessProbe:
```
mikeln: needs to be dnsmasq_image_v17

ansible/roles/kraken.services/templates/v1.7/dns.yaml.jinja2
```diff
@@ -0,0 +1,199 @@
...
+          resources:
+            requests:
+              cpu: {{dns_config.dnsmasq_cpu_req}}
+              memory: {{dns_config.dnsmasq_mem_req}}
+          volumeMounts:
+          - name: kube-dns-config
+            mountPath: /etc/k8s/dns/dnsmasq-nanny
+        - name: sidecar 
+          image: {{dns_config.dnsmasq_metrics_image_v16}} 
+          livenessProbe:
```
mikeln: needs to be dnsmasq_metrics_image_v17

ansible/roles/kraken.services/defaults/main.yaml
```diff
@@ -26,6 +27,7 @@ kubedns:
   kubedns_cpu_req: 100m
   kubedns_mem_req: 70Mi
   healthz_image: "gcr.io/google_containers/exechealthz-amd64:1.2"
+  healthz_image_v17: "gcr.io/google_containers/exechealthz-amd64:1.2"
```
mikeln: delete this line.  healthz is not uased as a separate image since 1.6+

ansible/roles/kraken.access/tasks/generate-admin-kubeconfig.yaml
```diff
@@ -26,7 +26,7 @@
 - name: create admin user entry from basicauth user
   command: >
     {{ kubectl }} config set-credentials {{ cluster.name }}_admin
-      --username={{ default_basic_kube_auth.user }}
+      --username={% if cluster.kubeAuth.authz.rbac %}{{ cluster.name }}_admin{% else %}{{ default_basic_kube_auth.user }}{% endif %}
```
samba: With 1.7, drawing from the 1.6 configuration previous, RBAC is set up and Ansible fails even to `get nodes` during its verification phase, so it quits. 

To get past that, I discovered (by manually altering my own kubeconfig) that setting the username to the generated one  `{{ cluster.name }}_admin` would allow me to authenticate and continue the deployment. This approach has worked reproducibly ever since.

If there's an alternate approach you'd suggest, I'm open to it, but this is the only way I found that worked... so far.

ansible/roles/kraken.config/files/config.yaml
samba: Hmm, somehow these were required; it kept failing on missing the key `etcd` during deployment. I'll take another look at that and see if I can find a way to proceed without them.

ansible/roles/kraken.services/defaults/main.yaml
```diff
@@ -15,6 +15,7 @@ kubedns:
   version: v17
   color: FFFF66
   kubedns_image: "gcr.io/google_containers/kubedns-amd64:1.8"
+  kubedns_image_v17: "gcr.io/google_containers/k8s-dns-kube-dns-amd64:1.14.4"
   kubedns_image_v16: "gcr.io/google_containers/k8s-dns-kube-dns-amd64:1.14.2"
   kubedns_image_v15: "gcr.io/google_containers/kubedns-amd64:1.9"
   kubedns_image_v14: "gcr.io/google_containers/kubedns-amd64:1.8"
```
coffeepac: 1.4 support is dropped with adding 1.7

ansible/roles/kraken.services/defaults/main.yaml
```diff
@@ -26,6 +27,7 @@ kubedns:
   kubedns_cpu_req: 100m
   kubedns_mem_req: 70Mi
   healthz_image: "gcr.io/google_containers/exechealthz-amd64:1.2"
+  healthz_image_v17: "gcr.io/google_containers/exechealthz-amd64:1.2"
   healthz_image_v15: "gcr.io/google_containers/exechealthz-amd64:1.2"
   healthz_image_v14: "gcr.io/google_containers/exechealthz-amd64:1.2"
```
coffeepac: same, will not repeat comment again but all 1.4 should be removed.

ansible/roles/kraken.nodePool/kraken.nodePool.node/templates/v1.7/units.kubelet.part.jinja2
```diff
@@ -0,0 +1,62 @@
+- name: kubelet.service
```
coffeepac: is this the same as the 1.6 file?   I think we talked about this briefly in person.  if so, maybe we move the 1.5 file (currently default) into a versioned directory and let this be the default.  

this applies to the networking configs as well.

coffeepac:
a.  thanks!  
b. if you have the time, can you make the default version of the files be the common file?  ie:  if the common file is for 1.6 & 1.7, please make that the default.  this will require checking that 1.5 still works so it is some scope creep.  if you don't have time for this, please file an issue so someone else can perform this cleanup and link it here.
c. please set the default version of kubernetes to be 1.7.1 so that CI can test it.  this PR is using 1.6.7 in CI right now.  you may also need to push a temporary k2-tools image (and use it in the Jenkinsfile) while samsung-cnct/k2-tools#97 keeps moving

ansible/roles/kraken.config/files/config.yaml
mikeln: That would indicate that your setup is reading the wrong template dir...  (i.e. v1.7 subdir is missing, in which case the default (1.4) is being read).

ansible/roles/kraken.access/tasks/generate-admin-kubeconfig.yaml
mikeln: We really need to understand why for this one.  The behavior is odd.

ansible/roles/kraken.config/files/config.yaml
samba: Thanks for the tip. I'm sure that was true when I made this change as I hadn't yet created the 1.7 directories. Will reevaluate.

mikeln:
Yeah, I have to second what Patrick said.  I think we need to adopt a "default is current" for anyplace with template subdirs, and then have version subdirs for past versions.
```
/templates
    whatever.jinja  <should be the v1.7 version>
    /v1.6
       whateverA.jinja 
    /v1.5
       whateverB.jinja
```
Also, from my initial look,  v1.7 should be created from v1.6 files.  (and for that matter, in almost all places it may be possible to make v1.6 files default, and remove the v1.6 directory.  In this way 1.6 and 1.7 will both use the same default file.  This would be perfectly ok, until some definite difference was found btwn 1.6 and 1.7).  

anything with 1.4 should be removed.  templates, variables, etc.

In kraken.config/files/config.yaml and gke-config.yaml.   We should be able to initially create a `<blah>17 `config at the `16` block. e.g.
```
 - &kubeVersionedFabric
      name: kubeVersionedFabric
      kind: versionedFabric
      type: canal
      kubeVersion:
        default: *defaultCanalFabric
        versions:
          v1.6: *defaultCanalFabric16
          v1.7: *defaultCanalFabric16
```
instead of creating a whole new block.  The only changes would be a version change, and those should apply to 16 also anyway.  The only time you would need a new 17 block is if the setup/vars are much different.

ansible/roles/kraken.access/tasks/generate-admin-kubeconfig.yaml
joejulian: Perhaps this is the same as #667

mikeln: No.  I just threw together a 1.7 setup (completely based on 1.6).   It came up fine.  So these change lines are not needed.

mikeln:
Actually, baring any changes to the directory structure to make `/template` defaults be 1.7, just duplicating the v1.6 directory to a new v1.7 directory is a good first step.   

You should not have to update DNS internal vars to _17. I think just running the 1.6 setup will work fine for now.

coffeepac:
CI is failing because k2-tools doesn't have a 1.7 compatible version of helm installed.  I recommend updating build-scripts/update-generated-config.sh to set the config flag to use the latest version of helm that is installed.  Can probably tack it onto the `name` setting regex by doing a newline, spaces and setting  "helmOverride: true"  or something.  

I may have made this same comment on the k2-tools PR.

samba:
@mikeln @coffeepac 
Based on comments here & discussion in Slack, I've outlined the process in `Documentation/UPGRADING_KUBERNETES.md`, for:

- Migrating (past) default templates to a version-specific template directory.
- Updating main cluster configuration to suit. 

I've also demonstrated it with the `v1.5` configuration templates.

Please review the outline and let me know if it sounds sensible. 

I'll follow up on the CI related tweaks to let this pass (e.g. helm version issues).

samba:
Also, for what it's worth:

1. Previous attempt failed because I was basically forking `1.5` configuration and trying to hack it into a viable `1.7` cluster, without the benefit of all the `1.6` work done previously. (my bad)
2. With @mikeln's rendition, I've got functional `1.7.1` and `1.5.7` clusters running currently.

samba:
(TO DO for @samba: verify that defaults match 1.7)

Issue #671, opened 2017-07-26
[Etcd]: k2 expecting to get cluster health from a private topology via ssh


alejandroEsc:
Currently we are seeing that if private ips are set than we cannot connect to the machines and so we cannot get a cluster health assessment. We need to unblock ourselves here and leave the current proposal to not run the check if the network is private.

PR #672, opened 2017-07-26
k2 expecting to get cluster health from a private topology via ssh

alejandroEsc:
- Added a check to run task: `Wait for etcd cluster to form` only when public ips are available.

should resolve: https://github.com/samsung-cnct/k2/issues/671

### DIFF
--- a/ansible/roles/kraken.provider/kraken.provider.aws/tasks/wait-for-etcd.yaml
+++ b/ansible/roles/kraken.provider/kraken.provider.aws/tasks/wait-for-etcd.yaml
@@ -19,6 +19,8 @@
   until: etcd_result | succeeded
   retries: "{{ etcd_retries }}"
   delay: "{{ etcd_delay }}"
+  when: item.nodeConfig.providerConfig.enablePublicIPs is undefined or
+  ( item.nodeConfig.providerConfig.enablePublicIPs is defined and item.nodeConfig.providerConfig.enablePublicIPs )
 
 - debug:
     var: etcd_result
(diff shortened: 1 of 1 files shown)


ansible/roles/kraken.provider/kraken.provider.aws/tasks/terraform-apply-etcd.yaml
```diff
@@ -26,10 +26,6 @@
     chdir: "{{ config_base | expanduser }}/{{ cluster.name }}"
   register: tf_etcd_plan_result
 
-- debug:
-    var: tf_etcd_plan_result
-    verbosity: 0
```
coffeepac: this should remain.  its noisy but we want to know what, if any, changes to the etcd nodes is going to happen

alejandroEsc:
I dont think this belongs in master, are the debug tools in a dev env. not
enough?

coffeepac:
This is not for dev purposes.  It is for operational/tracking purposes.  Etcd should never change after initial creation so we want to know if it is going to.

alejandroEsc:
@coffeepac dont mean to debate the point, I thought that a bug reporter was handling this very thing... anyway ill revert the change. Is it possible to get an image out available tonight? i believe it has a direct impact on comcast.

coffeepac:
there is a crash reporter, but I don't think its advanced enough for this kind of thing.  

and yes, I'll sit on this tonight.

alejandroEsc:
trying to figure out the issue... but testing locally I am seeing issues with etcd and raft... i think the issue that @mikeln is seeing where each node is trying to establish concensus but they dont get what they expect to see
```
Jul 27 04:13:00 ip-10-0-95-123.us-west-2.compute.internal rkt[1035]: 2017-07-27 04:13:00.878946 E | rafthttp: request cluster ID mismatch (got ba2c168f4d9a5089 want f18f883e11104e3)
Jul 27 04:13:00 ip-10-0-95-123.us-west-2.compute.internal rkt[1035]: 2017-07-27 04:13:00.879321 E | rafthttp: request cluster ID mismatch (got ba2c168f4d9a5089 want f18f883e11104e3)
```

Issue #676, opened 2017-07-27
ec2_elb_facts can cause API rate limit issues


coffeepac:
there is no way to throttle the ansible module nor does the ansible module provide a way to back off/retry if an API limit is hit.  we should either:
a. make an update to the module so it does do back off/retry
b. re-write this task to not use this module. 

using this module is a bit problematic for another reason:  #563

alejandroEsc:
we should watch this: https://github.com/ansible/ansible/issues/13796

sostheim:
@coffeepac - just checking, did you want this ticket to remain open to track the long term resolution to #677 I believe the "fix:" text in the description caused this to automatically close on merge.

coffeepac:
I do not want this to remain open.  We have another issue open to stop using this module entirely.

Issue #644, opened 2017-07-18
./down.sh + BYOvpc causes a API threshold hit every time


coffeepac:
this is the correct behavior if K2 creates the VPC and all resources inside it, however, that is an incorrect assumption when BYO vpc is going on.

coffeepac:
initial analysis was incorrect.  the list of ELBs to delete comes from kubernetes and not AWS ELB API.  Going to finish this as I *think* it'll fix #635.

coffeepac:
the solution I'm cooking up *might* leak ELBs but we'll address that if it happens.  Should be fine.

Issue #635, opened 2017-07-17
AWS API Threshold Limit Violation Exceeded On Every Run of `down.sh`


sostheim:
As a Cluster Operator on AWS Cloud Provider infrastructure, every time I try to run `./down.sh` on my K2 managed Kubernetes cluster, I experience a API threshold limit violation when the script begins the teardown of AWS ELB.

This may be a duplicate of, or artifact of, #268

sostheim:
Currently, only resolution is to tear down cluster components from AWS CLI, or Console, by hand.

coffeepac:
downgrading from p0 to p1.  this is annoying but it does not cause anyone to wake up at 2am to resolve.

coffeepac:
@sostheim can you please get the output from a down run that explodes?  We don't see this anywhere else on the regular like that.

joejulian:
Asked for logs privately

sostheim:
Will repro and capture logs.  Any specific advice on how best to capture a useful log?

👍 on the p1 status.

coffeepac:
`./down.sh <whatever flags get used> 2>&1 > down-log.<date>.log`  should do it

sostheim:
[down-log.7-18-17_standard_output.txt](https://github.com/samsung-cnct/k2/files/1157169/down-log.7-18-17_standard_output.txt)

Log file attached from attempting to bring down a cluster that exhibits this error.

I have a verbose log as well that I'll attach as soon as I'm done scrubbing to ensure that there isn't any sensitive information in the log.

sostheim:
Verbose output log, same error:
[down-log.7-18-17_verbose_output.txt](https://github.com/samsung-cnct/k2/files/1157200/down-log.7-18-17_verbose_output.txt)

sostheim:
Courtesy of research by @twilfong, potentially related issues from Ansible:
https://github.com/ansible/ansible/issues/13796
https://github.com/ansible/ansible-modules-core/issues/143

coffeepac:
just realized that k2 down is going to try and delete every ELB in the VPC as a cleanup mechanism.  this is a throwback to K2 assuming everything in the VPC must go.

that seems like a thing that needs to be fixed sooner than later.

alejandroEsc:
[...]

PR #677, opened 2017-07-27
if in a BYO vpc scenario, do not block for all ELBs

coffeepac:
it is probably/guarnateed that there will be ELBs in the VPC that
do not belong to kubernetes.

This doesn't resolve the underlying issue that the ansible module
can easily overwhelm the API threshold but it does unblock people
who certainly don't want to block on this.

long term fix:  #676 

fixes: #635 
fixes: #644

### DIFF
--- a/ansible/roles/kraken.services/tasks/kill-services.yaml
+++ b/ansible/roles/kraken.services/tasks/kill-services.yaml
@@ -127,7 +127,7 @@
   register: elb_facts
   vars:
     vpc_lookup: "elbs[?vpc_id=='{{ vpcid }}']"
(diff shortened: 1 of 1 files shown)


alejandroEsc:
lgtm, wow hope this resolves the api limit, how is it a elb thing, got some ideas?

sostheim:
Does this solution have the potential to leak ELBs as noted in #644?  Should we have, as a documented process, a check for left behind cluster ELBs?

coffeepac:
@sostheim I do not believe so.  I did not fully understand the design of this when I made that comment.

PR #679, opened 2017-07-27
pull aws_session_token from credentials profile when available to all…

MrBouncy:
…ow the use of STS temporary creds

### DIFF
--- a/ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_config.yaml
+++ b/ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_config.yaml
@@ -23,5 +23,6 @@
   set_fact:
     cluster: "{{ cluster | combine({'providerConfig':{'authentication':{'securityToken': lookup('ini', 'aws_session_token section=' + cluster.providerConfig.authentication.credentialsProfile + ' file=' + cluster.providerConfig.authentication.credentialsFile)}}}, recursive=True)}}"
   when:
+    - cluster.providerConfig.authentication.securityToken is defined
     - cluster.providerConfig.authentication.securityToken | is_empty
     - not (cluster.providerConfig.authentication.credentialsFile | is_empty)
--- a/ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_config.yaml
+++ b/ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_config.yaml
@@ -9,16 +9,19 @@
   set_fact: 
     cluster: "{{ cluster |combine({'providerConfig':{'authentication':{'accessKey': lookup('ini', 'aws_access_key_id section=' + cluster.providerConfig.authentication.credentialsProfile + ' file=' + cluster.providerConfig.authentication.credentialsFile)}}}, recursive=True)}}"
   when:
-    cluster.providerConfig.authentication.accessKey | is_empty and cluster.providerConfig.authentication.credentialsFile | is_empty == False
+    - cluster.providerConfig.authentication.accessKey | is_empty
+    - not (cluster.providerConfig.authentication.credentialsFile | is_empty)
 
 - name: set accessSecret if unset and credentialsFile set
   set_fact: 
     cluster: "{{ cluster | combine({'providerConfig':{'authentication':{'accessSecret': lookup('ini', 'aws_secret_access_key section=' + cluster.providerConfig.authentication.credentialsProfile + ' file=' + cluster.providerConfig.authentication.credentialsFile)}}}, recursive=True)}}"
   when:
-    cluster.providerConfig.authentication.accessSecret | is_empty and cluster.providerConfig.authentication.credentialsFile | is_empty == False
+    - cluster.providerConfig.authentication.accessSecret | is_empty
+    - not (cluster.providerConfig.authentication.credentialsFile | is_empty)
 
 - name: set securityToken if unset and credentialsFile set
   set_fact:
     cluster: "{{ cluster | combine({'providerConfig':{'authentication':{'securityToken': lookup('ini', 'aws_session_token section=' + cluster.providerConfig.authentication.credentialsProfile + ' file=' + cluster.providerConfig.authentication.credentialsFile)}}}, recursive=True)}}"
   when:
-    cluster.providerConfig.authentication.securityToken | is_empty and cluster.providerConfig.authentication.credentialsFile | is_empty == False
+    - cluster.providerConfig.authentication.securityToken | is_empty
+    - not (cluster.providerConfig.authentication.credentialsFile | is_empty)
--- a/ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_config.yaml
+++ b/ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_config.yaml
@@ -16,3 +16,9 @@
     cluster: "{{ cluster | combine({'providerConfig':{'authentication':{'accessSecret': lookup('ini', 'aws_secret_access_key section=' + cluster.providerConfig.authentication.credentialsProfile + ' file=' + cluster.providerConfig.authentication.credentialsFile)}}}, recursive=True)}}"
   when:
     cluster.providerConfig.authentication.accessSecret | is_empty and cluster.providerConfig.authentication.credentialsFile | is_empty == False
+
+- name: set securityToken if unset and credentialsFile set
+  set_fact:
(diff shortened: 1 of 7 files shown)


ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/main.yaml
```diff
@@ -12,6 +12,7 @@
     region: "{{ cluster.providerConfig.region }}"
     aws_access_key: "{{ cluster.providerConfig.authentication.accessKey or omit }}"
     aws_secret_key: "{{ cluster.providerConfig.authentication.accessSecret or omit }}"
+    security_token: "{{ cluster.providerConfig.authentication.securityToken or omit }}"
```
coffeepac: there are four locations that will require this change:
- ansible/roles/kraken.config/tasks/aws_config.yaml
- ansible/roles/kraken.services/tasks/aws_config.yaml
- ~ansible/roles/kraken.ssh/tasks/aws_config.yaml~ completed
- ansible/roles/kraken.update/tasks/aws_config.yaml

Please make this update to the other three locations

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_config.yaml
```diff
@@ -15,4 +15,10 @@
...
   when:
-    cluster.providerConfig.authentication.accessSecret | is_empty and cluster.providerConfig.authentication.credentialsFile | is_empty == False
\ No newline at end of file
+    cluster.providerConfig.authentication.accessSecret | is_empty and cluster.providerConfig.authentication.credentialsFile | is_empty == False
+
+- name: set securityToken if unset and credentialsFile set
+  set_fact:
+    cluster: "{{ cluster | combine({'providerConfig':{'authentication':{'securityToken': lookup('ini', 'aws_session_token section=' + cluster.providerConfig.authentication.credentialsProfile + ' file=' + cluster.providerConfig.authentication.credentialsFile)}}}, recursive=True)}}"
+  when:
+    cluster.providerConfig.authentication.securityToken | is_empty and cluster.providerConfig.authentication.credentialsFile | is_empty == False
```
coffeepac: securityToken should be added to the schema config.  file is here:
schemas/config/v1/awsAuthenticationConfig.json

This is not a required field, so do not add to the `required` list

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_config.yaml
```diff
@@ -15,4 +15,10 @@
   set_fact: 
     cluster: "{{ cluster | combine({'providerConfig':{'authentication':{'accessSecret': lookup('ini', 'aws_secret_access_key section=' + cluster.providerConfig.authentication.credentialsProfile + ' file=' + cluster.providerConfig.authentication.credentialsFile)}}}, recursive=True)}}"
   when:
-    cluster.providerConfig.authentication.accessSecret | is_empty and cluster.providerConfig.authentication.credentialsFile | is_empty == False
\ No newline at end of file
+    cluster.providerConfig.authentication.accessSecret | is_empty and cluster.providerConfig.authentication.credentialsFile | is_empty == False
+
+- name: set securityToken if unset and credentialsFile set
+  set_fact:
+    cluster: "{{ cluster | combine({'providerConfig':{'authentication':{'securityToken': lookup('ini', 'aws_session_token section=' + cluster.providerConfig.authentication.credentialsProfile + ' file=' + cluster.providerConfig.authentication.credentialsFile)}}}, recursive=True)}}"
```
coffeepac: what happens when you run this and your credentials file doesn't have an aws_session_token field defined?  ie does this generate a strange warning?

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_config.yaml
coffeepac: what happens when you do not have securityToken defined in your config file at all?  it needs to be an optional parameter for backwards compatibility so we can't require it for now.

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_config.yaml
joejulian: default is an empty string (unless specifically specified) so that should be ok as long as an empty string doesn't change the behavior elsewhere.

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_config.yaml
joejulian: As @coffeepac was saying, you'll need to test for `defined` before you can test `is_empty`.

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_config.yaml
```diff
@@ -15,4 +15,10 @@
   set_fact: 
     cluster: "{{ cluster | combine({'providerConfig':{'authentication':{'accessSecret': lookup('ini', 'aws_secret_access_key section=' + cluster.providerConfig.authentication.credentialsProfile + ' file=' + cluster.providerConfig.authentication.credentialsFile)}}}, recursive=True)}}"
   when:
-    cluster.providerConfig.authentication.accessSecret | is_empty and cluster.providerConfig.authentication.credentialsFile | is_empty == False
\ No newline at end of file
+    cluster.providerConfig.authentication.accessSecret | is_empty and cluster.providerConfig.authentication.credentialsFile | is_empty == False
```
joejulian: A couple style preferences that might be worth doing as part of this.
Multiple tests combined with `and` should be split into a list and `not` is preferred to `== False`, ie.

```
  when:
    - cluster.providerConfig.authentication.accessSecret | is_empty
    - not (cluster.providerConfig.authentication.credentialsFile | is_empty)
```

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_config.yaml
joejulian: Same style request as above

MrBouncy: sorry for the delay in circling back on this, I think the asks are all totally reasonable and I will work through them just haven't had the cycles to focus on this yet..

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/main.yaml
MrBouncy: the kraken.config path no longer includes these calls, but I added the initial cut of security token handling to the services and update role paths (still pending a pass for style changes)

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_config.yaml
MrBouncy: the session token is only required with STS temporary creds, I will do a test pass with static keys which won't have the session token defined (or required to let ansible finish its fact gathering) and verify that all behaves nicely and followup here on the results

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_config.yaml
MrBouncy: added to the schema definition but not in the required list

MrBouncy: confirmed that after prepending defined tests there are no errors without the empty securityToken key in my cluster config

coffeepac:
@MrBouncy thanks again for this PR!  We're having some CI problems with GKE being out of resources so I can't merge it yet.  If GKE hasn't sorted itself by tomorrow I'll start on a workaround.

coffeepac:
@MrBouncy my apologies, but do you mind rebasing against master?  GKE isn't going to be functional for us again for CI for a few days.  I have put a patch in to skip the GKE paths.

coffeepac:
failing CI for an actual reason.  I haven't been able to repro it yet (only tried once, need to run) but the error seems to be something is expecting securityToken to be exist.  you should be able to reproduce this yourself by running:

`bin/up.sh --config <path to config> -t dryrun`

If that works for you, I suggest generating a new config file and trying `dryrun` again.  I'll be able to look at this again tomorrow either way.

MrBouncy:
Found the error with a fresh config, and found the issue with my testing approach,
should have a fix ready shortly I hope..

joejulian:
Thanks for the contribution!

Issue #659, opened 2017-07-24
Create an optional bastion instance.


alejandroEsc:
to support private topologies, we should create a public facing bastion object that:

- has ssh authorized public keys for the user (initially 1)
- No running kubenertes services, and not managed by k8s
- launched in defined subnets 
- optional entry in the cluster config.
- should be  of instance type t2.micro? but we should agree on this
- should be managed by an asg
- OS should be coreos

any additional requirements to be discussed on this ticket.

coffeepac:
t2.micro seems fine.

I am unsure why it should be coreOS, we plan to support other OSes in the near to medium term.
why should the bastion be in an ASG?

alejandroEsc:
- coreos because that is what we are using, doesnt have to be you are right, but figure we do that since we already use it. Let me know if you are considering something else. ASG is because a bastion should always work if that is the only way we can get access the nodes directly, one per AZ  would be ideal. you dont need an asg, but i think it would be nice to make the bastion at least always reliably there, if costs are incurred due to it, i suppose just having to run the up/terraform apply should do i guess.

coffeepac:
OS:  okay.  it doesn't matter much, but was curious if there was a hard reason for it.  the bastion node should have an osConfig field which can be whatever.

ASG:  one per AZ?  ahh, for continued access in the face of AZ failure.  Okay, now I get it.  Sounds good.

coffeepac:
@alejandroEsc are you doing this?

alejandroEsc:
working on it since yesterday afternoon.

PR #682, opened 2017-07-28
Add Bastion Nodepool object

alejandroEsc:
First pass PR for adding nodepool object: Bastion

- DO NOT MERGE

- [x] Clean up
- [X] Confirm with @coffeepac that nodepool task and changes are ok.
- [x] Add working config example of private network with Bastion used. (BYO VPC only atm)

fixes: https://github.com/samsung-cnct/k2/issues/659


ansible/roles/kraken.config/tasks/main.yaml
```diff
@@ -94,17 +94,36 @@
     loop_var: a_cluster
   when: kraken_action =='up'
 
-- name: Ensure each nodePool corresponds to exactly one of the node roles (i.e. etcd, master, or node)
+# we have these in two parts since it would make it easier to maintain and scale over time and easy to change.
+- name: Ensure each nodePool item corresponds at least one of the node roles - etcd, master, node, or bastion
   fail:
     msg: "({{ i.0.name }}, {{ i.1.name }}) corresponds to more than one node role"
```
coffeepac: this error message is wrong.  should be ' ... corresponds to zero fully defined roles'

ansible/roles/kraken.config/tasks/main.yaml
```diff
@@ -94,17 +94,36 @@
...
     - "{{ kraken_config.clusters }}"
     - nodePools
   loop_control:
     loop_var: i
-  when: (i.1.etcdConfig is defined and (i.1.apiServerConfig is defined or i.1.kubeConfig is defined))
-         or (i.1.apiServerConfig is defined and i.1.kubeConfig is undefined)
-         or (i.1.etcdConfig is undefined and i.1.kubeConfig is undefined)
+  when:
+    - i.1.etcdConfig is undefined
+    - i.1.apiServerConfig is undefined
```
coffeepac: this should be `i.1.apiServerConfig is undefined OR (i.1.apiServerConfig is defined && i.1.kubeConfig is undefined)` to be logically consistent with the other entries, ie:  each entry in this list corresponds to one complete role definition.  its a bit awkward as the definition for a valid apiServer is a compound definition

ansible/roles/kraken.config/tasks/main.yaml
```diff
@@ -94,17 +94,36 @@
...
+  with_subelements:
+    - "{{ kraken_config.clusters }}"
+    - nodePools
+  loop_control:
+    loop_var: i
+  when:
+    (i.1.etcdConfig is defined and i.1.apiServerConfig is defined) or
+    (i.1.etcdConfig is defined and i.1.kubeConfig is defined) or
+    (i.1.etcdConfig is defined and i.1.bastionConfig is defined) or
+    (i.1.apiServerConfig is defined and i.1.kubeConfig is undefined) or
```
coffeepac: removed, should be part of the prior 'at least one valid role' check

ansible/roles/kraken.config/tasks/main.yaml
```diff
@@ -94,17 +94,36 @@
...
+    msg: "({{ i.0.name }}, {{ i.1.name }}) corresponds to more than one node role"
+  with_subelements:
+    - "{{ kraken_config.clusters }}"
+    - nodePools
+  loop_control:
+    loop_var: i
+  when:
+    (i.1.etcdConfig is defined and i.1.apiServerConfig is defined) or
+    (i.1.etcdConfig is defined and i.1.kubeConfig is defined) or
+    (i.1.etcdConfig is defined and i.1.bastionConfig is defined) or
```
coffeepac: these three lines should be rolled into one line:  `(i.1.etcConfig is defined and (i.1.kubeConfig is defined or i.1.bastion is defined))`  (don't need to check for apiServer as there is no valid config that has apiServer but not kubeConfig defined

schemas/config/v1/nodePool.json
```diff
@@ -15,8 +15,10 @@
       "type": "number"
     },
     "kvStoreConfig": { "$ref": "kvStoreConfig.json" },
+    "etcdConfig": { "$ref": "etcdConfig.json" },
```
coffeepac: either kvStoreConfig shouldn't be present or etcdConfig shouldn't be present as etcdConfig is a kvStoreConfig.  This feels like the validation script needs tightening up.

unless required by something for the PR, please remove the etcdConfig line.

ansible/roles/kraken.nodePool/kraken.nodePool.selector/tasks/main.yaml
```diff
@@ -14,7 +14,7 @@
   include_role:
     name: '{{ playbook_dir }}/roles/kraken.nodePool/kraken.nodePool.node'
 
-- name: Error Provider
+- name: Supported docker runtime check.
```
davidewatson: Is this really a check for a supported docker runtime? I think the task name should be `- name: Check for a supported node type (e.g. etcd, master, node, or bastion)`.

schemas/config/v1/definitions.json
```diff
@@ -51,6 +51,15 @@
       "minItems": 1,
       "type": "array"
     },
+    "bastionConfigs": {
+      "title": "Bastion configurations",
+      "description": "A list of bastion configurations within the configuration file.",
+      "items": {
+        "$ref": "bastionConfig.json"
+      },
+      "minItems": 0,
```
davidewatson: For consistency with the rest of the definition lists, should this be `minItems: 1`? Unused definitions should exist as a convenience for the user.

schemas/config/v1/bastionConfig.json
```diff
@@ -0,0 +1,26 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "http://judkins.house/apis/k2/v1/bastionConfig.json",
+  "$$target": "bastionConfig.json",
+  "title": "A Bastion configuration",
+  "description": "Properties pertaining to a Bastion.",
+
+  "properties": {
+    "name": {
+      "description": "Name of the Kubernetes configuration.",
```
davidewatson: Was this description intended to be `"description": "Name of the Bastion configuration.",`?

schemas/config/v1/bastionConfig.json
```diff
@@ -0,0 +1,26 @@
...
+      "type": "string"
+    },
+    "kind": {
```
davidewatson: `kind` is intended to differentiate between different `bastionConfig` formats. For a given format, I think you want to define a `type` property and give make it an enum with only one possible value, say `default`.

schemas/config/v1/definitions.json
alejandroEsc: would we expect the schema check error if I do not list an item in that case?

schemas/config/v1/nodePool.json
alejandroEsc: ohh for some reason i red kubeconfig, yes. I agree.

schemas/config/v1/bastionConfig.json
alejandroEsc: I think i want to leave this as kind, since I am thinking of (as Patrick mentions above) an option for LBed bastion and another  for a bastion with an ssh_config file.

ansible/roles/kraken.config/tasks/main.yaml
alejandroEsc: So i actually wrote the truth table to the cases for the previous code that was here. It did as you say @davidewatson selected the values we wanted to fail on (note bastion case is added here also). It was interesting but difficult to figure out what it was doing without writing the whole table out. To that end this case fails in the one hard to get right case where nothing is defined, but yeah @coffeepac that would take care of both of these cases at once.

schemas/config/v1/bastionConfig.json
alejandroEsc: Yup, Bastion configuration.

schemas/config/v1/nodePool.json
alejandroEsc: agreed. I completely forgot that etcdConfig is a type of kvStoreConfig.

schemas/config/v1/bastionConfig.json
alejandroEsc: As Patrick mentioned above there are options for bastions, i think i will allow for two types, a more traditional bastion that contains internally a ssh config file and an elb-ed bastion.

ansible/roles/kraken.config/tasks/main.yaml
alejandroEsc: agreed.

ansible/roles/kraken.config/tasks/main.yaml
alejandroEsc: @davidewatson i think i see your point, but preferably I want to simplify the logic and the more complex role filter in the next check, as you will see. Also kubelet will NOT run in a bastion node.

ansible/roles/kraken.config/tasks/main.yaml
alejandroEsc: sure.

ansible/roles/kraken.config/tasks/main.yaml
alejandroEsc: agreed.

schemas/config/v1/bastionConfig.json
alejandroEsc: yep!

schemas/config/v1/definitions.json
alejandroEsc: would this min item requirement mean that this should be listed?

ansible/roles/kraken.fabric/kraken.fabric.selector/tasks/fabric-type.yaml
```diff
@@ -17,7 +17,7 @@
     name: '{{ playbook_dir }}/roles/kraken.fabric/kraken.fabric.weave'
   when: cluster.fabricConfig.type == 'weave'
 
-- name: Error Fabric Type
+- name: Supported Fabric Type Check
```
alejandroEsc: I like these suggestions, and would like that written somewhere as even now i see tasks coming up that I had never seen before with no consistent style.

ansible/roles/kraken.nodePool/kraken.nodePool.selector/tasks/main.yaml
alejandroEsc: yep it is checking for supported docker runtime, please look at error message and when clause.

coffeepac:
@alejandroEsc those two options sound great.  if you want to implement the `bastion` type now and file and issue for `bastion-elb` that works for me to keep scope down on this PR

alejandroEsc:
@coffeepac agreed, will do bastion now and leave elb type in another ticket.

alejandroEsc:
@coffeepac ready for you to look at.

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_bastion_ssh_config.yaml
```diff
@@ -0,0 +1,38 @@
...
+    bastion_hosts_list: "{{ bastion_results.instances | map(attribute='public_dns_name') | list }}"
+  register: bastion_hosts_list
+  when:
+    - bastion_results.instances | length > 0
+
+# Do not want failure to push up ssh config file to be a source of cluster up/down failure
+# could not use 'copy' module since it requires python installed in host.
+- name: Send bastion_ssh_config to bastion hosts
+  with_items: "{{ bastion_hosts_list }}"
+  shell: "scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null {{ config_base | expanduser }}/{{ cluster.name }}/bastion_ssh_config core@{{ item }}:/home/core/.ssh/config"
```
coffeepac: hardcoded user name means we either have to always use ContainerOS for a bastion node (not ideal) or we need to have both the user name and user home directory stored in the OS Config block.  

Best answer for now may be 'file an issue saying this will be broken when we try and support a second OS'.  Might be best to abstract this out now.

coffeepac: Is this going to copy up an ssh config that contains the full list of bastion nodes?

ansible/roles/kraken.ssh/kraken.ssh.aws/templates/ssh_config.jinja2
```diff
@@ -4,11 +4,15 @@
 {% if instance.state == 'running' or instance.state == 'pending' %}
 {% set privateKeyPath = cluster.nodePools | selectattr("name", "equalto", instance.tags['k2-nodepool']) | list %}
 {% if privateKeyPath[0].keyPair.privatekeyFile is defined %}
+{% if instance.public_dns_name is defined and instance.public_dns_name != "" %}
```
coffeepac: I think we need an else clause here where instead of using the public dns entry we would have ssh proxy the command through the bastion host.  An article with an example:  https://blog.scottlowe.org/2015/11/21/using-ssh-bastion-host/

If we do that, maybe we don't need to copy anything to the bastion node and we can punt on the remote user question for now.

joejulian:
Well, it would be if it was on both networks...

ansible/roles/kraken.ssh/kraken.ssh.aws/templates/ssh_config.jinja2
alejandroEsc: That would work in the elb case just fine. However, we have not yet covered that case yet and its not clear why you would hit one bastion node repeatedly and not the others. I would rather, here allow you to have all three.

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_bastion_ssh_config.yaml
alejandroEsc: To answer your first question. Yes, my mistake! i completely forgot that we are going to change OS in the future! will remove that in a bit. I think we can add username and path in the OS config block now actually as part of the OS definition, and we might need it for other things anyway.

This is going to copy a config file that points to each of the private nodes in the network.

alejandroEsc:
@coffeepac Definitely let's chat about this asap. My thoughts are these, based on your good questions:

- We want to have and elb and non-elb case, but now that i think about it we probably should just do the elb case exclusively its probably a lot easier to work with anyway. I havent tried to see what happens when you proxy using an elb but my guess is that it should work. 

- Regardless if we go with elb case or not, the operator may not be the only person that has access to the bastions, and in that case I suggest that each bastion node having a pointer to the private nodes is good practice. If you feel its overly complicated then we can remove it.

ansible/roles/kraken.ssh/kraken.ssh.aws/templates/ssh_config.jinja2
coffeepac: Oooh right.  The bastion nodes are being maintained by asgs.  Right.

coffeepac:
Im not seeing where we are building the ssh config for all private nodes.  The bastion config template/ansible taks looks like its only going to loop over the bastion nodes. So each of the bastion nodes will have an ssh config to get to the other bastion nodes.  

Am i silly misreading that?

alejandroEsc:
Good question, that is done via aws_bastion_ssh_config with the templage bastion_ssh_config via the task `Generate bastion ssh inventory`

coffeepac:
okay, I was misreading the code.  Will add a readability comment in a minute.

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/main.yaml
```diff
@@ -7,18 +7,7 @@
 - include: aws_config.yaml
   when: cluster.providerConfig.provider == 'aws'
 
-- name: Gather inventory of all cluster nodes
```
coffeepac: Please put this task back into main.yaml.  It is used for both aws_ssh_config and aws_bastion_ssh_config.  this works because ansible has a global namespace for registered variables but its still a bit hard to follow.

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_bastion_ssh_config.yaml
```diff
@@ -0,0 +1,21 @@
...
+    src=bastion_ssh_config.jinja2
+    dest="{{ config_base | expanduser }}/{{ cluster.name }}/bastion_ssh_config"
+
+- name: Create a list of bastion host names
+  set_fact:
+    bastion_hosts_list: "{{ bastion_results.instances | map(attribute='public_dns_name') | list }}"
+  register: bastion_hosts_list
+
+- include: aws_bastion_coreos.yaml
+  when: os_type == "coreOs"
```
coffeepac: this works, I don't love it but it works.  the other option is to push the username and path up into the OS config.  I'd be happy with an issue to convert over to config values for the OS specific values as part of getting multi-OS support.

alejandroEsc: The problem with having to put this as part of the OS config is that, and maybe im wrong here, but the user would assume that they can put any value there, including path and it would mean that coreos will have their new username and path there when brought up. Not sure if that is what we would mean by adding it to the schema. Do we have a concept of internal schema vs inflight user schema definitions?

coffeepac: fair, please copy that comment into the linked ticket.  we'll handle it htere.

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/aws_ssh_config.yaml
```diff
@@ -0,0 +1,17 @@
+---
+- name: Generate ssh inventory
+  template:
+    src=ssh_config.jinja2
+    dest="{{ config_base | expanduser }}/{{ cluster.name }}/ssh_config"
+
+- name: Check if we should include bastion nodes
```
coffeepac: name should be 'Find any bastion nodes' and this block should be up a level for readability.  This also means there is only one step in this file so maybe move the template resolution up a level as well.

ansible/roles/kraken.config/files/config-private-topology.yaml
```diff
@@ -0,0 +1,535 @@
...
+              snapshot_id:
+              encrypted: false
+    - &defaultAWSBastionNode
+      name: defaultAWSBastionNode
+      kind: node
+      providerConfig:
+        provider: aws
+        enablePublicIPs: true
+        type: t2.micro
+        subnet: ["zone-1-public", "zone-2-public", "zone-3-public"]
```
joejulian: Am I missing something obvious? Wouldn't this node, then, only have access to the public network and still be isolated from the private network?

alejandroEsc: Nope, As long as both subnets are within the same vpc there should be no problem.

PR #684, opened 2017-07-28
gke default version now 1.6.7

DStorck:


coffeepac:
@DStorck you need to rebase off of master to pick up some fixes or this will never pass.

PR #685, opened 2017-07-28
Etcd Node Name Should be Unique for Discovery

mikeln:
Etcd node discovery via DNS SRV requires that the `--name` parameter had a unique value for each node.

Currently all etcd nodes have the name `etcdetcd` and all events nodes have the name `etcdetcdevents`.
This can result in discovery incorrectly thinking all nodes are discovered, when they are not.  

Exact results of this bug are unclear, however clusters may have been evaluated as "up" prematurely.

### DIFF
--- a/ansible/roles/kraken.nodePool/kraken.nodePool.etcd/templates/etcd.units.etcd.part.jinja2
+++ b/ansible/roles/kraken.nodePool/kraken.nodePool.etcd/templates/etcd.units.etcd.part.jinja2
@@ -34,7 +34,7 @@
      --mount volume=resolv-conf,target=/etc/resolv.conf \
      quay.io/coreos/etcd:{{cluster_node_tuple.1.etcdConfig.version}} \
      -- --data-dir /ephemeral/etcd \
-     --name etcd{{cluster_node_tuple.1.name}} \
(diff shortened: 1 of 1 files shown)


ansible/roles/kraken.nodePool/kraken.nodePool.etcd/templates/etcd.units.etcd.part.jinja2
```diff
@@ -34,7 +34,7 @@
      --mount volume=resolv-conf,target=/etc/resolv.conf \
      quay.io/coreos/etcd:{{cluster_node_tuple.1.etcdConfig.version}} \
      -- --data-dir /ephemeral/etcd \
-     --name etcd{{cluster_node_tuple.1.name}} \
+     --name etcd{{cluster_node_tuple.1.name}}${DEFAULT_IPV4} \
```
alejandroEsc: what does the concatenation of this string look like when evaluated?

mikeln: etcdetcd10.0.139.57
etcdetcdEvents10.0.131.83

the key is that it is unique.

mikeln:
ok ci e2e is failing on the old "namespace taking too long to delete"...

PR #687, opened 2017-07-29
Increase etcd instance sizes to conform with published CoreOS

### LINKED ISSUE #670: Increase etcd and etcdEvents AWS Machines to t2.medium
The split of etcd startup has shown many unsuccessful etcd cluster startups.
An initial thought was to try increasing the AWS machine size.  This seems to have greatly improved the startup reliability.

Recommend changing from t2.small to t2.medium for etcd and etcdEvent.

davidewatson:
recommendations (t2.large). Fixes #670.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,7 +2,7 @@
 github_org             = "samsung-cnct"
 quay_org               = "samsung_cnct"
 
-aws_cloud_test_timeout = 16   // Should be about 16 min (longer due to etcd cluster separation)
+aws_cloud_test_timeout = 32  // Should be about 16 min (longer due to etcd cluster separation)
 gke_cloud_test_timeout = 60  // Should be about 4 min but can be as long as 50 for non-default versions
 e2e_test_timeout       = 18  // Should be about 15 min
 cleanup_timeout        = 60  // Should be about 6 min
--- a/ansible/roles/kraken.config/files/config.yaml
+++ b/ansible/roles/kraken.config/files/config.yaml
@@ -194,7 +194,7 @@ definitions:
       providerConfig:
         enablePublicIPs: true
         provider: aws
-        type: t2.large
+        type: m4.large
         subnet: ["zone-1", "zone-2", "zone-3"]
         tags:
           -
@@ -234,7 +234,7 @@ definitions:
       providerConfig:
         enablePublicIPs: true
         provider: aws
-        type: t2.large
+        type: m4.large
         subnet: ["zone-1", "zone-2", "zone-3"]
         tags:
           -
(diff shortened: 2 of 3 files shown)


davidewatson:
Helm failed us:

> [aws] fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["/opt/cnct/kubernetes/v1.6/bin/helm", "init"], "delta": "0:00:30.559221", "end": "2017-07-29 23:01:46.778717", "failed": true, "rc": 1, "start": "2017-07-29 23:01:16.219496", "stderr": "Error: error installing: Timeout: request did not complete within allowed duration", "stderr_lines": ["Error: error installing: Timeout: request did not complete within allowed duration"]

Restarting.

davidewatson:
Actually, in view of k2/620, maybe we really should move to m4.16xlarge. Since that's too big for many typical clusters, I'm inclined to get this is in anyway. It's a more reliable baseline, and we can increase it again once we tackle the larger cluster scaling recommendations. Thoughts?

mikeln:
Combing etcd issues, network speed seems to be the biggest issue.  The smallest ec2 instance with enhanded networking is m4.large.  
As I recall, all t2 instance types, no matter the size, have "burst" networking.  The "burst" anecdotally seems to contribute to the startup/discovery failure (but no conclusive data on that).
That being said, for dev work where we really only need quorum t2.large may be ok.   Otherwise, m4.large may be the better option.

davidewatson:
I agree. The t2.large proposal was the smallest recommended size. Since the difference in cost between a t2.large and an m4.large is $0.006/per node hour, I will update this PR to use m4.medium.

mikeln:
ok, but m4.large is the first machine with enhanced networking (ie. faster networks).  m4.medium does not have that.

davidewatson:
CI continues to timeout:

> [aws] FAILED - RETRYING: Get all nodes (13 retries left).
[aws] Cancelling nested steps due to timeout
[aws] Sending interrupt signal to process
[aws] Terminated

Increasing the timeout as was done for k2/688.

davidewatson:
After increasing the CI stage timeout, the `kraken.readiness` role is failing.

```
rodin:k2 dwat$ ks get po | grep Error
kube-scheduler-ip-10-0-73-31.ec2.internal              0/1       RunContainerError   0          119d
rodin:k2 dwat$ 
```

and 

`ks log kube-scheduler-ip-10-0-73-31.ec2.internal` shows that there appears to be a leader election problem.

```
I0403 23:26:34.575600       1 leaderelection.go:247] lock is held by ip-10-0-57-226.ec2.internal and has not yet expired
```

Issue #673, opened 2017-07-26
etcd/etcdEvents Cluster Creation Errors


mikeln:
EtcdEvents cluster has periodic problems when starting a new cluster.   
Also, it appears that, if the EtcdEvents cluster was failing to be established...it would never resolve the issue with time (or another up.sh).   You have to tear down and restart the cluster.

cluster error (the ec2 machines were shown as ok in the aws console):
```
mikeln-bender:.kraken Mikel_Nelson$ ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -F /Users/Mikel_Nelson/dev/cloud/k2new/myconfig/.kraken/mikeln-v17/ssh_config etcdEvents-1 etcdctl --ca-file=/etc/etcd/ssl/client-ca.pem --cert-file=/etc/etcd/ssl/client.pem --key-file=/etc/etcd/ssl/client-key.pem --endpoints=https://127.0.0.1:2381 cluster-health
Warning: Permanently added 'ec2-54-201-86-117.us-west-2.compute.amazonaws.com,54.201.86.117' (ECDSA) to the list of known hosts.
cluster may be unhealthy: failed to list members
Error:  client: etcd cluster is unavailable or misconfigured
error #0: client: endpoint https://127.0.0.1:2381 exceeded header timeout
```
vs good cluster:
```
mikeln-bender:.kraken Mikel_Nelson$ ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -F /Users/Mikel_Nelson/dev/cloud/k2new/myconfig/.kraken/mikeln-v17/ssh_config etcdEvents-1 etcdctl --ca-file=/etc/etcd/ssl/client-ca.pem --cert-file=/etc/etcd/ssl/client.pem --key-file=/etc/etcd/ssl/client-key.pem --endpoints=https://127.0.0.1:2381 cluster-health
Warning: Permanently added 'ec2-54-202-34-222.us-west-2.compute.amazonaws.com,54.202.34.222' (ECDSA) to the list of known hosts.
member 41a3e112a65a837c is healthy: got healthy result from https://10.0.78.85:2381
member a2c4d959cd59f4f6 is healthy: got healthy result from https://10.0.63.5:2381
member b7f63ca0829db36f is healthy: got healthy result from https://10.0.234.11:2381
member cd552ef50bda685e is healthy: got healthy result from https://10.0.84.207:2381
[...]

PR #688, opened 2017-07-30
Check etcd cluster health for all nodes in each etcd cluster. This is…

davidewatson:
… related

to #673, but in part because of #672, it is not expected to fix either of them.
It only ensure traditional etcd clusters (with public IP addresses) should
start more reliably (all etcd nodes must be healthy, not a quorum of them).

It also includes an unconditional wait for clusters for which we have no way of determining etcd health.

### DIFF
--- a/ansible/roles/kraken.provider/kraken.provider.aws/tasks/wait-for-etcd.yaml
+++ b/ansible/roles/kraken.provider/kraken.provider.aws/tasks/wait-for-etcd.yaml
@@ -22,9 +22,19 @@
   with_items: "{{ item.count }}"
   loop_control:
     loop_var: idx
-  # See k2/674 for notes on a long term fix to eliminating this when statement.
-  when: item.nodeConfig.providerConfig.enablePublicIPs is undefined or item.nodeConfig.providerConfig.enablePublicIPs
+  # See k2/674 for notes on the correct fix to eliminating this when statement.
+  when: item.nodeConfig.providerConfig.enablePublicIPs is undefined or
+        item.nodeConfig.providerConfig.enablePublicIPs
 
+# When we have no other way to determine etcd cluster health, wait the maximum
+# expected time. See k2/674 for notes on the correct fix to eliminating this
+# pause.
+- name: Pause when traditional etcd health checks are unavailable
+  pause:
+    seconds: "{{ (etcd_retries * etcd_delay) | int }}"
+  when:
+    - item.nodeConfig.providerConfig.enablePublicIPs is defined
+    - not item.nodeConfig.providerConfig.enablePublicIPs
 - debug:
     var: etcd_result
     verbosity: 0
--- a/ansible/roles/kraken.provider/kraken.provider.aws/tasks/wait-for-etcd.yaml
+++ b/ansible/roles/kraken.provider/kraken.provider.aws/tasks/wait-for-etcd.yaml
@@ -9,7 +9,7 @@
   command: >
     ssh -o StrictHostKeyChecking=no
         -o UserKnownHostsFile=/dev/null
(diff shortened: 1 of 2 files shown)


coffeepac:
does this need an AWS timeout increase?  it timeout termed after five checks waiting for nodes to show up.  normally it dies around 25 checks.

davidewatson:
I suspect it is occurring 5 times, once for each etcd node. I think it's a problem with the use of the `when` statement here.

I also think we need to add something along the lines of this:

```
- name: Fail if any etcd node failed to report healthy
  fail:
    msg: >
      An etcd node failed to report a healthy status. Please run down and
      reattempt to run up again.
  when: etcd_result | failed
```

This is aggressive but will force the user always use a new uuid when creating a new etcd cluster. It also means that only cluster where the etcd cluster comes up on the first attempt will succeed. If that is not almost always the case, I think we need to find out why not and fix it.

davidewatson:
So [this](https://common-jenkins.kubeme.io/job/K2/job/PR-688/4/consoleFull) [1] run passed. I'm pessimistic. Let's try again.

[1]

```
[aws]                     "member 5ce0e8847216e738 is healthy: got healthy result from https://10.0.101.127:2379", 
[aws]                     "member 68b4e418b224d836 is healthy: got healthy result from https://10.0.62.134:2379", 
[aws]                     "member e6e1708c869d5ffd is healthy: got healthy result from https://10.0.85.193:2379", 
[aws]                     "member ec6de10a597f2e55 is healthy: got healthy result from https://10.0.226.6:2379", 
[aws]                     "member fcf0cc00c2eb7d5d is healthy: got healthy result from https://10.0.163.34:2379", 
```

and

```
[aws]                     "member 8b72fa12aa2f32 is healthy: got healthy result from https://10.0.218.35:2381", 
[aws]                     "member 39f7fc1595df65d9 is healthy: got healthy result from https://10.0.138.22:2381", 
[aws]                     "member 87e71a4ebce1e7b3 is healthy: got healthy result from https://10.0.24.182:2381", 
[aws]                     "member 91032f2fd513f5f5 is healthy: got healthy result from https://10.0.123.175:2381", 
[aws]                     "member f590624058c6b90c is healthy: got healthy result from https://10.0.74.151:2381", 
```

davidewatson:
PR-688/5 failed due to Cloud test stage timing out.

> [aws] FAILED - RETRYING: Get all nodes (14 retries left).
[aws] Cancelling nested steps due to timeout
[aws] Sending interrupt signal to process

I've seen this a few times in other PRs. We might want to increase it.

ansible/roles/kraken.provider/kraken.provider.aws/tasks/wait-for-etcd.yaml
```diff
@@ -19,8 +19,30 @@
...
+  loop_control:
+    loop_var: idx
+  # See k2/674 for notes on the correct fix to eliminating this when statement.
+  when: item.nodeConfig.providerConfig.enablePublicIPs is undefined or
+        item.nodeConfig.providerConfig.enablePublicIPs
+
+- name: Fail if any etcd node failed to report healthy
+  fail:
+    msg: >
+      An etcd node failed to report a healthy status. Please run down and
```
alejandroEsc: would be nice to have the node that failed, just the name should suffice in error message

davidewatson: Good call. I've moved the debug task before the failure task.

davidewatson:
Runs # 4 and # 7 passed. Run # 8 successfully passed `Test: Unit` and `Clean up`, but failed one of the e2e tests: `Unable to connect to the server: unexpected EOF`.

PR #690, opened 2017-07-30
Add (and make default) a uuidToken option to use uuidgen to force

### LINKED ISSUE #574: k2 etcd cluster names and tokens may not be unique

davidewatson:
each cluster formation attempt to occur exactly once. Fixes #574.


davidewatson:
I wonder if https://github.com/samsung-cnct/k2/pull/685 is necessary with this PR. I don't think it can hurt, but if we use these names for any administrative purpose, they are going to be a pain to verify and type.

Looking at https://coreos.com/etcd/docs/latest/v2/clustering.html (old, I know, but I think still default for our configs),

1) For DNS discovery (SRV) the names must match one of those in the SRV record. Why not just make the names the same as the hostnames (e.g. etcd-1, etcd-2, etc.) so that they never have to change?

2) For public discovery,

> Each member must have a different name flag specified. Hostname or machine-id can be a good choice. Or discovery will fail due to duplicated name.

But we don't use it that public discovery.

I'd like to think further about the ramifications of including IP addresses in the name. It can be useful if the IP address has meaning (e.g. the physical position of a node in a rack), but in the cloud, that's rarely the case.

davidewatson:
@joejulian pointed out that instead of using uuidgen, we could rely on the ansible [password module](http://docs.ansible.com/ansible/latest/playbooks_lookups.html#the-password-lookup) instead. As long as we limit the character set, that seems reasonable to me.

schemas/config/v1/etcdConfig.json
```diff
@@ -25,10 +25,15 @@
       "default": [ 2379, 4001 ]
     },
     "clusterToken": {
-      "description": "Token used during etcd cluster formation",
+      "description": "Token used during etcd cluster formation. This may be unsafe if an etcd cluster formation has to be restarted, or multiple clusters are created within the same VPC. It is highly recommended that the uuidToken option be used instead.",
       "type": "string",
       "default": "espouse-monger-rarely"
     },
+    "uuidToken": {
```
leahnp: So if we want to use this option what do we put in the config? If it's a token, where/how do we get/generate it?

davidewatson: You are right that the option may be strangely named. `uuidToken` was chosen to parallel the existing `clusterToken`. The semantics are different insofar as `uuidToken` indicates a procedure which will be used to generate the token, and not the token itself.

I have changed the type of the property and updated the schema description to clarify. Does this suffice?

leahnp: Yep that makes sense - thanks!

Issue #631, opened 2017-07-16
Ensure user taints are settable when provider is GKE


coffeepac:
Make sure that if a user selects the GKE provider and they set taints for a given nodepool that these taints are applied.

DStorck:
I'm trying to figure out if [this](https://github.com/samsung-cnct/k2/blob/2ab10e227d12e7e2e16664a4c24480afe28f9261/ansible/roles/kraken.provider/kraken.provider.gke/templates/deploymentmanager-create.yaml.jinja2) is the right place to start digging around

DStorck:
fixed with https://github.com/samsung-cnct/k2/pull/693

leahnp:
LGTM

PR #693, opened 2017-07-31
added gke taint options at cluster creation

DStorck:


DStorck:
fixes https://github.com/samsung-cnct/k2/issues/631

leahnp:
LGTM

PR #705, opened 2017-08-03
updated drunkensmee version to 0.6

leahnp:


davidewatson:
We really should set a fact so that this doesn't have to be updated in three places, but since this fixes an existing customer problem...

/lgtm

coffeepac:
needed customer feature.  in it goes.

PR #709, opened 2017-08-04
Move helm and kubernetes command configuration to config

joejulian:
This moves duplicated code which uses the found versions to choose which `helm` and `kubectl` executable path to use to configure so it need only be run once.

* Fixes DRY issues
* Fixes linting errors
* Allows for frigid temperatures in the underworld
* Modified `while` `and`s to be lists for readability

### DIFF
--- a/ansible/roles/kraken.config/tasks/main.yaml
+++ b/ansible/roles/kraken.config/tasks/main.yaml
@@ -123,6 +123,8 @@
 - include: helm-versioning.yaml
   with_items:
     - "{{ kraken_config.clusters }}"
+  loop_control:
+    loop_var: a_cluster
 
 - include: check-update.yaml
   with_items:
--- a/ansible/roles/kraken.config/tasks/main.yaml
+++ b/ansible/roles/kraken.config/tasks/main.yaml
@@ -75,76 +75,10 @@
          of API version {{ kraken_config_user.version }}.
   when: not validation.result.is_valid
 
-- name: Verify all clusters in the clusters list have names
-  fail:
-    msg: "Make sure that all 'clusters.name' are defined in {{ config_file }}"
-  when: item.name | is_empty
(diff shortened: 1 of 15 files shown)


ansible/roles/kraken.services/tasks/aws_config.yaml
```diff
@@ -6,13 +6,13 @@
...
 - name: set accessKey if unset and credentialsFile set
-  set_fact: 
+  set_fact:
     cluster: "{{ cluster |combine({'providerConfig':{'authentication':{'accessKey': lookup('ini', 'aws_access_key_id section=' + cluster.providerConfig.authentication.credentialsProfile + ' file=' + cluster.providerConfig.authentication.credentialsFile)}}}, recursive=True)}}"
   when:
     cluster.providerConfig.authentication.accessKey | is_empty and cluster.providerConfig.authentication.credentialsFile | is_empty == False
 
 - name: set accessSecret if unset and credentialsFile set
-  set_fact: 
+  set_fact:
```
davidewatson: In new code I've been writing I've tried to combine `set_fact` when appropriate.

ansible/roles/kraken.config/tasks/main.yaml
```diff
@@ -153,4 +89,10 @@
   template:
     src: query_max_version.jinja2
     dest: "{{ version_outfile | realpath }}"
-  when: kraken_action == 'max_k8s_version'
\ No newline at end of file
+  when: kraken_action == 'max_k8s_version'
+
+- include: cluster_loop.yaml
```
davidewatson: The name `cluster_loop` tells me how but not what. When I look at the yaml, it does two things: 1) checks for invalid configs and 2) sets common variables. Unfortunately, I'm having a hard time coming up with a better name without additional refactoring.

PR #712, opened 2017-08-04
upgrade services on k2 update

### LINKED ISSUE #453: Up should not be killing & rebuilding tiller
Up should be idempotent.  Update will have to kill and rebuild it, however.
See kill-services.yaml

joejulian:
Make helm upgrade rather than delete and install.
* Only run kill-services on `down`
* Add `manage-services` to be run on `update` instead
* Upgrade tiller as needed to match helm version

Fixes #453

### DIFF
--- a/ansible/roles/kraken.services/tasks/main.yaml
+++ b/ansible/roles/kraken.services/tasks/main.yaml
@@ -29,13 +29,6 @@
     - a_cluster.providerConfig.provider in ['aws']
   static: no
 
-- include: kill-services.yaml
-  with_items:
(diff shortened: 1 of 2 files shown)


joejulian:
Yes, I think a separate issue could be replacing most of run-services with manage-services, but that's not something I have time to test during this sprint. Currently there are distinct differences for most of the commands.

coffeepac:
That issue needs to be filed.  I'll go do it.

coffeepac:
cleanup:  #716

PR #720, opened 2017-08-08
initial commit for github project support docs

coffeepac:
initial cut for these files.  I'm looking for lots of feedback for
all of them.  This includes placing yourself as a primary owner
for parts of K2.  Right now the one file type I have is a stub to
show how its done.


coffeepac:
open question:  should the labelling guide (K2, feature request/bug, etc) be part of the CONTRIBUTING.md or get pushed into the default issue text that K2 doesn't have yet?

CODEOWNERS.md
```diff
@@ -0,0 +1,20 @@
...
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in the repo.
+*       @coffeepac
+*       @davidewatson
+*       @mikeln
+*       @dstorck
+*       @leahnp
+*       @yenicapotediaz
+*       @guineveresaenger
```
spiffxp: You may want to play with this a bit.  As-is I suspect @guineveresaenger is getting everything.

Putting everyone on one line, or using a GitHub Team may have the intended effect, or it may end up assigning _everyone_ instead of a random subset.

CONTRIBUTING.md
```diff
@@ -0,0 +1,85 @@
+# How to contribute
+We love open source contributions and want to make it as easy and painless
+as possible for you to contribute to this project.  To enable that we have
+laid out a light-weight contribution process.
```
spiffxp: lightweight

CONTRIBUTING.md
```diff
@@ -0,0 +1,85 @@
...
+
+The basis of this doc was shamelessly swiped from the puppet repo.
+
+## Getting Started
+
+* Make sure you have a [GitHub account](https://github.com/signup/free)
+* Submit a ticket for your issue, assuming one does not already exist.
```
spiffxp: ticket isn't github terminology, I would say "open a github issue"

which repo should the issue be opened against? this repo? the repo this file exists in? the issues repo?

how should I make sure an issue doesn't already exist?

CONTRIBUTING.md
```diff
@@ -0,0 +1,85 @@
...
+  * Clearly describe the issue including steps to reproduce when it is a bug.
+  * Make sure you fill in the earliest version that you know has the issue.
```
spiffxp: I'm assuming you mean "version of kraken2" in this case

CONTRIBUTING.md
```diff
@@ -0,0 +1,85 @@
...
+* Fork the repository on GitHub
+
+### Looking for an easy way to get started? ###
+We label easy to complete issues as 'Help Wanted'.  These issues require minimal
```
spiffxp: [link to a query for help wanted](https://github.com/samsung-cnct/k2/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22)

CONTRIBUTING.md
```diff
@@ -0,0 +1,85 @@
...
+context for Kraken2 and shouldn't require too much time.  If you do start working
+on one of these let us know so we can make sure no one on the core team picks it
```
spiffxp: let us know by commenting on the issue (or assigning to yourself if capable?)

CONTRIBUTING.md
```diff
@@ -0,0 +1,85 @@
...
+    document is not a concrete example.  This is a problem because the
+    contributor is left to imagine what the commit message should look like
+    based on a description rather than an example.  This patch fixes the
+    problem by making the example concrete and imperative.
+
+    Fixes: #<github issue number filed to describe the problem>
+````
+
+* Make sure you have added the necessary tests for your changes.
+* Run _all_ the tests to assure nothing else was accidentally broken.
```
spiffxp: How do I do this?

CONTRIBUTING.md
```diff
@@ -0,0 +1,85 @@
...
+    Without this patch applied the example commit message in the CONTRIBUTING
+    document is not a concrete example.  This is a problem because the
+    contributor is left to imagine what the commit message should look like
+    based on a description rather than an example.  This patch fixes the
+    problem by making the example concrete and imperative.
+
+    Fixes: #<github issue number filed to describe the problem>
+````
+
+* Make sure you have added the necessary tests for your changes.
```
spiffxp: What are necessary tests?  Tests that prove the problem existed and is now fixed?  Additional code coverage?

CODEOWNERS.md
```diff
@@ -0,0 +1,20 @@
...
+
+# Order is important. The last matching pattern has the most precedence.
+# So if a pull request only touches javascript files, only these owners
+# will be requested to review.
+*.sh    @joejulian
+
+
+# You can also use email addresses if you prefer.
+#docs/*  docs@example.com
```
mikeln: I assume this has a similar format as a .gitignore...?  I also assume we just want one of these files in the root directory of the project (vs as needed in each subdir).   I assume the selection goes from most specific to least.  We should be able to create a preferred ordered list for each ansible role.   Do you want to fill out this file, or have each person do a separate PR?

CONTRIBUTING.md
```diff
@@ -0,0 +1,85 @@
...
+````
+    Make the example in CONTRIBUTING imperative and concrete
+
+    Without this patch applied the example commit message in the CONTRIBUTING
+    document is not a concrete example.  This is a problem because the
+    contributor is left to imagine what the commit message should look like
+    based on a description rather than an example.  This patch fixes the
+    problem by making the example concrete and imperative.
+
+    Fixes: #<github issue number filed to describe the problem>
```
spiffxp: Putting the github issue ref in a commit means every push or rebase of that commit is going to trigger a new event on the issue.  It looks noisy on the receiving end.  I'm necessarily saying do-not-do-this, but it's frowned upon in kubernetes.

Also, it only takes [one issue ref with the right keyword](https://help.github.com/articles/closing-issues-using-keywords/) to close an issue, which is sometimes unintended, eg: "partially fixes #foo" will still close #foo automagically.  This is the case regardless of whether the ref is in a PR description or a commit.

CONTRIBUTING.md
```diff
@@ -0,0 +1,85 @@
...
+* Create a topic branch from where you want to base your work.
+  * This is usually the master branch.
+  * To quickly create a topic branch based on master; `git checkout -b
+    fix/master/my_contribution master`. Please avoid working directly on the
+    `master` branch.
+* Make commits of logical units.
+* Check for unnecessary whitespace with `git diff --check` before committing.
+* Make sure your commit messages are in the proper format.
+
+````
```
spiffxp: nit: I think you only need three backticks

CONTRIBUTING.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Documentation
+
+For changes of a trivial nature to comments and documentation, it is not
+always necessary to create a new issue in github. In this case, it is
+appropriate to start the first line of a commit with '(doc)' instead of
+a ticket number.
+
+````
+    (doc) Add documentation commit example to CONTRIBUTING
```
spiffxp: bikeshed color: kubernetes/charts uses [foo] instead of (foo) for its commit prefixes

SUPPORT.md
```diff
@@ -0,0 +1,5 @@
+To get support (bug fixes, feature requests or basic guidelins)
```
spiffxp: guidelines

CONTRIBUTING.md
spiffxp: who is the core team? how do I know if one of them has picked it up?

CODEOWNERS.md
coffeepac: its a (gitattribute)[https://git-scm.com/docs/gitattributes] format.  I think its the format `.gitignore` uses.

For this PR I'm fine with people signing up for roles they know very well and would want to maintain some level of quality in.  my initial thoughts on this are you for DNS, @joejulian for the dockerdev/lib folder in general, watson for the etcd work going on, etc.  However, i didn't want to sign anyone up without them knowing first.

CONTRIBUTING.md
coffeepac: issues repo is private, so this repo.

CONTRIBUTING.md
coffeepac: I actually should have removed that line.  its gone soon.

CODEOWNERS.md
mikeln: Reading the github doc on this..  I think you are right.  Need to play with this a bit for sure.  
The doc states "later precedence" (more in terms of specificity but still), also everyone listed on one line will be assigned to review.

CONTRIBUTING.md
coffeepac: Huh, I thought the kubernetes issue check was for this.  what's the preferred way then?  I'm going to defer to kubernetes policy as much as is reasonable and this seems reasonable.

CODEOWNERS.md
mikeln: yes, for me:  either kraken.services  or specific to the DNS files (which may be problematic), also kraken.access, flannel/canal fabric (maybe just all fabric), kraken.rbac (default rbac setup).   Secondary reviewer (if that is possible) on etcd.  Also, how do we break up schema validation... I assume it would actually have to be by file for the area of responsibility.

coffeepac: I think the finest granularity I'm okay with is Roles.  Anything less then that is going to cause too much churn.  

Schema: I'd like this to be whoever is up for it.  I am.  If its me that's fine.  But I don't want to split the schema up a ton, see above point for churn in the file.

joejulian: @mikeln Least specific to most, as I read it.
@coffeepac Yes for me.

mikeln: So, interesting Q I suppose... If someone makes a change to roleA, configA, and schemaA, will github assign potentially 3 reviewers for the PR based on these files?  I assume so.

CONTRIBUTING.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Making Changes
+
+* Create a topic branch from where you want to base your work.
+  * This is usually the master branch.
+  * To quickly create a topic branch based on master; `git checkout -b
+    fix/master/my_contribution master`. Please avoid working directly on the
+    `master` branch.
+* Make commits of logical units.
+* Check for unnecessary whitespace with `git diff --check` before committing.
```
joejulian: I don't know if you want to put this here, but if you install [pre-commit](http://pre-commit.com/) some simple linting will be done pre-commit as the configuration for it already exists in tree.

spiffxp:
meta comment: @coffeepac is doing an awesome thing by responding to feedback with additional commits instead of rebase/squashing over the old comment... fixup commits are noise when it comes to sifting through cinnut logs for historical context, and so are preferably squashed away prior to merge, but while a PR is being reviewed they play much more nicely with github's UI

CONTRIBUTING.md
spiffxp: So kubernetes prefers the issue link in a PR description, vs. directly in the commit message, to avoid the noise.

Why does kubernetes requires PR's to link to issues? We're, uh, [still having that debate](https://github.com/kubernetes/test-infra/issues/3704). It was supposed to help the release team keep track of release notes / of why PR's were going into a release.  It is being protested as needless friction by some / circumvented with `/approve no-issue` comments, and the issue links aren't actively being consumed by docs-generating tools at present.

Whether or why to require them is up to you.  Using "an issue must exist and be linked" as some amount of friction/context for larger, multi-PR efforts seems like a good idea to me.

Generally if you don't intend to close an issue, a nondescript link might look like "ref: org/repo#xyz"

CODEOWNERS.md
mikeln: Along this line, we should split what kubernetes may in the future call "add ons" out from the kraken.services role into their own role (e.g. kraken.addons).  Currently that would only be the DNS items.

CODEOWNERS.md
DStorck: the file is supposed to just be CODEOWNERS, does it matter that we have it as a markdown file?

CODEOWNERS
```diff
@@ -0,0 +1,21 @@
+# Lines starting with '#' are comments.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in the repo.
+*       @coffeepac, @davidewatson, @mikeln, @dstorck, @leahnp, @yenicapotediaz,@guineveresaenger
+
+# Order is important. The last matching pattern has the most precedence.
+*docs/              @davidewatson
```
spiffxp: you probably want `docs/**`, which is everything in docs, and everything in every subdirectory of docs

`docs/*` is everything just in docs, no subdirectories

you probably want the same pattern for the rest of these entries

joejulian: Actually, I think that no subdirectories may be the intention. The generated docs may live in a subdirectory of docs and thus needs no owner.

joejulian: But either way, probably want the leading slash.

`docs/**` isn't documented, so I'm unsure if that's correct
`docs/` is _any_ docks directory in _any_ subdirectory
`docs/*` is any _file_ in the docs directory
`/docs/` is anything under the root's docs directory

spiffxp: https://git-scm.com/docs/gitattributes points me to https://git-scm.com/docs/gitignore which has a section on the use of `**`.  It is long.  I agree, `/docs/` seems pretty unambiguous and simple.

notjames:
I've reviewed all of this and it all looks good. I am hesitant to do anything more than say LGTM, unless you want an official review from me.

yenicapotediaz:
LGTM

PR #732, opened 2017-08-09
if not on master, e2e failure is unstable but gogogo

coffeepac:


coffeepac:
@mikeln the last CI run had a failed e2e run, the build was marked as UNSTABLE but I haven't been able to find a way for github to mark that as anything but a failing check.  Ideas?  I would like there to be some indication that e2e failed.

mikeln:
The issue is github only allowing pass/fail statuses?
What about adding a 2nd PR check (that isn't required).   
So:
1) take out e2e in the "Required" check
2) add a second "non required" check that just runs e2e/conformance.

So the question becomes, how to do that without completely spinning up a new cluster for the 2nd step...or maybe we don't care how many we run?

coffeepac:
correct.  github only allows pass/fail.  

I wonder if its possible to manually report just the status of e2e, make it show up as its own check.  if that seems like less a few hours of work I'll do it.

I'd prefer not spinning up a second cluster mostly because that's so unreliable it would probably make things worse, not better.

mikeln:
funny, you beat me to it...edited my response with the same idea.

mikeln:
failing of no helm version supported.  perhaps due to this recent merge k2-tools 102?

coffeepac:
yep, working on a k2 fix.

Jenkinsfile
```diff
@@ -162,4 +168,14 @@ def customContainer(String name, Closure body) {
   }
 }
 
+void setBuildStatus(context, message, state) {
+    step([
+        $class: "GitHubCommitStatusSetter",
+        contextSource: [$class: "ManuallyEnteredCommitContextSource", context: context],
+        errorHandlers: [[$class: "ChangingBuildStatusErrorHandler", result: "UNSTABLE"]],
+        reposSource: [$class: "ManuallyEnteredRepositorySource", url: git_uri],
+        statusResultSource: [ $class: "ConditionalStatusResultSource", results: [[$class: "AnyBuildResult", message: message, state: state]] ]
```
mikeln: git_uri is not in scope here... it comes up blank.  You'll probably need to duplicate the code above to set that.

coffeepac:
unable to send messages back to github from the slave as the credentials for github aren't being auto-loaded in.  need to figure out how to get the credentials into the call and then how to get the credentials onto the slave pod.

coffeepac:
attempted to run the notify step on master in the hopes that would allow the class to find the credentials, no dice.  

going to try this plugin now:  https://plugins.jenkins.io/pipeline-githubnotify-step

coffeepac:
The listed plugin does what I want.  Need to tweak the overall pipeline to report success no matter the e2e status.

coffeepac:
keeps coming back unstable.  huh.

coffeepac:
the junit plugin will always set the build status to UNSTABLE if one or more tests have failed.  the healthScaleFactor deals with the overall health of the build, which is the health of the job over the last several runs.  

the github notification from jenkins is only for the last run of the job, which is UNSTABLE which github considers to be a failure.  the job being set to UNSTABLE occurs [here](https://github.com/jenkinsci/junit-plugin/blob/junit-1.21/src/main/java/hudson/tasks/junit/JUnitResultArchiver.java#L196).  There doesn't seem to be a way to override this via configuration. 

I believe the way forward for this is:
- compute our own overall build health and make it required
- make the default notification not required
- make the e2e notification not required

oneilcin:
Are there any e2e tests we do want to fail on?  We could separate out the critical ones and fail on them.

coffeepac:
@oneilcin we haven't spent any time going through the e2e conformance tests to determine if that's the case.  right now its a black box of tests.

PR #739, opened 2017-08-10
allow for helm check to always pass.  

coffeepac:
default will be to use helm if available but the override env var will allow the check to pass if helm is not present by not running helm

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,8 +46,8 @@ podTemplate(label: 'k2', containers: [
             // Dry Run Test
             withEnv(["helm_override_`echo ${JOB_BASE_NAME}-${BUILD_ID} | tr '[:upper:]' '[:lower:]' | tr '-' '_'`=false"]) {
                 stage('Test: Dry Run') {
-                    kubesh 'export helm_override_`echo ${JOB_BASE_NAME}-${BUILD_ID} | tr "[:upper:]" "[:lower:]" | tr "-" "_"`=false env'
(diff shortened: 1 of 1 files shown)


mikeln:
wait... is helm override supposed to be "true" to allow it to continue.. and "false" if you care about the version?

Jenkinsfile
```diff
@@ -44,8 +44,11 @@ podTemplate(label: 'k2', containers: [
                 kubesh "build-scripts/update-generated-config.sh cluster/gke/config.yaml ${env.JOB_BASE_NAME}-${env.BUILD_ID}"
         }
             // Dry Run Test
-            stage('Test: Dry Run') {
-                kubesh 'PWD=`pwd` && ./bin/up.sh --config $PWD/cluster/aws/config.yaml --output $PWD/cluster/aws/ -t dryrun'
+            withEnv(["helm_override_`echo ${JOB_BASE_NAME}-${BUILD_ID} | tr '[:upper:]' '[:lower:]' | tr '-' '_'`=false"]) {
+                stage('Test: Dry Run') {
+                    kubesh 'env'
+                    kubesh 'export helm_override_`echo ${JOB_BASE_NAME}-${BUILD_ID} | tr "[:upper:]" "[:lower:]" | tr "-" "_"`=false && PWD=`pwd` && ./bin/up.sh --config $PWD/cluster/aws/config.yaml --output $PWD/cluster/aws/ -t dryrun'
```
mikeln: I don't think you want "&&" in there... that is 2 statements.  without it the env will directly apply to that call to up.sh

PR #741, opened 2017-08-11
moving back to kraken

coffeepac:
only changing the pretty name of things for now.  If we want to use the kraken repo name we'll need to move the old kraken and then probably move the quay image and that requires a lot more work.

first cut, probably missing something.  will do k2-tools similarly if this is good.

open question:  k2cli.  I've left as is for now but could see krakencli.


joejulian:
In my opinion, k2cli should be kraken if that's _The_ interface. k2 would then be krakenlib (or something like that).

mikeln:
I think we k2cli should go to  krakenctl  or krakenadm which are more in line with the kuberenetes community.   I don't like krakenlib...it's isn't a library, maybe some reference more as a framework. e.g. krakenfwk.   But really I think kraken is fine.

alejandroEsc:
krakenctl = poseidon or maybe zeus?

alejandroEsc:
zeus release-the --config=<path_to_kraken_config>

coffeepac:
I like krakenctl personally

yenicapotediaz:
+1 to krakenctl

DStorck:
krackle?

davidewatson:
Taken as a given that we are not going to use the k2 name, I think `kraken` and `krakencli` make sense. Given all of the existing references to `k2`, I think we should continue using it internally and only change the repo, documentation, and command names for discoverability.

joejulian:
Sure, we like names. From the perspective of someone who's heard of kraken, searches for it, and wants to use it - you're creating confusion if you don't really want people using kraken but would rather have people use krakencli. Make the thing you want people to use have the name of the thing.

joejulian:
Oh, and as far as "it isn't a library", why isn't it? It's the bit that's included by k2cli to make k2cli work. If it's not a library of functions called from the user interface, then I'll eat a hat (one built to my exact specifications, not just some random hat).

davidewatson:
@joejulian: Fair point. I could live with `kraken` and `krakenlib`. I also agree with @mikeln that `krakenctl` is more natural.

guineveresaenger:
Regarding @joe's comment - I don't think krakenctl is worse than kraken, because I don't use `aws`, I use `awscli` in my terminal and that is fine and expected and confers meaning on what it is and what it does.

joejulian:
@guineveresaenger but you _do_ use aws. The _product_ that is AWS is the cloud service. One of the tools you can use to manage that product is the aws cli (also called `aws` according to the web page https://aws.amazon.com/cli/ ). Another is the gui web interface but it's recognizable that the thing that is aws is the _service_.

What is kraken? It's not a service, it's a tool. If you want to turn a bolt, you look for a wrench. You don't to looking for a wrenchhandle.

alika:
+1 to `krakenctl` or maybe `krakenktl` and then it's still kinda k2.

_Sent from my Motorola XT1585 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

guineveresaenger:
@joe point taken about kraken being a tool but I'm just going on the "name" vs "name+cli" situation we have currently.

guineveresaenger:
One thing I liked about k2cli is that it's easy and short to type. Is length of command a criterion in its ease of use? krakencli has almost twice the number of keystrokes.

coffeepac:
we don't need to maintain the same 'name + cli' paradigm.  

I'm starting to lean towards @joejulian suggestion of making the cli named `kraken`.  the current k2 repo is going to be undergoing a lot of change and will probably get split up as things progress.

alejandroEsc:
@davidewatson and myself have had this conversation in a serious note before. We both kinda came to the same conclusion, the cli tool should be kraken, and k2 can be whatever it is, just as long as we advertise it as kraken code base.

mikeln:
I really dislike the name krakenlib.  It strikes me too much of a .lib that gets linked into anything I build.  
It is more of a framework than a lib.   I'll recommend: `krakenfwk` instead.

coffeepac:
I like calling it a framework more then a library as well.  I expect there will be several pieces to this when its running:
- a container that gets excuted locally
- a golang interface (actual lib)
- a long-running service that gets deployed to the cluster

that is all future state so I'd like the name to give some hint about where we're going.

coffeepac:
there were several people (in slack) who have a stated preference for `krakenlib` over `krakenfwk`.  This needs to be decided by, oh, end of day Friday 8/18.  Gotta name it.

mikeln:
put up a companywide slack vote

alejandroEsc:
well lets get a few options first

alejandroEsc:
kraken-mantle <- mantle is the head part of a squid
https://en.wikipedia.org/wiki/Mantle_(mollusc)

alika:
I'll submit [cephalopod](https://en.wikipedia.org/wiki/Cephalopod), but as it stands right now a slack vote to rename k2 would look like:
1. `krakenlib`
2. `krakenfwk`
3. `kraken-mantle`

coffeepac:
leaning towards `kraken-lib` at this point.  works well with `kraken-tools`.

CONTRIBUTING.md
```diff
@@ -15,7 +15,7 @@ The basis of this doc was shamelessly swiped from the puppet repo.
 
 ### Looking for an easy way to get started? ###
 We label easy to complete issues as 'Help Wanted'.  These issues require minimal
-context for Kraken2 and shouldn't require too much time.  If you do start working
+context for Kraken-lib and shouldn't require too much time.  If you do start working
```
davidewatson: When referring to a repo I think the name should follow the capitalization of the repo. Some say repo names should be lower case.

DEVELOPERS.md
```diff
@@ -95,27 +95,27 @@ GitHub.
...
 ### Recommended Dev Tooling 
-The recommended K2 Developer Tool can be accessed in the `/hack` directory and is called `dockerdev`, a tool that pulls down a  [k2-tools](https://github.com/samsung-cnct/k2-tools) image. 
-The container will provide you with a console that contains the correct environment for developing and running k2. The intent is to mitigate version issues and 
+The recommended Kraken-lib Developer Tool can be accessed in the `/hack` directory and is called `dockerdev`, a tool that pulls down a  [Kraken-tools](https://github.com/samsung-cnct/k2-tools) image. 
+The container will provide you with a console that contains the correct environment for developing and running Kraken-lib. The intent is to mitigate version issues and 
 requirements, allowing you the Developer to focus on coding and not worry on dependencies.
 
-#### Developer Workflow with K2-Tools
-Here we document how a developer would use k2-tools to work with a cluster.
+#### Developer Workflow with Kraken-Tools
```
davidewatson: So this is interesting. Here we capitalized the K in Kraken and the T in Tools. This is inconsistent with `Kraken-lib` above. I suggest we lower case all of it, though we could upper case the first character of each word everywhere. Do you have an opinion @NancyHarvey?

docs/_sources/index.rst.txt
```diff
@@ -1,9 +1,9 @@
-.. k2 documentation master file, created by
+.. Kraken documentation master file, created by
```
davidewatson: I need to go back and fix the autogeneration of this files. Instead of adding history, should we skip the documentation in `docs/`?

CONTRIBUTING.md
coffeepac: I'm of no particular opinion on this.  Happy to go either way.

DEVELOPERS.md
coffeepac: holdover from how k2-Tools was named.  should probably be Kraken-tools (or kraken-tools pending comment above)

docs/_sources/index.rst.txt
coffeepac: should probably ignore them for now, will need a follow-up ticket to fix it.

davidewatson: #789 has been opened.

Issue #620, opened 2017-07-13
K2 defaults should conform to community large cluster config


countspongebob:
Community documentation here:

https://github.com/kubernetes/community/blob/master/sig-scalability/provider-configs.md

For the control plane config:
1) Split etcd
2) 5 node config 
3) Note instances sizes specified

This issue is specific to the control plane config, not the number of nodes brought up, which could also use some discussion.

countspongebob:
@lostlivio @jandrake  Opened per our conversation.

davidewatson:
1. and 2. are currently the case, however, the default k2 configuration uses `m4.large` for master nodes instead of `m4.16xlarge`. That equates to an 8x increase in memory and a 16x increase in cores.

I am not sure the defaults should assume such a large k8s installation. Our CI system is overwhelmingly the largest consumer of k2 defaults, and in that case, there are only 10 worker nodes. Instead, it might be faster and cheaper to have a weekly scalability CI test which specifically tests the proposed configuration.

countspongebob:
The default I'm requesting is smaller than the current default.

Alternatively, a single-node non-HA system would be reasonable.

countspongebob:
@davidewatson Now I see your concern. My primary issue was the size of the control plane config, not the default number of started nodes.

davidewatson:
Another area k2 defaults differ from the SIG scale testing proposal is 100GB vs. 256 GB etcd storage partitions for etcd.

countspongebob:
The 5 node cluster is total. Etcd configs share the same 5 nodes but the storage is partitioned for iops isolation. Total of 5 nodes for all control plane. @davidewatson do you want to open issues for these specific issues?

countspongebob:
@joejulian @sostheim Can you please comment based on customer requirements you are aware of? Note that customers should not be named in this thread.

venezia:
Since we know etcd has a fixed ceiling in size, I feel 100gb might be way too big?

Mike

davidewatson:
@countspongebob: Currently, k2 provisioned etcd servers must exist within a single k2 nodePool. Allowing multiple etcd instances to share a single node might be substantially easier once we support self-hosted k8s (e.g. #153 and #168), although, that is quite a bit of work, and may not even be acceptable for users large enough to require split etcd.

@venezia: The 256GB number was determined during SIG scale based on the number of IOPS required.

@coffeepac: We'll need to decide how to prioritize this.

[...]

PR #744, opened 2017-08-14
Update etcd versions

leahnp:
Initial work for: https://github.com/samsung-cnct/k2/issues/620

### DIFF
--- a/ansible/roles/kraken.config/files/config.yaml
+++ b/ansible/roles/kraken.config/files/config.yaml
@@ -135,7 +135,7 @@ definitions:
       clusterToken: espouse-monger-rarely
       peerPorts: [2380]
       ssl: true
-      version: v3.1.0
+      version: v3.2.5
     - &defaultEtcdEvents
       name: etcdEvents
       kind: kvStore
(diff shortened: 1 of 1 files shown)


mikeln:
Why not etcd 3.2.x (i.e. 3.2.4 currently)?

leahnp:
I was just following the k8s community cluster recs - but no reason not to get on the latest.

mikeln:
well, I'm not sure of the differences btwn 3.1.x and 3.2.x there may be some reason not to do it...

davidewatson:
According to the [etcd upgrade instructions](https://github.com/coreos/etcd/blob/master/Documentation/upgrades/upgrade_3_2.md) 

> 3.2 introduces two breaking changes.

This noteworthy, because it is a minor version change. Since the breaking changes are in the client and we vendor our golang dependencies, it doesn't affect us.

davidewatson:
According to the [etcd upgrade instructions](https://github.com/coreos/etcd/blob/master/Documentation/upgrades/upgrade_3_2.md) 

> 3.2 introduces two breaking changes.

This noteworthy, because it is a minor version change. However, since the breaking changes are in the client and we vendor our golang dependencies, it doesn't affect us.

countspongebob:
I've opened possibly a dup here

https://github.com/samsung-cnct/kraken/issues/206

@leahnp The reason "not to update to latest" is that we lose community CI alignment, support, and protection. On core kubernetes we should not be running ahead of community on core.

Issue #745, opened 2017-08-14
Turn off e2e ginkgo Log Colorization in CI Runs


mikeln:
Analyzing e2e logs from CI runs is hard with all the extra colorization control codes embedded in the output.   Add the `--nocolor` argument to the CI ginkgo e2e run command, to remove the colorization.

mikeln:
I think we are trying to set it already...
```
common_test_args=()
+ common_test_args+=("--ginkgo.v=true")
+ common_test_args+=("--ginkgo.noColor=true")
+ test_args=()
+ test_args+=("--ginkgo.focus=\[Conformance\]")
+ test_args+=("--ginkgo.skip=\[Serial\]")
+ test_args+=("--e2e-output-dir=${KUBE_CONFORMANCE_OUTPUT_DIR}")
+ test_args+=("--report-dir=${KUBE_CONFORMANCE_OUTPUT_DIR}")
+ cd /mnt/scratch/kubernetes
+ GINKGO_PARALLEL=y
+ go run hack/e2e.go -- --v --test '--test_args=--ginkgo.v=true --ginkgo.noColor=true --ginkgo.focus=\[Conformance\] --ginkgo.skip=\[Serial\] --e2e-output-dir=/home/jenkins/workspace/K2_PR-694-BFT7YJ63M6ACLYQ4C2KP4JAWU44Y53AFYFS2M4SRS2MIVPCWBJOA/output/artifacts --report-dir=/home/jenkins/workspace/K2_PR-694-BFT7YJ63M6ACLYQ4C2KP4JAWU44Y53AFYFS2M4SRS2MIVPCWBJOA/output/artifacts' --check-version-skew=false
```

However by this: https://github.com/kubernetes/community/pull/513/files 

it should be `noColor="y"`

mikeln:
Conflicting docs...looking at the code.
Also, the e2e test doc says it should be `go run hack/e2e.go -- -v --test`  (i.e. one dash on the -v)

mikeln:
trial and error shows that at the arg should just be `--ginkgo.noColor`  alone.  
however, results are still colored...looking for that switch.

mikeln:
ok, the  --ginkgo.noColor arg did not do anything. However, setting the GINKGO_NO_COLOR=y  ENV var did work:

```
[...]

PR #746, opened 2017-08-14
try to remove color

mikeln:
the current args are NOT turning off the color control chars in the e2e logs (though they should).

### DIFF
--- a/hack/parallel-conformance.sh
+++ b/hack/parallel-conformance.sh
@@ -35,7 +35,6 @@ function run_hack_e2e_go() {
 
   common_test_args=()
   common_test_args+=("--ginkgo.v=true")
-  common_test_args+=("--ginkgo.noColor")
   #common_test_args+=("--ginkgo.noColor=y")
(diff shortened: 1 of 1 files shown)


mikeln:
fixes #745 

Env VAR appears to be the only reliable way to turn of the colors in the ginkgo log.

mikeln:
e2e failures have nothing to do with colorization of the log output.

hack/parallel-conformance.sh
```diff
@@ -35,7 +35,7 @@ function run_hack_e2e_go() {
 
   common_test_args=()
   common_test_args+=("--ginkgo.v=true")
-  common_test_args+=("--ginkgo.noColor=true")
+  #common_test_args+=("--ginkgo.noColor=y")  # this didnt work
```
coffeepac: is there are reason to leave this comment in place?

mikeln: Yes, to notate that that argument doesn't work...it is shown in several kubernetes examples.

PR #753, opened 2017-08-15
Enforce taint key and value restrictions (fixes #751).

### LINKED ISSUE #751: Update schema to enforce restrictions on taint keys and values
Our restrictions on taint keys and values are too lax. Currently, the schema is:

```
    "key": {
      "description": "Taint key which must match toleration",
      "type": "string"
    },
    "value": {
      "description": "Taint value",
      "type": "string"
    },
```

However according to @venezia, the k8s documentation states: 

> The key must begin with a letter or number, and may contain letters, numbers, hyphens, dots, and underscores, up to 253 characters.

> The value must begin with a letter or number, and may contain letters, numbers, hyphens, dots, and underscores, up to 63 characters.

The schema should enforce these limits by specifying `pattern`, `maxLength`, and `minLength` for the properties.

davidewatson:


schemas/config/v1/kubeNodeTaint.json
```diff
@@ -8,10 +8,16 @@
...
       "description": "Taint key which must match toleration",
+      "pattern": "[A-Za-z0-9][A-Za-z0-9-._]*",
+      "minLength": 1,
+      "maxLength": 253,
       "type": "string"
     },
     "value": {
       "description": "Taint value",
+      "pattern": "[A-Za-z0-9][A-Za-z0-9-._]*",
+      "minLength": 1,
```
DStorck: taints don't actually need values, they can be empty strings

DStorck: ex: --register-with-taints=node-role.kubernetes.io/master=:NoSchedule

coffeepac:
this PR needs rebasing to pass CI (needs to pick up e2e notifications)

PR #757, opened 2017-08-15
Improve generation of SSH configs etc

### LINKED ISSUE #686: The help.txt template leaves off the last of each nodepool
```
    ssh -F /home/jjulian/.kraken/colonybead/ssh_config master-1
    ssh -F /home/jjulian/.kraken/colonybead/ssh_config master-2
```
for instance should also have
```
    ssh -F /home/jjulian/.kraken/colonybead/ssh_config master-3
```

### LINKED ISSUE #722: ssh_config File Is Missing Nodes
Not sure when this started, but after the etcd split out.

The generated ssh_config file for the cluster is missing entries.

It seems to always have the entries for both etcd clusters (etcd, etcdEvents).
But sometimes has nothing else, sometimes has one, or a few, of the other entries:
master, specialNodes, clusterNodes.

### LINKED ISSUE #766: `kraken.readiness` should check node status
The `kraken.readiness` role counts lines without checking node status according to k8s. This can result in `kraken.readiness` succeeding too soon, which can then cause `kraken.services` to fail unnecessarily.

This is the current code:

```yaml
- name: Get all nodes
  command: >
    {{ kubectl }} --kubeconfig={{ kubeconfig | expanduser }} get nodes --no-headers=true --v 7
  register: job_get_nodes_result
  until: job_get_nodes_result.stdout_lines | length >= needed_nodes | int
  retries: "{{ (( readiness_wait | int ) / retry_interval | int ) | int }}"
  delay: "{{ retry_interval }}"
```

An example:

`kraken.readiness` completes:

> FAILED - RETRYING: Get all nodes (33 retries left).
changed: [localhost]

But not all nodes are ready:

> rodin:k2 dwat$ ks get nodes
NAME                           STATUS     AGE
ip-10-0-101-90.ec2.internal    NotReady   5m
ip-10-0-123-255.ec2.internal   NotReady   5m
ip-10-0-123-99.ec2.internal    NotReady   5m
ip-10-0-134-225.ec2.internal   NotReady   28s
ip-10-0-135-158.ec2.internal   NotReady   1m
ip-10-0-137-94.ec2.internal    NotReady   5m
ip-10-0-16-17.ec2.internal     NotReady   5m
ip-10-0-181-126.ec2.internal   Ready      2m
ip-10-0-2-249.ec2.internal     Ready      5m
ip-10-0-214-51.ec2.internal    NotReady   1m
ip-10-0-230-149.ec2.internal   NotReady   2m
ip-10-0-38-49.ec2.internal     Ready      5m
ip-10-0-43-3.ec2.internal      NotReady   5m
ip-10-0-54-161.ec2.internal    NotReady   5m
ip-10-0-95-106.ec2.internal    NotReady   5m
rodin:k2 dwat$

Until later:

> rodin:k2 dwat$ ks get nodes
NAME                           STATUS    AGE
ip-10-0-101-90.ec2.internal    Ready     7m
ip-10-0-123-255.ec2.internal   Ready     7m
ip-10-0-123-99.ec2.internal    Ready     7m
ip-10-0-134-225.ec2.internal   Ready     2m
ip-10-0-135-158.ec2.internal   Ready     2m
ip-10-0-137-94.ec2.internal    Ready     6m
ip-10-0-16-17.ec2.internal     Ready     7m
ip-10-0-181-126.ec2.internal   Ready     4m
ip-10-0-2-249.ec2.internal     Ready     7m
ip-10-0-214-51.ec2.internal    Ready     3m
ip-10-0-230-149.ec2.internal   Ready     4m
ip-10-0-38-49.ec2.internal     Ready     7m
ip-10-0-43-3.ec2.internal      Ready     7m
ip-10-0-54-161.ec2.internal    Ready     7m
ip-10-0-95-106.ec2.internal    Ready     7m
rodin:k2 dwat$

samba:
There are at least two distinct cases to address:

- `ssh_config` is missing nodes. This appears to be a byproduct of cluster spin-up failure, possibly timeout of long instantiation cycles. This is mitigated by using the `ssh_only` tag for `up.sh`, and regenerating the config once available nodes reach a state of readiness.
- `ssh_config` has incorrect hostnames/IP addresses for various hosts. This _was_ due to the arbitrary application of the host index each time the config was generated. I've mitigated this by interpreting those indexes from the relevant tags in AWS.

Fixes #722.
Fixes #686.

edit from @coffeepac
Fixes #766


ansible/roles/kraken.provider/kraken.provider.aws/tasks/wait-for-etcd.yaml
```diff
@@ -5,7 +5,7 @@
 
 # We purposely use ssh to query etcd over the loopback interface.
 # This allows us to support stricter firewall rules.
-- name: Wait for etcd cluster to form
+- name: Wait for etcd cluster to form ({{nodePool.name}})
```
joejulian: Please use spaces after `{{` and before `}}`

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/main.yaml
```diff
@@ -16,10 +16,10 @@
     filters:
       "tag:k2-nodepool": "{{ item.name }}"
       "tag:Name": "{{ cluster.name }}_{{ item.name }}*"
-      instance-state-name: running
   with_items: "{{ cluster.nodePools }}"
   register: ec2_results
 
+
```
joejulian: extra blank line should be removed

ansible/roles/kraken.ssh/kraken.ssh.aws/templates/ssh_config.jinja2
```diff
@@ -1,19 +1,31 @@
...
 {% for instance in result.instances %}
 {% if instance.state == 'running' or instance.state == 'pending' %}
 {% set privateKeyPath = cluster.nodePools | selectattr("name", "equalto", instance.tags['k2-nodepool']) | list %}
 {% if privateKeyPath[0].keyPair.privatekeyFile is defined %}
-{% if instance.public_dns_name is defined and instance.public_dns_name != "" %}
-Host {{instance.tags['k2-nodepool']}}-{{validHostCounter}}
-  HostName {{instance.public_dns_name}}
+{% set publicAddress = instance.public_dns_name or instance.public_ip_address %}
+{% if publicAddress %}
+
```
joejulian: Between these two and the two blank lines below, that'll make 4 blank lines between entries in the ssh_config file. Can we shrink this down to one blank line between entries please?

ansible/up.yaml
```diff
@@ -7,7 +7,7 @@
       }
     - {
         role: 'roles/kraken.cluster_common',
-        tags: [ 'assembler', 'provider', 'ssh', 'readiness', 'services', 'all', 'dryrun' ]
+        tags: [ 'assembler', 'provider', 'ssh', 'ssh_only', readiness', 'services', 'all', 'dryrun' ]
```
joejulian: What's the difference between `ssh` and `ssh_only`. I was under the impression that `ssh` was used to rebuild the ssh_config file before.

samba: I derived it from the existing `ssh_only` tag that was already on the `kraken.ssh.selector` role. 

These two roles address generation of relevant config + help files. Using the `ssh_only` tag seems to run noticeably faster than using `ssh`, once the cluster is already (at least partially) operational.

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/main.yaml
```diff
@@ -16,10 +16,10 @@
     filters:
       "tag:k2-nodepool": "{{ item.name }}"
       "tag:Name": "{{ cluster.name }}_{{ item.name }}*"
-      instance-state-name: running
```
coffeepac: is it possible to pass a list into this option?  I am thinking if a node was just recycled by an ASG it will appear twice.  once as terminated and once as running or pending.

if this doesn't take a list its a really small window of possible pain.

ansible/up.yaml
coffeepac: what do we need in cluster_common for `ssh_only`?

ansible/roles/kraken.ssh/kraken.ssh.aws/tasks/main.yaml
coffeepac: actually, looks like the 'running or pending' filter gets applied later.  can ignore this if that's the case.

samba: My read of [the docs][1] dictates that `filters` a simple key/value pair. I considered putting a list in there. 

The Jinja template that consumes this property `ec2_results` also filters for the states `running` or `pending` as well. In the case that a node is `pending` but already has public addresses assigned, I would consider that useful for inclusion in `ssh_config`. 

The use of ASGs complicates this detail along the whole lifecycle of the cluster, so making it easy to re-generate the config would be valuable.



[1]: http://docs.ansible.com/ansible/latest/ec2_remote_facts_module.html

ansible/up.yaml
samba: `cluster_common` generates the help text that was missing nodes (off-by-one)

samba:
Found an issue... seems the cluster locking behavior, once a cluster is completely "up", preempts my config re-generation. I'm looking into it and will update here accordingly.

samba:
@joejulian @coffeepac 
Regenerating the ssh_config after the cluster is already up required dodging the `fail` block triggered by the `cluster.status.lock` file. Patched accordingly. 

cc @davidewatson

samba:
Looks like several of you were automatically assigned. Sorry! :)

samba:
@coffeepac @joejulian @mikeln With the last commit, this change has been reliable in generating `ssh_config` with all expected hosts, in my work so far... Anything else required to merge?

ansible/up.yaml
```diff
@@ -23,6 +23,7 @@
       }
     - {
         role: 'roles/kraken.ssh/kraken.ssh.selector',
+        phase: 'one',
```
coffeepac: I don't see this variable being used anywhere.  is it?

samba: In my observation Ansible seemed to skip (seemingly) duplicate tests. Adding this variable to the context seemed to mitigate that, and enforce running that step as intended.

coffeepac: what value do we get from the first run of this role?

samba: I'd defer to others for specifics... my understanding was that a partial `ssh_config` with the etcd nodes present was required in one of the intervening roles. I haven't tried running it without the first iteration of it.

samba: It appears that initial pass on generating `ssh_config` may not actually be necessary, in subsequent tests. Another commit forthcoming to remove it.

ansible/roles/kraken.cluster_common/tasks/main.yaml
```diff
@@ -8,6 +8,7 @@
   loop_control:
     loop_var: a_cluster
   when: a_cluster.providerConfig.type == 'cloudinit'
+        and not cluster_status_lock_file.stat.exists
```
coffeepac: I believe this prevents the section `Fail when cluster.status.lock file exists` from ever being hit.  

I believe the correct course of action is to put the cluster creation check back to how it was and only re-generate the ssh-config file through the update.sh shell script

samba: The `update.sh` script expects that user is actually modifying a nodepool, and it errors if none is specified. 

Updating the `ssh_config` file should be viable without modifying any nodepools, as (for example) the ASGs lifecycle will periodically replace nodes that receive different IP addresses, invalidating previously generated `ssh_config`, even though no other cluster modification is warranted. (Consider updating `ssh_config` more like a "pull" operation, while other updates are more like "push".)

This locking change was primarily intended to support that case, as well as recovering from partially-up clusters, which I'd observed having lock files even so, in a few timeout scenarios.   

On advice from @davidewatson, it seemed the cluster lock itself was no longer necessary, since the main locking concern is around etcd, which has its own, separate lock.  Everything else would be managed by Terraform, Kubernetes, etc, and effectively idempotent. 
@davidewatson feel free to clarify if I've overlooked something.

If there are other components of the cluster that are really sensitive, I'd propose we move the locking constraints more closely to those details in the ansible architecture. 

There are certainly a few other approaches to consider, but I chose this one due to lowest impact (as I understood it).

coffeepac: if the cluster.status.lock file is no longer a useful item then it shouldn't exist at all.  Currently the only thing it can do is confuse a future developer.

this would also be a large change so I would appreciate updating the title of this PR to reflect that you are removing this lock file.

davidewatson: We don't have sufficient testing around `up` to know for sure if this lock is needed. Now that etcd is protected with a different lock, it may or may not still be necessary. Since auditing the code today will not ensure it continues to work tomorrow, I'd recommend leaving it for another PR or three.

Instead of overloading `update` with more functionality, why don't we just create another playbook with the express purpose of regenerating the ssh_config?

Finally, it is important that it be possible to include the `kraken.config` role anywhere as it is almost always necessary. To do this, I suggest checks for the `cluster.status.lock` be moved into a `kraken.pre` role to match `kraken.post`, which currently creates the lock file. Only `up` need include these roles.

samba: @coffeepac  On further consideration, I'm exploring/testing a different approach that alleviates the need to change the locking behavior at all. Will follow up with any changes.

mikeln:
Is this hung up?  looks like it isn't running in CI...does it need a rebase?

samba:
@mikeln thanks for checking on it. looks like it timedout on etcd in a few cases. I've triggered another test run.

coffeepac:
@samba please rebase this PR to pick up a stack of etcd updates.

ansible/roles/kraken.config/tasks/check-clusters-lock.yaml
```diff
@@ -7,8 +7,3 @@
      path: "{{ config_base | expanduser }}/{{ cluster.name }}/cluster.status.lock"
     register: cluster_status_lock_file
```
coffeepac: the above task needs to be immediately before the check.  please cut/paste this task to `ansible/roles/kraken.cluster_common/tasks/cluster-start.yaml` and delete this file.

ansible/roles/kraken.cluster_common/tasks/cluster-start.yaml
```diff
@@ -8,6 +8,18 @@
...
+- name: Check that the cluster.status.lock file exists
+  stat:
+    path: "{{ config_base | expanduser }}/{{ cluster.name }}/cluster.status.lock"
+  register: cluster_status_lock_file
+
+- name: Fail when cluster.status.lock file exists
+  fail:
+    msg: >-
+      Spinning up  {{ cluster.name }} cluster has been canceled due to its already completed.
+  when:  cluster_status_lock_file.stat.exists == True
```
coffeepac: this should only happen when kraken_action == 'up'.  sorry, this was present before and I missed it.  need to add it as a conditional to the 'when'

samba: 👍

Issue #680, opened 2017-07-28
taints should be able to be applied to nodepools


joejulian:
Currently, taints can only be applied to node definitions. These should be able to be applied to nodepools.

DStorck:
@joejulian Should this be in addition to the ability to add them per nodeConfig, or is per nodePool just a better way to handle this ?

joejulian:
It probably should be in addition. I don't think that would be difficult to do. Adding them to the nodePools is likely to be used for organizational needs where nodeConfig is more likely for "physical" constraints.

DStorck:
work is done here: https://github.com/samsung-cnct/k2/pull/758

DStorck:
Update - working towards moving taints and labels to live together in a nodePool config called podScheduling.

DStorck:
fixed all requested changes on PR

PR #758, opened 2017-08-15
option to add taints per nodepool

DStorck:
old method was to apply taints at cluster creation via nodeConfig (essentially type), now taints can be added per nodePool

fixes https://github.com/samsung-cnct/k2/issues/680


ansible/roles/kraken.nodePool/kraken.nodePool.node/templates/v1.6/units.kubelet.part.jinja2
```diff
@@ -43,6 +43,7 @@
       --api_servers=https://apiserver.{{ cluster.name }}.internal \
       --register-node=true \
       {% if node.nodeConfig.taints is defined %}--register-with-taints={% set comma = joiner(",") %}{% for taint in node.nodeConfig.taints  %}{{ comma() }}{{taint.key}}={{taint.value}}:{{taint.effect}}{% endfor %}{% endif %} \
+      {% if node.taintConfig is defined %}--register-with-taints={% set comma = joiner(",") %}{% for taint in node.taintConfig.taints  %}{{ comma() }}{{taint.key}}={{taint.value}}:{{taint.effect}}{% endfor %}{% endif %} \
```
coffeepac: can we specify multiple `--register-with-taints` to the kubelet or do we need to join these two lists?

ansible/roles/kraken.nodePool/kraken.nodePool.node/templates/v1.7/units.kubelet.part.jinja2
```diff
@@ -43,6 +43,7 @@
       --api_servers=https://apiserver.{{ cluster.name }}.internal \
       --register-node=true \
       {% if node.nodeConfig.taints is defined %}--register-with-taints={% set comma = joiner(",") %}{% for taint in node.nodeConfig.taints  %}{{ comma() }}{{taint.key}}={{taint.value}}:{{taint.effect}}{% endfor %}{% endif %} \
+      {% if node.taintConfig is defined %}--register-with-taints={% set comma = joiner(",") %}{% for taint in node.taintConfig.taints  %}{{ comma() }}{{taint.key}}={{taint.value}}:{{taint.effect}}{% endfor %}{% endif %} \
```
coffeepac: same as above, I don't think we can specify this switch twice but I could be wrong!

ansible/roles/kraken.nodePool/kraken.nodePool.node/templates/v1.6/units.kubelet.part.jinja2
DStorck: it has been working just fine, but I'm okay to rework it if we think it is cleaner to just to it once

DStorck:
@davidewatson Thanks for the reminder on definitions.json - that is done. 
@coffeepac Your comment as you know was dealt with and now we have schedulingConfig.

schemas/config/v1/definitions.json
```diff
@@ -113,6 +113,14 @@
       },
       "minItems": 1,
       "type": "array"
+    },
+    "schedulingConfig": {
```
davidewatson: This should be plural.

schemas/config/v1/schedulingConfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "http://judkins.house/apis/k2/v1/schedulingConfig.json",
+  "$$target": "schedulingConfig.json",
+  "title": "Scheduling configuration",
+  "description": "Describes configuration of taints added per nodePool",
+
+  "properties": {
+    "taints": { "$ref": "kubeNodeTaint.json" }
```
davidewatson: Should this be an array of taints?

schemas/config/v1/schedulingConfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "http://judkins.house/apis/k2/v1/schedulingConfig.json",
+  "$$target": "schedulingConfig.json",
+  "title": "Scheduling configuration",
+  "description": "Describes configuration of taints added per nodePool",
+
+  "properties": {
```
davidewatson: Configs objects should have a name and kind to facilitate extensions.

schemas/config/v1/schedulingConfig.json
DStorck: our plan was to have this also be the home of 'labels' in the future

davidewatson: According to your documentation, this is an array:

```
+  schedulingConfig:
 +    - &defaultScheduling
 +      taints:
 +        - key: testKey
 +          value: testValue
 +          effect: NoSchedule
```

schemas/config/v1/schedulingConfig.json
DStorck: i put them in @davidewatson , let me know if that is what you were thinking.

schemas/config/v1/schedulingConfig.json
```diff
@@ -0,0 +1,26 @@
...
+      "description": "Scheduling configuration for taints and labels.",
+      "enum": ["scheduling"],
+      "type": "string"
+    },
+    "taints": {
+      "description": "List of taints to apply to node",
+      "items": {
+        "$ref": "kubeNodeTaint.json"
+      },
+      "type": "array"
```
davidewatson: This line needs to be swapped with line above. `type` belongs in the `taints` stanza.

DStorck: are you sure? putting it a line above would make it part of the `items`

davidewatson: You're right! I misread this CI error: ```An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ValueError: Expecting object: line 26 column 2 (char 727)```

PR #767, opened 2017-08-18
githubnotify is failing in the release job, exclude it.  it isn't use…

coffeepac:
…ful here

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -85,7 +85,9 @@ podTemplate(label: 'k2', containers: [
                     // This keeps the stage view from deleting prior history when the E2E test isn't run
                     if (err) {
                         stage('Test: E2E') {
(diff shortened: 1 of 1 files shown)


leahnp:
LGTM @coffeepac what is the official procedures for ignoring e2e? Do we retry the tests a certain number of times - or wait for the PR owner to say it is good to go?

coffeepac:
@leahnp for e2e failure ignore the reviewer needs to:
- see what the failure is
  - if something basic like 'unable to launch a pod' and there is only one failure we ignore it. There are many other tests that launch a pod
- make a best effort determination if the PR has an impact on the failed tests
- if the PR does likely affect the test, rerun the CI job to see if the failure occurs in the same place on followup runs
   - if same error each time, author needs to fix
   - if different,  unrelated errors, ignore results

If regularly seeing timeouts we may need to alter the cluster config we use for CI.  Give a holler in channel and we can discuss

PR #771, opened 2017-08-18
updating k8s to 1.7.4

alika:
This PR is associated w/the on-call version check for week 8/14 samsung-cnct/issues#38


mikeln:
reran to check e2e.  different failures in e2e so not this change.

PR #791, opened 2017-08-31
Don't recreate cloud-config if it exists

joejulian:
Appending files does so in a random order. This causes a change to the
cloud-config which will cause all the ASG to get recreated. This is
undesired behavior so avoid changing cloud-config.

### DIFF
--- a/ansible/roles/kraken.assembler/tasks/nodes.yaml
+++ b/ansible/roles/kraken.assembler/tasks/nodes.yaml
@@ -23,6 +23,7 @@
     dest_path: "{{ config_base | expanduser }}/{{ cluster.name }}/cloud-config/{{ item.0.name }}.cloud-config.yaml"
   when:
     - ( dest_path | is_file ) == false
+    - item.1.name == "etcd" or item.1.name == "etcdEvvents"
   with_together:
     - "{{ cluster.nodePools }}"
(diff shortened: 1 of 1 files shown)


davidewatson:
Won't this prevent us from upgrading the k8s version? The k8s component manifests are stored in the cloud config...

ansible/roles/kraken.assembler/tasks/nodes.yaml
```diff
@@ -18,7 +18,12 @@
 - name: Generate node cloud init
   template:
     src: node.yaml.jinja2
-    dest: "{{ config_base | expanduser }}/{{ cluster.name }}/cloud-config/{{ item.0.name }}.cloud-config.yaml"
+    dest: "{{ dest_path }}"
+  vars:
+    dest_path: "{{ config_base | expanduser }}/{{ cluster.name }}/cloud-config/{{ item.0.name }}.cloud-config.yaml"
+  when:
+    - ( dest_path | is_file ) == false
+    - item.1.name == "etcd" or item.1.name == "etcdEvvents"
```
coffeepac: this check won't work in case the user doesn't name their etcd clusters in this way. it should be 'item.1.etcdConfig is defined'

PR #796, opened 2017-09-01
Fix logic to exclude etcd nodePools if they exist

joejulian:


davidewatson:
@joejulian: Since https://github.com/samsung-cnct/k2/pull/795 was merged, can we close this?

joejulian:
@coffeepac made noises like he was still interested in having this, too. Not sure where he landed on that.

coffeepac:
I would like to merge this as well.  it allows us to trust that the on-disk files match what's in the cloud.  this should allow for manual triaging to be easier.

however, I can be talked out of it.  its a nice to have for me and we have it so I'd like a reason to not merge it if there is one.

coffeepac:
Imma merge this.  if people want this backed out can do later.

PR #798, opened 2017-09-01
Do not delete namespaces on e2e tests

joejulian:
The framework "After" deferral is failing to find a resource unrelated
to the namespace it's attempting to delete. Let's skip it for now.


coffeepac:
going to close in favor of finding the root cause.

PR #800, opened 2017-09-06
Do not use jinja2 templates in ansible when statements

davidewatson:
(cf. https://github.com/openshift/openshift-ansible/pull/5287).

### DIFF
--- a/ansible/roles/kraken.provider/kraken.provider.gke/tasks/process-cluster.yaml
+++ b/ansible/roles/kraken.provider/kraken.provider.gke/tasks/process-cluster.yaml
@@ -4,8 +4,5 @@
   register: deployment_check
   ignore_errors: yes
 
-set_fact:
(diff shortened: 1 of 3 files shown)


ansible/roles/kraken.provider/kraken.provider.gke/tasks/process-cluster.yaml
```diff
@@ -4,5 +4,8 @@
   register: deployment_check
   ignore_errors: yes
 
+set_fact:
```
joejulian: Can you use `vars:` in the include instead of setting a fact?

davidewatson: Good idea.

davidewatson: Actually, I am not sure how that would work. Instead removed the `set_fact` in favor of concatenating the strings within the test.

PR #805, opened 2017-09-07
Use a while loop instead of depending on systemd (fixes #673).

### LINKED ISSUE #673: etcd/etcdEvents Cluster Creation Errors

davidewatson:


leahnp:
LGTM - do you mind giving me a brief overview of the pros/cons of this approach vs our old systemd one?

ansible/roles/kraken.config/files/config.yaml
```diff
@@ -460,7 +460,7 @@ deployment:
           # 5 nodes recommended for production/high availability clusters. This
           # configuration is required to tolerate a single node failure during
           # node upgrades.
-          count: 1
+          count: 5
```
coffeepac: oh?

davidewatson:
@leahnp:

tl;dr; the bash while loop is simpler (and more reliable) because you do not need to understand as much about systemd.

"Restart=on-failure only applies to process failures (does not apply to failure due to dependency failures)", and "If there's even a small chance that a dependency might fail on start and you care about resiliency, don't use Before/After and instead perform a check on some artifact that the dependency produces" - https://unix.stackexchange.com/a/216254

But, having to create an artifact is more trouble than we want than we'd like if we can avoid it.

PR #809, opened 2017-09-08
Remove prediscovery unit since systemd dependency failures may not be

davidewatson:
restarted.


davidewatson:
"Restart=on-failure only applies to process failures (does not apply to failure due to dependency failures)", and "If there's even a small chance that a dependency might fail on start and you care about resiliency, don't use Before/After and instead perform a check on some artifact that the dependency produces" - https://unix.stackexchange.com/a/216254

davidewatson:
CI runs three through five passed the Cloud stage. Run four also passed e2e.

PR #813, opened 2017-09-08
bump default GKE version to 1.6.9

coffeepac:
really?  on a friday?


joejulian:
Only the e2e failed to start but the cluster test succeeded. #814 passed with the same change. I'm going to merge this.

Issue #817, opened 2017-09-11
Security Group gets reset (sometimes) on cluster update - undesirable


venezia:
It appears that if you add and remove a nodepool, some times (but not always) security groups for nodepools are reset to their original intention through the configuration.

This does not sound bad, but can have a disastrous impact on an existing cluster.  k8s aws integration creates a security group for each load balancer and adds this security group to the whitelist of traffic in security groups that nodes are using.

Once reset, k8s aws integration does not reconcile the missing whitelist entries.  In effect the services are cut off from their corresponding ELB.  New ELBs can be created and linked correctly.

What may be desirable is to tell terraform _not_ to to reset this security group.  This also means communicating to the user that any security group ingress/egress rules will need to be handled manually. :(

joejulian:
This bug only affects AWS.

Until this bug is addressed, it can be worked around by adding a "from 0/0" rule with the allowed ports to the node's security groups.

coffeepac:
first attempt:
- create cluster
- manually add rule to security_group
- run `terraform apply` from cli
- manually added rule removed

coffeepac:
second attempt, using pre-existing cluster (from first attempt):
- manually add rule to security_group
- modify existing nodePool (change node count from 1 -> 2)
- manually added rule removed

coffeepac:
third attempt, using pre-existing cluster (from first attempt):
- manually add rule to security group
- modify running kraken-lib to ignore changes to ingress/egress rules for security group
- modify existing nodePool (change node count from 2 -> 1)
- manually added rule ...

coffeepac:
have a fix for an existing cluster, things to sort out still:
- does this work for a new cluster (ie do the initial rules get created then we effectively hand over control to k8s for long-term maintence
- why do I keep seeing that the route table wants to be updated by a raw `terraform plan` but kraken isn't updating it

coffeepac:
it appears there is a similar situation with the vpc route table.  making the same fix there and testing.

fourth attempt, using pre-existing cluster (from first attempt):
- manually add route to route table
- modify running kraken-lib to ignore changes to routes for vpc route table
- modify existing nodePool (change node count from 1 -> 2)
- manually added route (in progress)

coffeepac:
last check:  destroy/create cluster and ensure that cluster is created correctly.  if this works I'll open a PR and get CI after it.

coffeepac:
[...]

PR #818, opened 2017-09-11
prevent kraken from removing some kubernetes created resources

coffeepac:
kraken creates the vpc route table and the vpc security group , however
kubernetes will make modifications to both assets as the cluster runs.
we should not let kraken modify either of these resources going forward.

there is no way to specify custom routes in the config file so we are fine there
we need to modify the documentation on the security group to specify that
the ingress and egress rules for the security will only be applied at cluster
creation.

Fixes: #817

### DIFF
--- a/ansible/roles/kraken.provider/kraken.provider.aws/templates/kraken.provider.aws.module.cluster_secgroup.tf.jinja2
+++ b/ansible/roles/kraken.provider/kraken.provider.aws/templates/kraken.provider.aws.module.cluster_secgroup.tf.jinja2
@@ -57,7 +57,13 @@ resource "aws_security_group" "vpc_kubernetes_secgroup" {
     Name = "${var.secgroup_name}_vpc_secgroup",
     KubernetesCluster = "${var.secgroup_name}"
   }
-}
+
+{% if kraken_action == 'update' %}
+  lifecycle {
+    ignore_changes = ["ingress", "egress"]
+  }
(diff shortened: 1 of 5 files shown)


ansible/roles/kraken.provider/kraken.provider.aws/templates/kraken.provider.aws.module.cluster_secgroup.tf.jinja2
```diff
@@ -57,7 +57,13 @@ resource "aws_security_group" "vpc_kubernetes_secgroup" {
     Name = "${var.secgroup_name}_vpc_secgroup",
     KubernetesCluster = "${var.secgroup_name}"
   }
-}
+
```
davidewatson: Can we plumb these in through template parameters like we do for etcd? For example, from `ansible/roles/kraken.provider/kraken.provider.aws/tasks/main.yaml`:

```
# Terraform resources which have prevent_destroy set to true will not be
# destroyed, even when `terraform destroy` is called. To avoid the potential
# for infrastructure data loss, some resources are only allowed to be deleted
# when all of the infrastructure is being destroyed by the down action.
- include: aws-template.yaml
  vars:
    prevent_destroy: "{{ 'false' if kraken_action == 'down' else 'true' }}"
```

I worry that the terraform templates are already unwieldy and will continue to get worse the more kraken logic we embed in them.

PR #824, opened 2017-09-12
823 too many expanduser

### LINKED ISSUE #823: Reduce the number of "expanduser" filters to a minimum
There are over 200 lines with the exact same `expanduser` usage. Reduce these to the minimum number required to produce the same results.

joejulian:
By ensuring that `config_base` and `config_file_default` are expanded early, this merge request mostly eliminates the use of the expanduser filter making the code much easier to read.

Fixes #823

### DIFF
--- a/ansible/roles/kraken.config/defaults/main.yaml
+++ b/ansible/roles/kraken.config/defaults/main.yaml
@@ -1,6 +1,6 @@
 ---
 kraken_action: up
-config_base: "~/.kraken"
-config_file_default: "{{ config_base | expanduser }}/config.yaml"
+config_base: '{{ "~/.kraken" | expanduser }}'
+config_file_default: "{{ config_base }}/config.yaml"
 config_file: "{{ config_path | expanduser | default(config_file_default) }}"
 dryrun: false
--- a/ansible/roles/kraken.config/defaults/main.yaml
+++ b/ansible/roles/kraken.config/defaults/main.yaml
@@ -1,6 +1,6 @@
 ---
 kraken_action: up
-config_base: '{{ "~/.kraken" | expanduser }}'
+config_base: "~/.kraken"
 config_file_default: "{{ config_base }}/config.yaml"
-config_file: "{{ config_path | expanduser | default(config_file_default) }}"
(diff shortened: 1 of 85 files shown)


ansible/roles/kraken.config/defaults/main.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 kraken_action: up
-config_base: "~/.kraken"
-config_file_default: "{{ config_base | expanduser }}/config.yaml"
+config_base: '{{ "~/.kraken" | expanduser }}'
```
davidewatson: What about the case where `config_base` is passed into Ansible from the `./up.sh` or `kraken`? I think you need to use the `expanduser` filter from a task list and not the defaults, maybe in `kraken.config` + `/tasks/global-setup.yaml`?

mikeln:
e2e tests do not appear to relate to the change made.  should be ok to merge.

ansible/roles/kraken.config/defaults/main.yaml
joejulian: I handled that in `lib/common.sh` by expanding the path passed in. Do you think that should _also_ be done in ansible for clarity?

davidewatson: Our defaults are scattered and only partially documented. For example, does the default come from `./lib/common.sh`, ansible defaults (e.g. `kraken.*/defaults/main.yaml`), an ansible task list, or jsonschema? I'd prefer if they were all set in one place. I think that means an ansible task list, preferably in `kraken.config`. It's your call.

Interestingly, `realpath` doesn't actually expand tilde:

```
bash-4.3# realpath ~/
/root
bash-4.3# realpath '~/'
realpath: ~/: No such file or directory
bash-4.3# 
```

So your current code is depending on bash.

joejulian: Huh, interesting. It worked in Linux.

I was already planning on removing that and putting the expansion, once, in tasks instead but I'm waiting to rebase it after #826 is merged.

PR #826, opened 2017-09-14
Add ansible linting to Jenkins

### LINKED ISSUE #827: Add ansible linting to Jenkins
Much of another PR I'm working on keeps touching bits that do not match expected formatting which causes me to make changes that are not specific to the feature I'm adding. I fear that this will make my PR less comprehensible and harder to review, thus I'd like to correct all linting errors and add ansible linting to Jenkins.

To that end, I'm splitting off all the syntactic changes into a PR (#826) that is specific to this need. This will also add lint checking to Jenkins to help prevent this issue in the future.

joejulian:
* Fix ansible lint errors
* Add ansible lint checking to Jenkins

Fixes #827

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,12 +16,14 @@ custom_jnlp_version    = "0.1"
 
 jnlp_image             = "quay.io/${quay_org}/custom-jnlp:${custom_jnlp_version}"
 k2_tools_image         = "quay.io/${quay_org}/k2-tools:${k2_tools_image_tag}"
+ansible_lint           = "quay.io/${quay_org}/ansible-lint:latest"
 e2e_tester_image       = "quay.io/${quay_org}/e2etester:${e2etester_version}"
 docker_image           = "docker"
 
 podTemplate(label: 'k2', containers: [
     containerTemplate(name: 'jnlp', image: jnlp_image, args: '${computer.jnlpmac} ${computer.name}'),
     containerTemplate(name: 'k2-tools', image: k2_tools_image, ttyEnabled: true, command: 'cat', alwaysPullImage: true, resourceRequestMemory: '1Gi', resourceLimitMemory: '1Gi'),
+    containerTemplate(name: 'ansible-lint', image: ansible_lint, ttyEnabled: true, command: 'cat', alwaysPullImage: true, resourceRequestMemory: '1Gi', resourceLimitMemory: '1Gi'),
     containerTemplate(name: 'e2e-tester', image: e2e_tester_image, ttyEnabled: true, command: 'cat', alwaysPullImage: true, resourceRequestMemory: '1Gi', resourceLimitMemory: '1Gi'),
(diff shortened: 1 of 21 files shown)


coffeepac:
@joejulian I believe you need to rebase this onto master to get etcd stability fixes and e2e notifications messages.

Jenkinsfile
```diff
@@ -57,7 +59,9 @@ podTemplate(label: 'k2', containers: [
 
             // Unit tests
             stage('Test: Unit') {
-                kubesh 'true' // Add unit test call here
+                customContainer('ansible-lint') {
+                    kubesh 'ansible-lint ansible/*.yaml'
```
coffeepac: is this recursive?

ansible/roles/kraken.access/tasks/generate-admin-cert.yaml
```diff
@@ -10,26 +10,27 @@
...
   when: (cluster.kubeAuth.authz.rbac is defined)
 
-- name: Generate end-user admin csr 
+- name: Generate end-user admin csr
   command: >
     openssl req -new -key {{ config_base | expanduser }}/{{ cluster.name }}/certs/end-user.key
       -subj {{ endsubj }} -out {{ config_base | expanduser }}/{{ cluster.name }}/certs/end-user.csr
-        creates={{ config_base | expanduser }}/{{ cluster.name }}/certs/end-user.csr
+  args:
+    creates: "{{ config_base | expanduser }}/{{ cluster.name }}/certs/end-user.csr"
```
coffeepac: this isn't the only arg here.  it was the only arg without a leading dash and with an equal sign but this is inconsistent within this command.

ansible/roles/kraken.access/tasks/generate-admin-kubeconfig.yaml
```diff
@@ -22,6 +22,12 @@
       --certificate-authority={{ config_base | expanduser }}/{{ cluster.name }}/certs/ca.pem
       --embed-certs=true
       --kubeconfig={{ kubeconfig }}
+  vars:
+    kconfig: "{{ kubeconfig | is_file | ternary(lookup('file',kubeconfig),'---') | from_yaml }}"
```
coffeepac: do you want to use this var in the command? does that work?

joejulian: Nope, this does not work.

joejulian:
Regarding args, I agree it should be consistent. It's already inconsistent throughout the entire project. Fixing that and adding that custom lint rule is outside the scope of this change but I'll create an issue.

Jenkinsfile
joejulian: Yes, specifying *.yaml will recurse down through the entire playbooks.

joejulian:
I would prefer this be a "rebase and merge" to retain the individual commits. I think they're useful history.

PR #829, opened 2017-09-15
added check for existence of config file and added proper error message

### LINKED ISSUE #630: add check to K2 for config file not existing
if the config file passed into K2 via the --config option doesn't exist, K2 exits with a an error about the user using an old-style config file.  this is misleading and wrong.  

add a check to the kraken.config role to make sure the config file actually exists.

guineveresaenger:
Fixes #630.

### DIFF
--- a/ansible/roles/kraken.config/tasks/main.yaml
+++ b/ansible/roles/kraken.config/tasks/main.yaml
@@ -1,4 +1,10 @@
 ---
+- name: Fail if user specified config file does not exist
+  fail:
+    msg: >-
+         The configuration file ({{ config_file }}) cannot be found. Please make sure a file exists and you specified the correct file path.
+  when: not config_file | is_file
+
 - name: "Expand ~ paths"
   set_fact:
     config_base: "{{ config_base | expanduser }}"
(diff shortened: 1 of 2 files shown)


ansible/roles/kraken.config/tasks/validate-config.yaml
```diff
@@ -1,4 +1,14 @@
 ---
+- name: Check if config file exists
```
davidewatson: The bug reported in #630 is occurring because the `kraken.config` role was originally written to allow only defaults to be used if no config were specified. This stopped working when we removed the default cluster name. See `ansible/roles/kraken.config/tasks/main.yaml`:

```
- name: Load default configuration file
  include_vars:
    file: "{{ role_path }}/files/config.yaml"
    name: kraken_config_defaults

- name: Load user specified configuration file
  include_vars:
    file: "{{ config_file }}"
    name: kraken_config_user
  when: config_file | is_file

- name: Expand default configuration
  set_fact:
    kraken_config: "{{ kraken_config_defaults.deployment | expand_config }}"
  when: not config_file | is_file

- name: Non-recursively combine and expand the defaults and user configuration
  set_fact:
    kraken_config: "{{ kraken_config_defaults.deployment | combine(kraken_config_user.deployment) | expand_config }}"
  when: config_file | is_file
```

Instead of adding a stat task in another file, can you follow the pattern above using `is_file` in main.yaml?

ansible/roles/kraken.config/tasks/main.yaml
```diff
@@ -4,6 +4,15 @@
...
     name: kraken_config_defaults
 
+- debug:
+    msg: "{{ config_file }}"
+
+- name: Fail if user specified config file does not exist
+  fail:
+    msg: >-
+         The configuration file ({{ config_file }}) cannot be found. Please make sure a file exists and you specified the correct file path.
+  when: not config_file | is_file
```
davidewatson: After this failure task, the `Expand default configuration` task below becomes unused. Same with the `when: config_file | is_file` checks.

PR #833, opened 2017-09-18
Application default credentials support for GKE provider

maratoid:
Allow for use of [application default credentials](https://developers.google.com/identity/protocols/application-default-credentials) in GKE provider.

Without this patch applied, when running on GKE, talking to various GCP 
services like stackdriver, bigquery, etc. would require manually managing 
.p12 or .json credential keys.


ansible/roles/kraken.provider/kraken.provider.gke/tasks/destroy-cluster.yaml
```diff
@@ -19,4 +23,6 @@
       --zone {{ cluster.providerConfig.zone.primaryZone }}
       --no-async
       --quiet
-  when: cluster_check.stdout == cluster.name
+  when: cluster_check.stdout == "{{ cluster.name }}"
```
davidewatson: "[You] don’t need to use {{ }} to use variables inside conditionals, as these are already implied."

schemas/config/v1/gkeKeyPairConfig.json
```diff
@@ -13,12 +13,17 @@
...
+      "description": "Default application credentials key file.",
+      "type": "string"
     }
   },
 
   "required": [
     "serviceAccount",
-    "serviceAccountKeyFile"
+    "serviceAccountKeyFile",
+    "defaultApplicationCredentials"
```
davidewatson: Did this ever work? We can't add required fields without breaking backwards compatibility.

maratoid: defaultApplicationCredentials should probably be optional. you're right.

Documentation/kraken-configs/keypair/gke.md
```diff
@@ -5,6 +5,7 @@
 | Key Name | Required | Type | Description|
 | --- | --- | --- | --- |
 | username | Optional | String | GCE user name for the privatekeyFile/publickeyFile. I.e. 'core' |
-| serviceAccount | Optional | String | Service account name, i.e serviceaccount@project-name.iam.gserviceaccount.com |
-| serviceAccountKeyFile | Optional | String | Service acount key file. |
+| serviceAccount | Required | String | Service account name, i.e serviceaccount@project-name.iam.gserviceaccount.com |
```
joejulian: It's my understanding that these are not required (and would be a hindrance) if you have the serviceAccountPasswordFile - pkcs12 certificate based authentication.

joejulian:
This PR is still awaiting requested changes. Will you be able to address these @maratoid?

Documentation/kraken-configs/keypair/gke.md
maratoid: So, I guess, either one or the other is required. Do we resolve this through documentation?

maratoid:
I'm otherwise occupied at the moment, but I will be able to get back to this.

joejulian:
Until you're ready to get back to this, I'm adding the "do not merge" tag so I can avoid clicking on it every time I look at open PRs. Feel free to remove it when you get back to it.

PR #836, opened 2017-09-21
[Feature Request] kraken kubectl suppress ansible output by default.

alejandroEsc:
changed output of computed_kubectl to suppress ansible output, optionally allows you to turn it on via verbosity flag.

### DIFF
--- a/bin/max_k8s_version.sh
+++ b/bin/max_k8s_version.sh
@@ -23,7 +23,18 @@ source "${my_dir}/../lib/common.sh"
 trap control_c SIGINT
 LOCAL_KEV="kraken_action=max_k8s_version version_outfile=${OUTFILE}"
 
-ansible-playbook ${K2_VERBOSE} -i ansible/inventory/localhost \
-	ansible/max_k8s_version.yaml --extra-vars \
-	"${KRAKEN_EXTRA_VARS}${LOCAL_KEV}" \
-    || echo "max_version failed"
+function runcmd(){
+    ansible-playbook ${K2_VERBOSE} \
+        -i ansible/inventory/localhost \
+        ansible/max_k8s_version.yaml \
+        --extra-vars "${KRAKEN_EXTRA_VARS}${LOCAL_KEV}" \
+        || echo "max_version failed" > out
+}
+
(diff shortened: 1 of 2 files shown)


bin/max_k8s_version.sh
```diff
@@ -23,7 +23,18 @@ source "${my_dir}/../lib/common.sh"
...
-ansible-playbook ${K2_VERBOSE} -i ansible/inventory/localhost \
-	ansible/max_k8s_version.yaml --extra-vars \
-	"${KRAKEN_EXTRA_VARS}${LOCAL_KEV}" \
-    || echo "max_version failed"
+function runcmd(){
+    ansible-playbook ${K2_VERBOSE} \
+        -i ansible/inventory/localhost \
+        ansible/max_k8s_version.yaml \
+        --extra-vars "${KRAKEN_EXTRA_VARS}${LOCAL_KEV}" \
+        || echo "max_version failed" > out
```
davidewatson: Why are we creating this file?

bin/max_k8s_version.sh
```diff
@@ -23,7 +23,18 @@ source "${my_dir}/../lib/common.sh"
...
+}
+
+if [[ ! -z $K2_VERBOSE ]]; then
+    runcmd
+else
+    out=`runcmd`
```
davidewatson: Maybe I misunderstand, but why not redirect to /dev/null instead of capturing the output in a variable?

alejandroEsc: hmm ok that is a good point! doh, yeah ill update a new pr with that which includes one for ssh.

bin/max_k8s_version.sh
alejandroEsc: not m preference. This is the pattern that kraken follows for getting the version of kubectl that is required based on k8s version. Probably would need to figure out how to get that info from the config file. Seriously we need to parse the schema into go already....

davidewatson: The result will be saved to ${OUTFILE}...

davidewatson: BTW, we may be closer to golang structs with https://github.com/samsung-cnct/jsonsvalidator

Issue #837, opened 2017-09-21
./update.sh --addnodepools causes helm version downgrade


sostheim:
On an long running cluster, the act of adding a node pool to the cluster via the update script, causes the helm version to be downgraded.

Specifically, from within a running cluster, using the kraken-lib (k2:latest) image base on the `krak8s` Pod, running the command:
```
./bin/update.sh --config /root/.kraken/config.yaml --output /root/.kraken/ --addnodepools nexusNodes
```
Causes the following behavior:
```
helm registry install quay.io/samsung_cnct/mongodb-replicaset --namespace test-ns --name test-mongodb --values /tmp/prefix868881021 --version 1.2.0-0
usage: appr [-h]
            {pull,run-server,show,inspect,list,delete-package,helm,version,logout,deploy,plugins,push,login,config,channel,jsonnet}
            ...
appr: error: 
message: 'Error: incompatible versions client[v2.6.1] server[v2.5.1]
```

sostheim:
On the krak8s Pod node:
```
# helm version
Client: &version.Version{SemVer:"v2.6.1", GitCommit:"bbc1f71dc03afc5f00c6ac84b9308f8ecb4f39ac", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.5.1", GitCommit:"7cf31e8d9a026287041bae077b09165be247ae66", GitTreeState:"clean"}
```
Then proceeding with a helm upgrade:
```
/ # helm init --upgrade
Creating /etc/helm/repository 
Creating /etc/helm/repository/cache 
Creating /etc/helm/repository/local 
Creating /etc/helm/starters 
Creating /etc/helm/cache/archive 
Creating /etc/helm/repository/repositories.yaml 
$HELM_HOME has been configured at /etc/helm.

Tiller (the Helm server-side component) has been upgraded to the current version.
Happy Helming!
/ # helm version
Client: &version.Version{SemVer:"v2.6.1", GitCommit:"bbc1f71dc03afc5f00c6ac84b9308f8ecb4f39ac", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.6.1", GitCommit:"bbc1f71dc03afc5f00c6ac84b9308f8ecb4f39ac", GitTreeState:"clean"}
```
After which, the helm install shown in the issue description above succeeds.

sostheim:
This appears to be related to the failures described in https://github.com/samsung-cnct/krak8s/issues/10

[...]

PR #839, opened 2017-09-21
trim down roles being run for update

coffeepac:
we have been running all roles weather needed or not on update to
ensure we don't miss anything.  since the update process was added
we have cleaned up the interfaces between roles significantly and
can now split out exactly which roles are necessary for making
changes to a running kraken-managed kubernetes cluster


sostheim:
related to #837

PR #843, opened 2017-09-21
remove the 'ignore errors' for running the taints task in gke

coffeepac:
ansible's boolean evaluation is 'short circuit-able' so we can test
for the parent object being defined and parsing will stop before
checking for the child.  this will prevent error messages that get
ignored and make the system slightly safer


DStorck:
@coffeepac thanks, I couldn't figure out how to stop that from erroring out. LGTM

PR #846, opened 2017-09-24
Created ssh command shell script to allow ssh from container.

alejandroEsc:
Part of the solution for https://github.com/samsung-cnct/kraken/issues/109


alejandroEsc:
@leahnp sure thing i will have to document it, probably be best with a help option. However its meant to run in the kraken-lib docker container, and a way to allow for using ssh to run a command remotely, so the idea is you would run kraken in a bastion, or your local machine, which ever you prefer.

coffeepac:
@alejandroEsc would it be better to not accept an ssh_config file at all but rather accept a cluster name, optionally accept an output location (for nonstandard kraken output locations) and then we can find the ssh_config file?  this way we're encapsulating the ssh_config work and we can accept the short name as specified in the ssh_config file to determine which node to ssh to.  

that might be a bit much but I think would be well in-line with letting kraken control infra and then asking kraken to help connect to that infra.

alejandroEsc:
@coffeepac  Hmm, thats one option, my concern there is that users may want to share the ssh_config file, but yeah true, maybe thats not expected behavior, or they should reproduce the directory structure.

coffeepac:
I was thinking if users are sharing the config they  probably won't be using the ssh bundled with kraken.  That user group probably knows what they are doing and don't need a helper script.

alejandroEsc:
@coffeepac so after thinking about this a bit more, i think the best thing is not to make assumptions at this level about where the ssh_config is and probably let it happen a level above, mostly because it would be difficult to pick out where the config file exists in order to find that ssh_config file.

alejandroEsc:
@coffeepac oh sorry about that. I mean kraken (cli) will make that determination, based on input config file etc..

alejandroEsc:
Note: bridge code, for kraken (CLI)

PR #851, opened 2017-09-26
Adjusting config.yaml to remove unnecessary mount

venezia:
on etcd nodes, docker is not used.  This removes a 100GB EBS volume that serves only to increase someone's aws bill.

Would like it if kraken could support `io1` disk type (needs additional option for iops setting)


ansible/roles/kraken.config/files/config.yaml
```diff
@@ -195,13 +195,8 @@ definitions:
     - &defaultAwsEtcdNode
       name: defaultAwsEtcdNode
       kind: node
-      mounts:
```
joejulian: This leaves broken yaml

venezia:
retest

Issue #778, opened 2017-08-23
Remove --api_servers from kubelet unit file


venezia:
Looking at kubelet service logs:

```
Aug 22 13:18:57 ip-10-0-30-119.us-west-2.compute.internal rkt[1165]: Flag --api-servers has been deprecated, Use --kubeconfig instead. Will be removed in a future version.
```

Looking further, we provide a kubeconfig file with the API server endpoint listed, so we should be safe to remove this argument.

This is used in a bunch of files - https://github.com/samsung-cnct/k2/search?utf8=%E2%9C%93&q=--api_servers&type=

gochist:
I think we can close this issue. It was already resolved  by #852

coffeepac:
@gochist I agree.  closing.

PR #852, opened 2017-09-26
--api-servers is not needed …

venezia:
… if kubeconfig is provided and indicates api-server, this is deprecated and breaks in v1.8

This answers #778


venezia:
Sure, but I would do that in a different PR -- this is to correct a blocking v1.8 issue

venezia:
retest

coffeepac:
Not sure what the problem is but this PR doesn't seem to work in CI.  The failure symptom is that all nodes are not joining the cluster.  This occurring before the fabric is setup so we are expecting all nodes to show up in a `kubectl get nodes` but they should be 'NotReady'.

coffeepac:
added a patch to the kubelet units so they work.  rebased to resolve the conflict.

joejulian:
I believe this was also done in the 1.8 PR. Please confirm and rebase or close.

davidewatson:
The only difference is this PR passes `--require-kubeconfig=true` whereas HEAD passes `--require-kubeconfig`.

coffeepac:
Will close.  This was required for 1.8 but we seem to have done that already.

PR #854, opened 2017-09-28
[Refactor] computedkubectl and ssh_command

alejandroEsc:
improved handling of flags, and easier code to read, in general creating better handling or arguments that are being passed. Also included recommendations from previous PR.  Changes in kraken as aresult to be referenced here. 

Note this doesnt change tools, just improves usage and deterministic behavior.

### DIFF
--- a/bin/computed_kubectl.sh
+++ b/bin/computed_kubectl.sh
@@ -8,18 +8,49 @@ set -o nounset
 set -o pipefail
 
 my_dir=$(dirname "${BASH_SOURCE}")
+source "${my_dir}/../lib/common.sh"
 
 VERSIONFILE=$(mktemp /tmp/computed_kubectl.XXXXXX)
-rm ${VERSIONFILE}
+
+while [[ $# -gt 0 ]]
+do
+key="$1"
+
+case $key in
+    -c|--config)
+    KRAKEN_CONFIG="$2"
+    shift 2
+    ;;
+    -v|--verbose)
+    VERBOSE=true
+    shift 1
+    ;;
+    *)    # unknown option
+    POSITIONAL+=("$1") # save it in an array for later
+    shift # past argument
+    ;;
+esac
+done
+
+if [ ${#POSITIONAL[@]} -eq 0 ]; then
+    POSITIONAL=("")
+fi
+
+set -- "${POSITIONAL[@]}" # restore positional parameters
+
+if [ -z ${KRAKEN_CONFIG+x} ]; then
+    error "please enter valid config file"
+    exit 1
+fi
 
 trap '{ rm -f -- "${VERSIONFILE}"; }' INT TERM HUP EXIT
+
 # Call separate script to hide our args from lib/common.sh
-if [[ -z ${2+x} ]]; then
-    ${my_dir}/max_k8s_version.sh ${VERSIONFILE} --config $1
+if [[ "${VERBOSE}" == true  ]]; then
+    ${my_dir}/max_k8s_version.sh ${VERSIONFILE} --config ${KRAKEN_CONFIG} --verbose "-vvv"
 else
-    ${my_dir}/max_k8s_version.sh ${VERSIONFILE} --config $1 --verbose $2
+    ${my_dir}/max_k8s_version.sh ${VERSIONFILE} --config ${KRAKEN_CONFIG}
 fi
-shift
 
 K8S_VERSION=$(cat ${VERSIONFILE} | cut -d . -f 1-2)
 rm ${VERSIONFILE}
--- a/bin/max_k8s_version.sh
+++ b/bin/max_k8s_version.sh
@@ -10,31 +10,33 @@ set -o pipefail
 # pull in utils
 my_dir=$(dirname "${BASH_SOURCE}")
 
+# first argument is the output file, shift to next argument when this is read.
+# fail if not found.
 OUTFILE=$1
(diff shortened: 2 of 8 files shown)


bin/computed_kubectl.sh
```diff
@@ -12,14 +12,51 @@ my_dir=$(dirname "${BASH_SOURCE}")
 VERSIONFILE=$(mktemp /tmp/computed_kubectl.XXXXXX)
 rm ${VERSIONFILE}
```
joejulian: There is already an error function in `lib/common.sh`. Let's just use that one.

bin/computed_kubectl.sh
```diff
@@ -12,14 +12,51 @@ my_dir=$(dirname "${BASH_SOURCE}")
...
+    error "please enter valid config file"
+    exit 1
+fi
+
 trap '{ rm -f -- "${VERSIONFILE}"; }' INT TERM HUP EXIT
+
 # Call separate script to hide our args from lib/common.sh
-if [[ -z ${2+x} ]]; then
-    ${my_dir}/max_k8s_version.sh ${VERSIONFILE} --config $1
+if [[ "${VERBOSE}" == true  ]]; then
```
joejulian: Can we move this verbose checking to common.sh as well?

bin/computed_kubectl.sh
```diff
@@ -12,14 +12,51 @@ my_dir=$(dirname "${BASH_SOURCE}")
...
+esac
+done
+
+if [ ${#POSITIONAL[@]} -eq 0 ]; then
+    POSITIONAL=("")
+fi
+
+set -- "${POSITIONAL[@]}" # restore positional parameters
+
+if [ -z ${KRAKEN_CONFIG+x} ]; then
```
joejulian: Can we move this config file check to a function in common.sh to simplify code maintenance?

bin/ssh_command.sh
```diff
@@ -1,116 +1,108 @@
 #!/usr/bin/env bash
 
 USE_HOST_NAME=true
+VERBOSE=false
+HELP=false
+POSITIONAL=()
+PORT=22
+SSH_OPTIONS="-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
 
 function error {
```
joejulian: common.sh

bin/ssh_command.sh
```diff
@@ -1,116 +1,108 @@
...
-  echo -e "\033[0;31mERROR: $1\033[0m"
+    echo -e "\033[0;31mERROR: $1\033[0m"
 }
 
 function inf {
```
joejulian: common.sh

bin/computed_kubectl.sh
alejandroEsc: @joejulian The problem with doing that is that common.sh parses flags (maybe not the best way) that I am not interested in. Im not sure what the behavior would be entirely and I do not want to deal with collisions, specially for one minor change. If anything I would remove the error function in that case. Would you be ok with that?

bin/computed_kubectl.sh
alejandroEsc: Sure that can be done, will it break any expectations?

alejandroEsc:
@joejulian thanks for checking, let me know what you think, will hopefully get some of the changes in tomorrow. Could you also look at the related PR: https://github.com/samsung-cnct/kraken/pull/178

bin/computed_kubectl.sh
joejulian: common parses flags?!?! <smh>

alejandroEsc: oh sorry i mean arguments, yeah it does. It can be cleaned btw, but ill leave that to another PR.

bin/max_k8s_version.sh
```diff
@@ -10,31 +10,33 @@ set -o pipefail
...
+
+source "${my_dir}/../lib/kraken_arguments.sh"
 
 # setup a sigint trap
 trap control_c SIGINT
 LOCAL_KEV="kraken_action=max_k8s_version version_outfile=${OUTFILE}"
 
 function runcmd(){
-    ansible-playbook ${K2_VERBOSE} \
+    echo "$(ansible-playbook ${K2_VERBOSE} \
```
joejulian: Was this for debugging?

alejandroEsc: No its not. Just a way to get the command to run and display output.

bin/computed_kubectl.sh
coffeepac: nope.  early exits are good exits.

bin/computed_kubectl.sh
```diff
@@ -8,18 +8,49 @@ set -o nounset
...
+    error "please enter valid config file"
```
coffeepac: this message needs to be more descriptive about what is not valid.  I think what's not valid is 'config file doesn't exist'.  any bit of 'non-valid' otherwise will be coming from the config file validator.

alejandroEsc: The error should occur when config flag isnt pass, so its really is that you need to enter a config file, and not that it doesnt exist.

coffeepac: it can be more explicit.  'config file not specified, please specify with `--config <config file>`' or something similar.

coffeepac:
merging because this is causing kraken to break.

PR #855, opened 2017-09-28
Clean up templates

joejulian:
There is a lot of cruft and duplication in the template structure. 
* set the default templates to the latest version
* remove 1.4 support
* delete any files that are identical to the default

This does not affect usage and no documentation changes are necessary.


coffeepac:
1. need to resolve those conflicts
2. looks fine
3. are you dropping 1.5 support in this PR?  We need to do that but maybe that should wait until we support 1.8 (#856)
4. I will need to spin up all supported version clusters to make sure they still work

coffeepac:
@joejulian you probably want to rebase this against master.  it looks like its got an old version of GKE for its default (1.6.9) master is currently 1.7.5

joejulian:
Odd, I thought I did branch from master. <shrug>

joejulian:
@coffeepac 
1. bad branch - fixed
2. thanks
3. No, the only 1.5 files that were removed were those that did not differ from the default. We only _need_ a versioned template if it differs from the default. In the case where all the versioned files matched the default, the directory is removed because empty directories are not maintained by git.
4. Try not to have too much fun with that.

coffeepac:
manually tested with kubernetes version 1.5.8, success.

coffeepac:
manually tested with kubernetes version 1.6.11, success.

PR #864, opened 2017-10-02
helm appr pathing sorted

coffeepac:
this change allows the app registry plugin to execute the epxected
version of helm when it runs.  the app registry plugin is called
from helm via `helm appr` and it will make a call back to helm via
a shell invocation `helm`.  This means app registry will execute
whichever helm it finds in the system path.  We don't want this.

we now put the path for the calling helm at the front of the system
path so it gets called by the app registry plugin


coffeepac:
tested with 1.6.11 and the cluster came up and heapster (helm installed) was present.  testing a 1.5.8 cluster now.

will let CI be the judge of 1.7 clusters

guineveresaenger:
hurray!

coffeepac:
successful cluster creation using kubernetes 1.5.8.  Once CI passes this is g2g (last failure CI: #2 was due to resource exhaustion in AWS, I have corrected that)

PR #867, opened 2017-10-03
Update kubedns to fix CVE-2017-14491.

davidewatson:


ansible/roles/kraken.services/defaults/main.yaml
```diff
@@ -39,16 +39,16 @@ kubedns:
   readiness_delay: 3
   readiness_timeout: 5
   dnsmasq_image: "gcr.io/google_containers/kube-dnsmasq-amd64:1.4.1"
-  dnsmasq_image_v16: "gcr.io/google_containers/k8s-dns-dnsmasq-nanny-amd64:1.14.2"
-  dnsmasq_image_v15: "gcr.io/google_containers/kube-dnsmasq-amd64:1.4.1"
+  dnsmasq_image_v16: "gcr.io/google_containers/k8s-dns-dnsmasq-nanny-amd64:1.14.5"
+  dnsmasq_image_v15: "gcr.io/google_containers/kube-dnsmasq-amd64:1.14.5"
   dnsmasq_image_v14: "gcr.io/google_containers/kube-dnsmasq-amd64:1.4.1"
```
joejulian: While you're in here, can you delete v14 stuff?

davidewatson: I created a separate PR because I don't want to slow down this fix and there are probably other things which must happen to fully remove k8s 1.4 support.

Issue #828, opened 2017-09-14
Add `--record` To All Kubernetes Deployments


mikeln:
Cluster creation creates a number of deployments via kubectl.
We should have the `--record` argument set so the history of the deployment contains useful  information:
```
mikeln-bender:jaeger-kubernetes Mikel_Nelson$ kubectl --kubeconfig=/Users/Mikel_Nelson/dev/cloud/k2new/myconfig/.kraken/mikeln-v1709m/admin.kubeconfig rollout history deploy/jaeger-deployment
deployments "jaeger-deployment"
REVISION	CHANGE-CAUSE
1		kubectl create --kubeconfig=/Users/Mikel_Nelson/dev/cloud/k2new/myconfig/.kraken/mikeln-v1709m/admin.kubeconfig --filename=https://raw.githubusercontent.com/jaegertracing/jaeger-kubernetes/master/all-in-one/jaeger-all-in-one-template.yml --record=true
```
versus not very useful info (no --record):
```
mikeln-bender:jaeger-kubernetes Mikel_Nelson$ kubectl --kubeconfig=/Users/Mikel_Nelson/dev/cloud/k2new/myconfig/.kraken/mikeln-v1709m/admin.kubeconfig rollout history deploy/kube-dns --namespace=kube-system
deployments "kube-dns"
REVISION	CHANGE-CAUSE
1		<none>
```

`kubectl create --record ...`
`kubectl apply --record ...`

guineveresaenger:
Pr: #872

PR #872, opened 2017-10-05
added --record to all kubectl apply calls in krakenlib

guineveresaenger:
Also cleaned up unnecessary task in update-cluster.yaml.


guineveresaenger:
Addresses #828

guineveresaenger:
No `create` calls, as we've been trying to make every action idempotent.
I actually searched for "kubectl" rather than "create" since that made more sense, but then went through and did a check on "create" also. Didn't find any.

PR #893, opened 2017-10-17
Changed max len of cluster name to not hit max len for AWS ELB name 

leahnp:
Change max length for cluster names to 21. AWS ELB name max length is 32 and we append `-master-elb` which add 11 chars. 

Fixes: https://github.com/samsung-cnct/kraken-lib/issues/883

### DIFF
--- a/schemas/config/v1/cluster.json
+++ b/schemas/config/v1/cluster.json
@@ -10,7 +10,7 @@
       "description": "Name of this cluster.",
       "type": "string",
       "pattern": "^[A-Za-z0-9-]+$",
-      "maxLength": 32,
+      "maxLength": 21,
       "minLength": 1
     },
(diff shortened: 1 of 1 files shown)


leahnp:
@coffeepac should we be handling any backwards compatibility for this change?

coffeepac:
oh hey, good point, re: backwards compat.  this change will mean we can't tear down any pre-existing clusters cleanly as terraform will be looking for the old name.  

might be easier to leave the name suffix alone and set the cluster name length to be 22  :/  Unless you have an idea for how to handle this transition?

leahnp:
I think that is the cleanest approach. I will amend this. @coffeepac

leahnp:
@coffeepac this passed CI

PR #894, opened 2017-10-17
Use cert specifically for kube-proxy

joejulian:
kubelet and kube-proxy were sharing the same kubeconfig. Using the
correct username (CN) and group (O) in the node cert breaks kube-proxy.
This creates a cert specifically for kube-proxy.

This PR requires changes in drunkensmee (The referenced container tag is
the github commit sha).

### DIFF
--- a/ansible/roles/kraken.config/defaults/main.yaml
+++ b/ansible/roles/kraken.config/defaults/main.yaml
@@ -4,3 +4,4 @@ config_base: "~/.kraken"
 config_file_default: "{{ config_base }}/config.yaml"
 config_file: "{{ config_path | default(config_file_default) }}"
 dryrun: false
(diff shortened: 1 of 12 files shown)


ansible/roles/kraken.config/defaults/main.yaml
```diff
@@ -4,3 +4,4 @@ config_base: "~/.kraken"
 config_file_default: "{{ config_base }}/config.yaml"
 config_file: "{{ config_path | default(config_file_default) }}"
 dryrun: false
+drunkensmee: "quay.io/samsung_cnct/drunkensmee-container:e0e0c05"
```
guineveresaenger: Is kraken.config/defaults just a place to set global variables? How does it work? Can we be assured that everything will work on `cluster down` (i.e. will we need drunkensmee on `cluster down?`)

PR #897, opened 2017-10-19
anonymize docs

leahnp:
I didn't know if this should be `~`, `$HOME` or is there another convention we should use here?


philoserf:
$HOME is best as there are conditions when ~ isn’t expanded. Not critical here. Just a good practice.

PR #905, opened 2017-10-23
 changed defaults

guineveresaenger:
Added dns_only to deafults file

### DIFF
--- a/ansible/roles/kraken.config/defaults/main.yaml
+++ b/ansible/roles/kraken.config/defaults/main.yaml
@@ -5,3 +5,4 @@ config_file_default: "{{ config_base }}/config.yaml"
 config_file: "{{ config_path | default(config_file_default) }}"
 dryrun: false
 drunkensmee: "quay.io/samsung_cnct/drunkensmee-container:e0e0c05"
+dns_only: false


guineveresaenger:
Addresses https://github.com/samsung-cnct/kraken-lib/issues/904

alejandroEsc:
@guineveresaenger can you also make the changes to the kraken_arguments.sh file:

```# pull in common utils
my_dir=$(dirname "${BASH_SOURCE}")
source "${my_dir}/common.sh"

DNS_ONLY=false
KRAKEN_DRYRUN=false```

i would like to see those defaults there always, and have the switch to true only when conditions are met. @coffeepac i dont think the case for nil on these variables really exists, so I dont expect this to be a tri-valued problem.

guineveresaenger:
Hmmm, I do see your point @alejandroEsc

alejandroEsc:
@guineveresaenger let's ask @joejulian then because the defaults in the bash to me take precedence of the defaults defined in the main.yaml file. I would like those removed if possible as the script should be the source of that truth.

coffeepac:
@alejandroEsc I don't see a need to change how the variables are defaulted in the shell script.  right now, if you use the shell script to invoke the playbook things will work as expected.  the problem is that kraken isn't using the shell script. 

so, we need to make invoking the ansible directly work.

alejandroEsc:
@coffeepac that is correct. However there is a ticket up about how kraken needs to be consistent about making calls to kraken-lib, which may mean call the scripts directly or keep just making calls to ansible. Depending on the choice we make we may need to make sure these vars are passed consistently or keep status quo. Either way, the issue still stands, as @guineveresaenger pointed out we have too many places were these things are happening.

joejulian:
Right, since kraken bypasses the scripts the defaults need to be in kraken-lib's ansible defaults.

coffeepac:
kraken doesn't always bypass the scripts, this is why the issue @alejandroEsc references exists.  my preference is we stop invoking ansible directly, but that's out of scope for this issue.

at the moment, if you use the script the variables are defaulted but not in the manner you suggest.  its a style preference.

coffeepac:
anyway, any objections to me merging this as-is?

guineveresaenger:
The test would currently object...

alejandroEsc:
@coffeepac no objections, let's capture your comments about invoking ansible directly on the kraken ticket, ill post later. When the PR passes CI ill go ahead and merge it. Just wanted us all to be on the same page here.

PR #906, opened 2017-10-23
Fixes to json schemas, and correct node validation

oneilcin:
This was all tested using the ansible jsonschema in our kraken.config ansible role.

### DIFF
--- a/schemas/config/v1/cluster.json
+++ b/schemas/config/v1/cluster.json
@@ -32,6 +32,8 @@
       "description": "Provider (e.g. AWS, GKE, etc.) configuration.",
       "$ref": "providerConfig.json"
     },
+    "helmOverride": {
+    },
     "nodePools": {
       "description": "A list of nodePools.",
       "items": {
@@ -74,6 +76,6 @@
     "helmConfig",
     "dnsConfig"
   ],
-
+  "additionalProperties": false,
   "type": "object"
 }
--- a/schemas/config/v1/cluster.json
+++ b/schemas/config/v1/cluster.json
@@ -33,6 +33,8 @@
       "$ref": "providerConfig.json"
     },
     "helmOverride": {
+      "description": "Set to true when using your own helm version.",
+      "type": ["string","null"]
     },
     "nodePools": {
       "description": "A list of nodePools.",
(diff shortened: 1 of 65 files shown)


schemas/config/v1/cluster.json
```diff
@@ -32,6 +32,8 @@
       "description": "Provider (e.g. AWS, GKE, etc.) configuration.",
       "$ref": "providerConfig.json"
     },
+    "helmOverride": {
```
davidewatson: Please give this a description and type.

schemas/config/v1/kubeConfig.json
```diff
@@ -22,7 +22,7 @@
     },
     "version": {
       "description": "Version of the hyperkube binary.",
-      "format": "symver",
+      "format": "semver",
```
davidewatson: You also need update the format name here: https://github.com/samsung-cnct/kraken-lib/blob/master/ansible/library/jsonschema.py#L65

schemas/config/v1/nodeConfig.json
```diff
@@ -4,11 +4,10 @@
...
   "description": "Describes the configuration of a node within a nodePool.",
-
-  "items": {
-    "anyOf": [
-      { "$ref": "awsNodeConfig.json" },
-      { "$ref": "gkeNodeConfig.json" }
-    ]
-  }
+  
+  "anyOf": [
```
davidewatson: Can we make this `oneOf`?

schemas/config/v1/cluster.json
oneilcin: yes

schemas/config/v1/nodeConfig.json
oneilcin: That works

oneilcin:
@davidewatson Requested changes made, thanks!

schemas/config/v1/cluster.json
```diff
@@ -33,6 +33,8 @@
       "$ref": "providerConfig.json"
     },
     "helmOverride": {
+      "description": "Set to true when using your own helm version.",
+      "type": ["string","null"]
```
davidewatson: This should be `["boolean", "string", "null"]` since there are [22 different](http://yaml.org/type/bool.html) strings which are converted to booleans by YAML parsers.

PR #908, opened 2017-10-24
Revert "Fixes to json schemas, and correct node validation"

davidewatson:
Reverts samsung-cnct/kraken-lib#906 so that we can debug additional failures not caught by CI.


alejandroEsc:
lgtm

PR #912, opened 2017-10-24
Print exceptions and explanations even when not in verbose mode.

davidewatson:


ansible/roles/kraken.config/tasks/validate-config.yaml
```diff
@@ -60,12 +67,18 @@
   when:
     - config_forced == 'false'
 
+- name: Print compatibility errors
+  debug:
+    msg: "{{ compatibility.explainations }}"
```
guineveresaenger: typo - can we have it be "explanations"?

davidewatson: Fixed. Thanks!

PR #915, opened 2017-10-26
More reliable creation of storageclass

guineveresaenger:
Addresses https://github.com/samsung-cnct/kraken-lib/issues/898


ansible/roles/kraken.services/tasks/create-storage-class.yaml
```diff
@@ -26,6 +26,6 @@
     {{ kubectl }} --kubeconfig={{ kubeconfig }} apply -f "{{ item }}" --record
   with_first_found:
     - files:
```
joejulian: We should have the version-specific entry in here, too, in case we later need to override.

PR #919, opened 2017-10-27
Tiller namespace

DStorck:
this will install till into a `--tiller-namespace` of `kraken`. Currently it is setup to run tiller as a cluster-admin, giving it permission over all namespaces.  We can change this if folks want, to instead give it access only over `kraken`, `kube-networking`, `kube-system`, and any other namespaces listed in `helmConfigs`.  We could also open another ticket if we think this is good enough for now. 

I would love feedback on if anyone thinks it is necessary to change the info at cluster creation given to a user who runs `dockerdev`, seeing as right now the old: `to use helm`... isn't going to give them access to the releases.  They would need to be instructed to add `--tiler-namespace kraken` to their helm command.

### DIFF
--- a/ansible/roles/kraken.services/files/helm-rbac-config.yaml
+++ b/ansible/roles/kraken.services/files/helm-rbac-config.yaml
@@ -15,4 +15,4 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: tiller
-    namespace: kraken 
+    namespace: kraken
--- a/ansible/roles/kraken.services/files/helm-rbac-config.yaml
+++ b/ansible/roles/kraken.services/files/helm-rbac-config.yaml
@@ -1,0 +1,18 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: tiller
+  namespace: kraken
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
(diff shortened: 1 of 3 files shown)


coffeepac:
what is the minimum version of helm that supports this flag? I'm having trouble finding when this feature was landed.  it looks like it might be helm v2.6 which is problematic as we are using helm v2.5 for kubernetes 1.6 clusters.

DStorck:
bah, @coffeepac 2.6.0

DStorck:
I'm looking into moving out default version of helm for 1.6 clusters to 2.6.2. This is much cleaner than trying to make this `--tiller-namespace` change backwards compatible, and will help with image size.

DStorck:
@coffeepac This actually seems like it's working with the default version of helm for 1.6 clusters.  --tiller-namespace is available, so I don't see a reason to change it specifically for this issue.  If you'd like , I can open up another ticket to bump the version just for simplification.

coffeepac:
Nope.  If it works as is we will let the normal backlog prioritization handle getting the helm version guff removed.

coffeepac:
And unless there is something else this is ready for merge.

joejulian:
Support to allow tiller to be installed into a particular namespace was added in PR https://github.com/kubernetes/helm/pull/1691 . Oldest tag containing that commit is v2.2.0

Issue #909, opened 2017-10-24
Clean-up of 1.5 references from kraken-lib


guineveresaenger:
There are still references in the docs, as well as the default config file, to k8s version 1.5, which is no longer supported.

They should be cleaned up. Also, investigate whether current default settings in config file still make sense for >1.5, which may render some ansible tasts unnecessary.

philoserf:
from `kraken generate`d `$HOME/.kraken/config.yml` this snippet remains:

```yaml
   - &rbacKubeAuth
      authz:
        rbac:
          # super_user is required until kubernetes 1.5 is no longer supported by k2.
          # It is not used by kubernetes 1.6 or later.
          super_user: "placeholder"
```

philoserf:
The comment to be removed is located in the files included in this old/unused PR #922

oneilcin:
PR #934

PR #922, opened 2017-10-27
Remove 1.5 related default config, super_user

philoserf:
- addresses #909

### DIFF
--- a/Documentation/security/README.md
+++ b/Documentation/security/README.md
@@ -40,11 +40,6 @@ definitions:
       rbac:
         super_user: admin
    - &rbacKubeAuth
-      authz:
-        rbac:
-         # super_user is required until kubernetes 1.5 is no longer supported by k2.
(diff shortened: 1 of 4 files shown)


guineveresaenger:
We refer to kubeAuth.authz throughout krakenlib. This issue may be a somewhat more involved change than just removing a config file bit.

I'm not sure what the correct approach is here - perhaps we do still need that stanza as-is (in which case we should remove the confusing comment) or perhaps we need to clean this up throughout kraken-lib. @joejulian might have some ideas?

philoserf:
I'll dig in on this. Generically I see a few outcomes right now.

- we keep this config and remove the deprecation comment
- we replace the config with an appropriate placeholder value
- we update the tests to reflect the the current situation

or best of all words

- we find and remove code no longer needed with k8s 1.5 support retirement

coffeepac:
@philoserf to close PR and open a tracking issue (or link to the existing issue)

PR #928, opened 2017-11-02
revert to community standard etcd version

coffeepac:
fixes samsung-cnct/kraken#206

we should be following the community guidelines for what version of
etcd we use unless we have a compelling reason

### DIFF
--- a/ansible/roles/kraken.config/files/config-private-topology.yaml
+++ b/ansible/roles/kraken.config/files/config-private-topology.yaml
@@ -134,7 +134,7 @@ definitions:
       clusterToken: espouse-monger-rarely
       peerPorts: [2380]
       ssl: true
-      version: v3.1.0
+      version: v3.0.17
     - &defaultEtcdEvents
       name: etcdEvents
       kind: kvStore
@@ -143,7 +143,7 @@ definitions:
       clusterToken: animism-training-chastity
       peerPorts: [2382]
       ssl: true
-      version: v3.1.0
+      version: v3.0.17
 
   apiServerConfigs:
     - &defaultApiServer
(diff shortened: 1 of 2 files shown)


alejandroEsc:
@coffeepac since we will be running three different versions of k8s with presumably different blessed versions of etcd. I think we should consider something like 

```
kraken generate --kubernetes_version 1.8.1
````

and have one config file per version of kubernetes.

venezia:
I'm confused.  https://github.com/kubernetes/community/blob/master/sig-scalability/provider-configs.md#additional-configuration-requirements states 3.1.8 is the minimum.

It also does not state a maximum either.

kube-aws uses 3.2.9

alejandroEsc:
@coffeepac alright i can see that, so we will need some validation. Sounds good.

coffeepac:
@venezia apparently the docs are out of date: https://github.com/kubernetes/community/issues/1321  the recommendtion is based on 'what has passed the 5k node test'.

coffeepac:
[PR that moved us to 3.1.0](https://github.com/samsung-cnct/kraken-lib/pull/208) There doesn't seem to be any reason listed at the time to use 3.1 instead of simply a newer version of 3.0.  

There was a known problem with 3.0.3 that we were upgrading away from.

joejulian:
The initial issue which that PR addressed stated:
> upgrading etcd to at least 3.1 fixes the bug

coffeepac:
removed the 'DO NOT MERGE' label because I found the PR that moved us to 3.1.0 and from 3.1.0 to 3.2.5.  Both moves were fine but there was no specific reason to move to the exact version we chose.  

I recommend we merge this when the reviewers are comfortable with the e2e results.

oneilcin:
Why are there two versions of very similar config files?

coffeepac:
@oneilcin the second config is a reference config for a fully private topology of kraken

PR #932, opened 2017-11-05
Update default k8s version for GKE to 1.7.8 to avoid auto-upgrade

davidewatson:
during cluster creation.


oneilcin:
Did you test bringing GKE cluster up with this setting?  On Thursday last week this was failing for me: 
since the default master was being set at 1.7.6.  All zones had not yet been migrated.

davidewatson:
That's odd. There is no mention of that limitation in the [release notes](https://cloud.google.com/container-engine/release-notes). In any case, CI stands up both an AWS and GKE cluster and it worked.

PR #948, opened 2017-11-09
AWS Cluster-Autoscaling 

leahnp:
Related to: https://github.com/samsung-cnct/kraken-lib/issues/150

**What it does:**
Add Kubernetes cluster-autoscaling addon to clusterNodes and specialNodes via Helm chart.

**Why we need it:**
AWS autoscaling will never be triggered because Kubernetes does not over-schedule nodes, so clusters might be overloaded but because Kubernetes just keeps pods "pending" AWS autoscalers won't know there is a problem since they can only know if a node is at 100% capacity.

The Kubernetes cluster-autoscaler stays in contact with the API and knows if pods are waiting for room to be scheduled and does the work to spin up or down nodes to meet the requirements, restrained by the maxNodes and minNodes. This work means `node.count` will be honored when the cluster is first spun up, but once pods are running the cluster-autoscaler will work to adjust the cluster size. In theory there should never be unused nodes just handing around!

It might seem `node.count` is no longer relevant, but we have future work planned: https://github.com/samsung-cnct/kraken-lib/issues/936 to disable cluster-autoscaling and return to the desired `node.count`.

### DIFF
--- a/ansible/roles/kraken.config/files/config.yaml
+++ b/ansible/roles/kraken.config/files/config.yaml
@@ -147,8 +147,9 @@ definitions:
 
   autoScalingConfigs:
     - &defaultAutoScaling
-      minNodes: 0
+      minNodes: 1
       maxNodes: 10
+      enabled: true
 
   osConfigs:
     - &defaultCoreOs
@@ -418,7 +419,7 @@ deployment:
           osConfig: *defaultCoreOs
           nodeConfig: *defaultAwsClusterNode
           keyPair: *defaultKeyPair
-          autoScaling: *defaultAutoscaling
+          autoScaling: *defaultAutoScaling
         - name: specialNodes
           count: 2
           kubeConfig: *defaultKube
@@ -426,7 +427,7 @@ deployment:
           osConfig: *defaultCoreOs
           nodeConfig: *defaultAwsSpecialNode
           keyPair: *defaultKeyPair
-          autoScaling: *defaultAutoscaling
+          autoScaling: *defaultAutoScaling
       fabricConfig: *kubeVersionedFabric
       kubeAuth: *rbacKubeAuth
       helmConfig: *defaultHelm
--- a/ansible/roles/kraken.config/files/config.yaml
+++ b/ansible/roles/kraken.config/files/config.yaml
@@ -408,6 +408,9 @@ deployment:
           keyPair: *defaultKeyPair
         - name: clusterNodes
           count: 10
+          # cluster auto-scaling
+          maxNodes: 10
+          minNodes: 1
           kubeConfig: *defaultKube
           containerConfig: *defaultDocker
           osConfig: *defaultCoreOs
@@ -416,6 +419,8 @@ deployment:
         - name: specialNodes
           count: 2
           # m4.large
+          maxNodes: 10
+          minNodes: 1
           kubeConfig: *defaultKube
           containerConfig: *defaultDocker
           osConfig: *defaultCoreOs
--- a/ansible/roles/kraken.config/files/config.yaml
+++ b/ansible/roles/kraken.config/files/config.yaml
@@ -419,7 +419,7 @@ deployment:
           osConfig: *defaultCoreOs
           nodeConfig: *defaultAwsClusterNode
           keyPair: *defaultKeyPair
-          autoScaling: *defaultAutoScaling
+          autoScalingConfig: *defaultAutoScaling
         - name: specialNodes
           count: 2
           kubeConfig: *defaultKube
@@ -427,7 +427,7 @@ deployment:
           osConfig: *defaultCoreOs
           nodeConfig: *defaultAwsSpecialNode
           keyPair: *defaultKeyPair
-          autoScaling: *defaultAutoScaling
+          autoScalingConfig: *defaultAutoScaling
       fabricConfig: *kubeVersionedFabric
       kubeAuth: *rbacKubeAuth
       helmConfig: *defaultHelm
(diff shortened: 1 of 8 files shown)


ansible/roles/kraken.config/files/config.yaml
```diff
@@ -408,6 +408,9 @@ deployment:
           keyPair: *defaultKeyPair
         - name: clusterNodes
           count: 10
+          # cluster auto-scaling
```
coffeepac: I was thinking we would have an optional key at the nodepool level named 'autoscaling'.  the autoscaling config block would then have settings for max/min and a boolean 'enabled'.  that way we can do things like:
- prevent specific nodepools from being autoscaled
- allow future autoscaling config to be bucketed off

leahnp: What if you wanted to enable `autoscaling` for 2 different node groups with different min/maxNodes?

coffeepac: reference two different autoscaling configs. 

I"m looking at the nodeConfig block for prior art.

ansible/roles/kraken.config/files/config.yaml
```diff
@@ -415,12 +415,16 @@ deployment:
           keyPair: *defaultKeyPair
         - name: specialNodes
           count: 2
-          # m4.large
           kubeConfig: *defaultKube
           containerConfig: *defaultDocker
           osConfig: *defaultCoreOs
           nodeConfig: *defaultAwsSpecialNode
           keyPair: *defaultKeyPair
+      autoscaling:
```
coffeepac: ahh, I see.  no, I explained poorly.  Repeating myself incase my comment is lost to time:  

autoscaling is an individual nodepool property, like nodeConfig.  so each nodepool would have (or share) an autoscaling config reference.

leahnp: Got it - thanks

schemas/config/v1/autoScalingConfig.json
```diff
@@ -0,0 +1,31 @@
...
+
+  "properties": {
+    "minNodes": {
+      "description": "Minimum number of nodes the cluster auto-scaler should scale down to.",
+			"minimum": 1,
+      "type": "number"
+    },
+    "maxNodes": {
+      "description": "Maximum number of nodes the cluster auto-scaler should scale up to.",
+      "type": "number"
```
coffeepac: this should also have a minimum of 1.

schemas/config/v1/autoScalingConfig.json
```diff
@@ -0,0 +1,31 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "http://judkins.house/apis/k2/v1/autoScalingConfig.json",
+  "$$target": "autoScalingConfig.json",
+  "title": "AutoScaling configurations",
+  "description": "Describes the autoScaling configuration for a nodePool.",
+
+  "properties": {
+    "minNodes": {
+      "description": "Minimum number of nodes the cluster auto-scaler should scale down to.",
+			"minimum": 1,
```
coffeepac: too much spacing at the start of this line

ansible/roles/kraken.services/tasks/run-services.yaml
```diff
@@ -198,3 +198,36 @@
...
+    caTag: v0.6.0
+  when:
+    - ('{{ kubernetes_minor_versions[cluster.name] }}' == 'v1.7')
+
+
+- name: Set cluster-autoscaling version for v1.6 clusters
+  set_fact:
+    caTag: v0.5.0
+  when: 
+    - ('{{ kubernetes_minor_versions[cluster.name] }}' == 'v1.6')
```
coffeepac: would be cleaner, I think, to create a hash of caTag values and then access that hash with the minor version.  So the hash would be:

caTags = 
```
v1.6 => v0.5.0
v1.7 => v0.6.0
v1.8 => v1.0.1
```
and then you would access it like so:
`caTags[kubernetes_minor_versions[cluster.name]]`

This would make updating the list easy as we update our supported kubernetes versions

schemas/config/v1/awsNodeConfig.json
```diff
@@ -26,6 +26,9 @@
    "providerConfig": {
       "$ref": "awsNodeProviderConfig.json"
     },
+    "autoScalingConfig": {
+      "$ref": "autoScalingConfig.json"
+    },
```
coffeepac: I don't understand why this block is being added.  autoScaling isn't a property of the awsNode as far as I can tell.

ansible/roles/kraken.services/tasks/run-services.yaml
```diff
@@ -198,3 +198,36 @@
...
+      --namespace kube-system
+  environment:
+    KUBECONFIG: "{{ kubeconfig }}"
+    HELM_HOME: "{{ helm_home }}"
+    PATH: "{{ helm_path }}:{{ ansible_env.PATH }}"
+  with_items:
+    - "{{ cluster.nodePools }}"
+  when:
+    - item.autoScalingConfig is defined
+
```
coffeepac: this empty line is causing the ansible linter to fail.

ansible/roles/kraken.services/tasks/run-services.yaml
```diff
@@ -198,3 +198,24 @@
     - cluster_services is defined
     - helm_command is defined
     - item.registry is defined
+
+- name: Set cluster-autoscaling version for v1.8 clusters
```
coffeepac: this is for all clusters, not just 1.8

Issue #939, opened 2017-11-07
Document a minimal method by which a bare-metal cluster can be brought up 


joejulian:
Kraken does not yet have bare-metal support. Document a way in which a functional cluster can be created with a combination of kraken and manual steps.

coffeepac:
Expected process (details to be ironed out while testing):
- (manual) create MAAS "pod" that has a private network (assume 10.0.0.0/8 for now)
- (manual) assign 5 nodes to this pod
- (manual) label nodes as etcd (1), master (1), worker (3)
- (?) generate cloud_config files
  - one of:  manual template to fill out, scripted template to execute, kraken tag
  - required output: CA directory
- (?) allocate, deploy etcd node and grab ip
  - one of: script, kraken tag
  - required output: ip of node
- (?) allocate, deploy master node using etcd ip
  - one of: script, kraken tag
  - required output: ip of node
- (?) allocate, deploy worker nodes using master ip
  - one of: script, kraken tag
- (kraken) configure kubernetes cluster using master ip and CA directory

countspongebob:
Looks like what we discussed on the board. Needs expansion, though.

coffeepac:
@countspongebob first cut with expansion

Expected process (details to be ironed out while testing):

  The first several steps to creating a kraken managed kubernetes cluster in MAAS 
are manual and can probably be handled entirely from the MAAS GUI.  We will not go 
into detail here about how to perform these actions as the operator is assumed to 
be knowledgable about MAAS.
  1. Create a new "pod" in MAAS.  This pod should have its own private network space,
  preferably a full 10.0.0.0/8 block.  Nodes in this pod will need access to the 
  internet at large to download images.  
  2. Assign 5 nodes to this pod.  All nodes should have at least 2 cores an 4GB of 
  memory, more is fine.
  3. Label the five nodes as:
    - etcd (x1)
    - master (x1)
    - worker (x3)

  The next step will be using kraken-lib directly to generate the cloud config files.
The exact incantation will be provided later but it will look something like:
  `docker run -v <volumes> -e EXTRA_VARS=<specific vars> quay.io/samsung_cnct/kraken-lib:v0.11 /bin/up.sh --tags cloud-config-only`

This command will produce an output directory that contains, among other things,
a diretory named `cloud-config` that contains the cloud config files.  It will
also create a directory named `ca` that contains SSL certificate information 
that is necessary for secure intra-cluster communications.  Please make a note of
the full path to both of these directories.
[...]

PR #950, opened 2017-11-09
changes to add a post_provider ansible tag

coffeepac:
the post_provider ansible tag runs all roles necessary to complete
the configuration of a cluster after it has been provisioned using
cloud-config files generated by kraken.  this is useful when
manually creating the nodes in a new cloud provider and using
kraken for everything else.

this is part of #939

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -18,7 +18,7 @@ KRAKEN_HELP=false
 UPDATE_NODEPOOLS=''
 ADD_NODEPOOLS=''
 REMOVE_NODEPOOLS=''
-KRAKEN_ENDPOINT=''
+K8S_ENDPOINT=''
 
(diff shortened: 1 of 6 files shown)


lib/common.sh
```diff
@@ -18,6 +18,7 @@ KRAKEN_HELP=false
 UPDATE_NODEPOOLS=''
 ADD_NODEPOOLS=''
 REMOVE_NODEPOOLS=''
+KRAKEN_ENDPOINT=''
```
leahnp: Can you explain this variable to me?

coffeepac: KRAKEN_ENDPOINT is the url of the API server for the kubernetes cluster.  it is normally something that gets output by terraform.  The tag added in this PR assumes that terraform isn't being run but some other (manual, different tool, etc.) is doing the node provisioning.

PR #957, opened 2017-11-10
admission control and runtime config flags now exposed

alejandroEsc:
Allows for them  to be changed them with our own, for example:

```
 apiServerConfigs:
    - &defaultApiServer
      name: defaultApiServer
      kind: apiServer
      loadBalancer: cloud
      state:
        etcd: *defaultEtcd
      events:
        etcd: *defaultEtcdEvents
      flags:
        admissionControl: "NamespaceLifecycle,LimitRanger,ServiceAccount,NodeRestriction,PersistentVolumeLabel,DefaultStorageClass,ResourceQuota,DefaultTolerationSeconds,Initializers"
        runtimeConfig: "batch/v2alpha1=true,admissionregistration.k8s.io/v1alpha1=true"
```

allows us to now run istio with automatic injection. A configmap may be required in future work.

### DIFF
--- a/ansible/roles/kraken.nodePool/kraken.nodePool.master/templates/api-server-manifest.yaml.jinja2
+++ b/ansible/roles/kraken.nodePool/kraken.nodePool.master/templates/api-server-manifest.yaml.jinja2
@@ -27,7 +27,7 @@
         - --allow-privileged=true
         - --service-cluster-ip-range={{ cluster.network }}
         - --secure-port=443
-{% if node.apiServerConfig.flags is defined and if node.apiServerConfig.flags.admissionControl is defined %}
+{% if node.apiServerConfig.flags is defined and node.apiServerConfig.flags.admissionControl is defined %}
         - --admission-control={{ node.apiServerConfig.admissionControl }}
 {% else %}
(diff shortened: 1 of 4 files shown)


davidewatson:
/lgtm

The [Initializers](https://kubernetes.io/docs/admin/admission-controllers/#initializers-alpha) admission controller is an alpha feature in both 1.7 and 1.8, so we can't add it by default.

~~I have some reservations about allowing users to choose arbitrary admission controllers. We depend on the existing controllers, so if a user leaves one out, things will break in ways which might be difficult to diagnose, leading to a poor user experience.~~ That said, our config file is rife with similar options.

Do we want to add the default options to the config? Or leave it (semi) undocumented?

davidewatson:
I think we should leave it out. Otherwise everytime we change the set of default controllers we will have to update the config file and the template. This is an advanaced option.

alejandroEsc:
@davidewatson oh i forgot to remove that as the default. didnt mean to include it.

alejandroEsc:
@davidewatson those arent there?

davidewatson:
@samsung-cnct/kraken-reviewers: Can someone else review this? I am ready to merge, but since it is a user-facing change, which we may have to support for some time, I would like another pair of eyes on it.

PR #960, opened 2017-11-10
only allow lowercase cluster names

leahnp:
GKE only allows lowercase letters.

This work was requested here: https://github.com/samsung-cnct/kraken/pull/211

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,7 +10,7 @@ gke_cloud_test_timeout = 60  // Should be about 4 min but can be as long as 50 f
 e2e_test_timeout       = 18  // Should be about 15 min
 cleanup_timeout        = 60  // Should be about 6 min
 
-e2e_kubernetes_version = "v1.8.2"
+e2e_kubernetes_version = "v1.8.3"
 e2etester_version      = "0.2"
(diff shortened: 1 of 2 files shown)


alejandroEsc:
is this change backwards compatible, if my cluster is already up will i be able to update?

leahnp:
@alejandroEsc Good point - @coffeepac thoughts?

coffeepac:
man, i feel like I've hit this before.  okay, no, we can't do it because it could cause existing clusters to break.

alejandroEsc:
If this is a gke specific issue, im guessing clusters were not brought up with caps? if so then we can just go to that branch and make the validations there. If not, then further upstream we can check and make lower case all the names?

leahnp:
I added some documentation about this to kraken - and if you TRY to spin up a GKE cluster you will see the error in this issue: https://github.com/samsung-cnct/kraken/issues/210

So it is pretty clear to the user what the problem is, just kind of annoying.

alejandroEsc:
How about capturing the warning instead then and displaying it to the user.

leahnp:
The warning is displayed to the user. Sorry if I wasn't clear about that.

PR #961, opened 2017-11-10
default cluster auto-scaling enabled for GKE clusters

leahnp:
Fixes: https://github.com/samsung-cnct/kraken-lib/issues/150
Related to: https://github.com/samsung-cnct/kraken-lib/pull/948

Enables cluster-autoscaling in GKE clusters.

NOTES: 
- There is not a Helm chart for GKE cluster auto-scaling. 
- The `--quiet` flag is in a unconventional location, but it does not work listed after the command on a new line.

### DIFF
--- a/ansible/roles/kraken.provider/kraken.provider.gke/tasks/create-cluster.yaml
+++ b/ansible/roles/kraken.provider/kraken.provider.gke/tasks/create-cluster.yaml
@@ -47,6 +47,7 @@
       --max-nodes={{ item.autoScalingConfig.maxNodes }} 
       --node-pool={{ item.name }}
       --zone {{ cluster.providerConfig.zone.primaryZone }}
+      --project 'k8s-work'
   with_items:
     - "{{ cluster.nodePools }}"
   when:
--- a/ansible/roles/kraken.provider/kraken.provider.gke/tasks/create-cluster.yaml
+++ b/ansible/roles/kraken.provider/kraken.provider.gke/tasks/create-cluster.yaml
@@ -40,4 +40,17 @@
     KUBECONFIG: "{{ config_base }}/{{ cluster.name }}/admin.kubeconfig"
     CLOUDSDK_CONTAINER_USE_CLIENT_CERTIFICATE: True
 
+- name: Enable autoscaling
+  command: >
+    gcloud --quiet beta container clusters update {{ cluster.name }} --enable-autoscaling 
+      --min-nodes={{ item.autoScalingConfig.minNodes }} 
+      --max-nodes={{ item.autoScalingConfig.maxNodes }} 
+      --node-pool={{ item.name }}
+      --zone {{ cluster.providerConfig.zone.primaryZone }}
+  with_items:
+    - "{{ cluster.nodePools }}"
+  when:
+    - item.autoScalingConfig is defined
+    - item.autoScalingConfig.enabled == true
+
 - include: update-cluster.yaml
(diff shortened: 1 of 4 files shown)


leahnp:
I ran a few other kraken PR's and they are not having issues with the GKE cluster (re: the above error). 

I am wondering if something is being updated to an incompatible version when GKE enabled autoscaling, I am going to try a few tests locally to see if I can figure out what exactly is happening.

leahnp:
So on my local machine, when enabling the `cluster auto-scaling` I see: 

```"You do not currently have this command group installed. Using it \nrequires the installation of components: [beta]\n\n\nYour current Cloud SDK version is: 162.0.0\nInstalling components from version: 162.0.0\n\n+---------------------------------------------+\n| These components will be installed.```

But then it successfully installs the command group, updates and enables the addon. In CI I think the core issue is: 

```         "Restarting command:",
         "  $ gcloud --quiet beta container clusters update pr-961-3 --enable-autoscaling --min-nodes=1 --max-nodes=10 --node-pool=clusternodes --zone us-east1-b",
         "",
         "ERROR: (gcloud.beta.container.clusters.update) The required property [project] is not currently set.",
         "You may set it for your current workspace by running:",
         "",
         "  $ gcloud config set project VALUE",
         "",
         "or it can be set temporarily by the environment variable [CLOUDSDK_CORE_PROJECT]"```

**Comment by leahnp** [comment]
This is "passing" CI now @davidewatson

File: ansible/roles/kraken.provider/kraken.provider.gke/tasks/create-cluster.yaml [hunk 1]
```diff
@@ -40,4 +40,18 @@
...
+      --min-nodes={{ item.autoScalingConfig.minNodes }} 
+      --max-nodes={{ item.autoScalingConfig.maxNodes }} 
+      --node-pool={{ item.name }}
+      --zone {{ cluster.providerConfig.zone.primaryZone }}
+      --project {{ cluster.providerConfig.project }}
+  with_items:
+    - "{{ cluster.nodePools }}"
+  when:
+    - item.autoScalingConfig is defined
+    - item.autoScalingConfig.enabled == true
```
davidewatson: From (PEP8](https://www.python.org/dev/peps/pep-0008/) "Don't compare boolean values to True or False using ==."

davidewatson:
This job continues to fail CI because an etcd cluster fails to form. I've restarted it. I don't see anything since you started this patch series which would explain it. etcd is still flaky.

davidewatson:
The latest failure looks like a CI problem. From https://common-jenkins.kubeme.io/job/K2/job/PR-961/20/console

> Could not update commit status. Message: Server returned HTTP response code: 201, message: 'Created' for URL: https://api.github.com/repos/samsung-cnct/kraken-lib/statuses/373c62dceb7755468e47ca4c261933889c243e41

> an exception which occurred:
	in field _processor
	in field cause
	in field cause
	in field cause
	in field v
	in field returnAddress
	in field e
	in field program
	in field threads
	in object org.jenkinsci.plugins.workflow.cps.CpsThreadGroup@f797705
Caused: java.io.NotSerializableException: com.fasterxml.jackson.core.json.ReaderBasedJsonParser
	at org.jboss.marshalling.river.RiverMarshaller.doWriteObject(RiverMarshaller.java:860)

@coffeepac: ^^

leahnp:
@davidewatson looks like this passed CI: https://common-jenkins.kubeme.io/job/K2/view/change-requests/job/PR-961/21/

Just one E2E failure, that is expected. Let me know if we can get this merged today.

Issue #963, opened 2017-11-13
BAC-99 ⁃ Cluster autoscaler configuration does not support multiple availability zones/node pools


davidewatson:
tl;dr

I think we should disable the auto-scaler by default until it better supports the multi-AZ configurations we use by default.

# Issues

1) The cluster auto-scaler only partially [supports multiple AZs](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#im-running-cluster-with-nodes-in-multiple-zones-for-ha-purposes-is-that-supported-by-cluster-autoscaler). A complete solution may have to wait until the auto-scaler takes balancing into account when scaling down. Until then, we should at least pass the `--balance-similar-node-groups` flag using `extraArgs` chart values. 

2) The `kraken-lib` implementation installs an instance of the auto-scaler for each `nodePool`. Instead, we should install a single instance and pass in multiple `autposcalingGroups`(one for each kraken `nodePool`).

3) The default `autoscalingConfig` minimum is 1. It should be at least 3 in order to tolerate AZ failure.

leahnp:
RE the issues outlined:

1- I feel like this could be a good opportunity to do some testing and contribute some work/features upstream. You are probably right it should be disabled by default, but finding a way to encourage our team to use it during development could help us find a way to make it better.
2- What are the benefits of this approach?
3- Agreed.

leahnp:
Also, if we are gong to disable this by default, we have a list of our concerns and regularly cross reference with the auto-scaling addon release notes (perhaps in the weekly on-call version).

davidewatson:
1. I agree this might be a good opportunity to contribute. I am not sure when we should encourage other developers to test this feature though. Right now, I think the defaults are causing logging to fail to install.
2. According to the [FAQ](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-key-best-practices-for-running-cluster-autoscaler) using a single auto-scaler is the best practice. I am not sure if/how the auto-scaler avoids continually creating and deleting nodes when there are multiple auto-scalers.
3. PR submitted.

davidewatson:
You're right that we would need a way to follow up on this. Instead of adding to the weekly work, maybe we should have p2 ticket to follow up? I don't expect this to be fixed immediately. At least I did not find a ticket describing the work required.

PR #964, opened 2017-11-13
Change cluster autoscaling defaults to off with a minimum of 3

davidewatson:
nodes (see #963 for discussion).

### DIFF
--- a/ansible/roles/kraken.config/files/config.yaml
+++ b/ansible/roles/kraken.config/files/config.yaml
@@ -147,9 +147,9 @@ definitions:
 
   autoScalingConfigs:
     - &defaultAutoScaling
-      minNodes: 1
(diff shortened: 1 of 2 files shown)


leahnp:
I just made another comment on #963 
`Also, if we are gong to disable this by default, we have a list of our concerns and regularly cross reference with the auto-scaling addon release notes (perhaps in the weekly on-call version). `

Can we put that in the weekly on-call versioning ticket before we merge this so it doesn't fall through the cracks? Or at least make a ticket to check the addon status in the near future. 

Otherwise LGTM

coffeepac:
@leahnp that should be a new issue, maybe not part of weekly on call.  Do you mind filing it?   Should be p1 and on the board.

PR #965, opened 2017-11-13
Convert the default cloud_config data to standard

joejulian:
Change the default cloud_config data to work with standard cloud-init
and move the coreos specific things to coreos specific task modules.


joejulian:
oops, still need to add rkt to this

ansible/roles/kraken.cluster_common/tasks/units.yaml
```diff
@@ -10,15 +10,15 @@
...
-    - mount.part.jinja2
-  when: node.nodeConfig.mounts is defined
-
 - name: "Generate network-environment service units"
   include: cloud_config/units.yaml
   with_items:
     - setup-network-environment.part.jinja2
   when: node.bastionConfig is not defined
+
+- name: Create disk units
```
philoserf: - replaces 'Generate common mount service units'
- looks like more change than it is because of the placing of the replacement

ansible/roles/kraken.cluster_common/tasks/cluster-start.yaml
```diff
@@ -27,5 +35,11 @@
   when: cluster_status_lock_file.stat.exists == True and kraken_action == 'up'
 
 - include: certificate_authority.yaml
+- include: units.yaml
```
philoserf: - looks like more change than it is because of order change

lib/kraken_arguments.sh
```diff
@@ -104,14 +104,16 @@ else
     DNS_ONLY=false
 fi
 
-KRAKEN_EXTRA_VARS="config_path=${KRAKEN_CONFIG} config_base=${KRAKEN_BASE} \
-                   config_forced=${KRAKEN_FORCE} dryrun=${KRAKEN_DRYRUN} \
+KRAKEN_EXTRA_VARS="${KRAKEN_EXTRA_VARS:-""} \
```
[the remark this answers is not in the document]
philoserf: Ah, the name of the commit told me.

bin/update.sh
```diff
@@ -26,9 +26,9 @@ LOG_FILE=$"/tmp/crash-logs"
 # exit trap for crash app
 trap crash_test_update EXIT
 
-K2_CRASH_APP=$(which k2-crash-application)  
-if [ $? -ne 0 ];then  
-	DISPLAY_SKIPPED_HOSTS=0 ansible-playbook ${K2_VERBOSE} -i ansible/inventory/localhost ansible/update.yaml --extra-vars "${KRAKEN_EXTRA_VARS}kraken_action=update" || show_update_error
+K2_CRASH_APP=$(which k2-crash-application)
+if [ $? -ne 0 ];then
+	DISPLAY_SKIPPED_HOSTS=0 ansible-playbook ${K2_VERBOSE} -i ansible/inventory/localhost ansible/update.yaml --extra-vars "kraken_action=update ${KRAKEN_EXTRA_VARS}" || show_update_error
```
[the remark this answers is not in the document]
joejulian: To allow a trailing space instead of a leading space within the quotes.

ansible/cloud_config/coreos/units.yaml
```diff
@@ -0,0 +1,14 @@
+---
+- name: "Rendering service unit: {{ template_name }}"
```
oneilcin: Please explain where the template_name comes from (i.e. an example valid name)

ansible/cloud_config/default/units.yaml
```diff
@@ -0,0 +1,35 @@
...
+    # Create a dict for merging back in to cloud_config containing the new lists.
+    new_data: '{{ { cluster.name: { node.name: { "write_files": write_files } } } }}'
+    new_units: '{{ { cluster.name: { node.name: units + [ unitdata.name ] } } }}'
+
+  # Do this with the first template found. This allows us to override a template on a per-kubernetes-version basis.
+  with_first_found:
+    - "templates/{{ kubernetes_minor_versions[ cluster.name ] }}/{{ template_name }}"
+    - "templates/{{ template_name }}"
+  loop_control:
+    loop_var: template_file
```
oneilcin: Can you please give an overview of what this code is doing?  In the assembly role I saw we copied over what I thought were unit files for non-core-os distro.  Maybe you can explain the relationship between that file : sensible/roles/kraken.assembler/tasks/node.yaml and this one?

ansible/cloud_config/units.yaml
```diff
@@ -3,16 +3,26 @@
   set_fact:
     template_name: "{{ item }}"
```
joejulian: template_name is defined here.

ansible/cloud_config/coreos/units.yaml
joejulian: https://github.com/samsung-cnct/kraken-lib/pull/965#discussion_r153342519

ansible/cloud_config/default/units.yaml
joejulian: coreos has a `coreos.unit` section in the cloud_config user data. The _real_ cloud-init doesn't even have a `coreos` section. Instead, to install and start a service you need to `write_file` and `run_cmd` as two separate dict entries.

Since it's more efficient to start/restart all units from a single command and take advantage of systemd's dependency matrix, I save all the unit names here and make a single run_cmd in assemble that restarts the entire list of units.

Since cloud-init is a standard tool that's used by every other distribution (that I'm aware of), that's why I'm considering that "default".

joejulian: The other piece of magic that's happening here is that you cannot merge list objects in a dict. If you merge dicts, the existing list will be replaced. Since our goal is to add this entry to a list, we first extract the list from the dict, add our item, then build a dict that we can merge back in.

oneilcin: Thanks for the overview! Helps a lot.
LGTM

philoserf:
LGTM
👍 +1:

ansible/roles/kraken.assembler/tasks/nodes.yaml
```diff
@@ -1,32 +1,42 @@
 ---
 - name: Set Cluster Fact
   set_fact:
-    cluster: "{{ a_cluster }}"
+    cluster: "{{ cluster_node_tuple.0 }}"
+    node: "{{ cluster_node_tuple.1 }}"
```
coffeepac: nodepool, yeah?

ansible/roles/kraken.assembler/tasks/nodes.yaml
```diff
@@ -1,32 +1,42 @@
...
 
 - name: Make sure generated folder for full cloud configs is there
   file:
     path: "{{ config_base }}/{{ cluster.name }}/cloud-config"
     state: directory
 
+- name: Generate systemctl start if there are units to start
```
coffeepac: there is no conditional based on there being units.  name is misleading.

ansible/roles/kraken.assembler/tasks/nodes.yaml
joejulian: Yes, changed.

ansible/roles/kraken.assembler/tasks/nodes.yaml
joejulian: True, test added.

PR #970, opened 2017-11-15
added kubectl

DStorck:
fixes https://github.com/samsung-cnct/kraken-lib/issues/962

### DIFF
--- a/ansible/roles/kraken.update/kraken.update.aws/tasks/update-nodes.yaml
+++ b/ansible/roles/kraken.update/kraken.update.aws/tasks/update-nodes.yaml
@@ -2,8 +2,6 @@
 
 - set_fact:
     nodepool: "{{ a_nodepool }}"
-    kubectl: "{{ kubectl_commands[cluster.name] }}"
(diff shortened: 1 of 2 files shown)


ansible/roles/kraken.update/kraken.update.aws/tasks/update-nodes.yaml
```diff
@@ -2,6 +2,8 @@
 
 - set_fact:
     nodepool: "{{ a_nodepool }}"
+    kubectl: "{{ kubectl_commands[cluster.name] }}"
+    kubeconfig: "{{ config_base }}/{{ cluster.name }}/admin.kubeconfig"
```
coffeepac: this fact is being set in update-cluster (which calls update-nodes).  would it make more sense to push the kubectl defintion up to update-cluster and then remove these two lines from here?

DStorck: sure, let me try and make sure it works

PR #972, opened 2017-11-17
gke config file generation

leahnp:
Added `--provider` flag to allow users to specify `GKE` config instead of default `AWS`

Fixes: https://github.com/samsung-cnct/kraken-lib/issues/926

### DIFF
--- a/lib/kraken_arguments.sh
+++ b/lib/kraken_arguments.sh
@@ -34,10 +34,7 @@ case $key in
   shift
   ;;
   -g|--generate)
-  KRAKEN_GENERATE_PATH="${2-"${HOME}/.kraken/config.yaml"}"
-  if [ -n "${2+x}" ]; then
-    shift
-  fi
+  KRAKEN_GENERATE_PATH="${HOME}/.kraken/config.yaml"
   ;;
   -o|--output)
   KRAKEN_BASE="$2"
@@ -51,6 +48,10 @@ case $key in
   K2_VERBOSE="$2"
   shift
   ;;
+  -v|--provider)
+  KRAKEN_PROVIDER="$2"
+  shift
+  ;;
   -h|--help)
   KRAKEN_HELP=true
   shift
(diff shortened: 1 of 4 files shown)


leahnp:
I think I will need to update kraken's Jenkinsfile to explicitly use `--config`

lib/kraken_arguments.sh
```diff
@@ -61,21 +62,28 @@ esac
...
+if [ -n "${KRAKEN_GENERATE_PATH+x}" ]; then
+  KRAKEN_GENERATE_PATH=${KRAKEN_CONFIG}
+
+  if [[ -n ${KRAKEN_PROVIDER+x} ]]; then
+    if [ $KRAKEN_PROVIDER == "GKE" ]; then
+        generate_config "${KRAKEN_GENERATE_PATH}" GKE
+    fi
+  fi
+
+  generate_config "${KRAKEN_GENERATE_PATH}" AWS
```
davidewatson: Won't this will unconditionally generate the AWS config instead of defaulting to it? I think we should parse, default and validate options at the top and then unconditionally use them throughout the remainder of the code.

leahnp:
Actually not kraken Jenkinsfile, but kraken-tools did need an update: https://github.com/samsung-cnct/kraken-tools/pull/155

lib/kraken_arguments.sh
```diff
@@ -34,10 +34,7 @@ case $key in
   shift
   ;;
   -g|--generate)
-  KRAKEN_GENERATE_PATH="${2-"${HOME}/.kraken/config.yaml"}"
-  if [ -n "${2+x}" ]; then
-    shift
-  fi
+  KRAKEN_GENERATE_PATH="${HOME}/.kraken/config.yaml"
```
davidewatson: Depending on the order of the options, this default may overwrite the `--config` value. For example: `kraken --config ~/.kraken/dewDrops.yaml --generate`.

leahnp: Hmmm, I just tested this out on my branch and I am not seeing any buggy behavior - can you explain your concern a bit more?

leahnp: I tried: `$ ./bin/up.sh --provider GKE --config . --generate` & `$ ./bin/up.sh --config . --generate`

lib/kraken_arguments.sh
leahnp: I don't think so - I have tested every iteration and combination of these flags I can think of and the proper provide config ends up in the expected directory. Let me know if you find something that is broken though!

davidewatson: Ah. `generate_config` calls exit. Okay, I'm wrong.

PR #973, opened 2017-11-17
Reformat master terraform template for plausibly readability.

davidewatson:


davidewatson:
Dashes disable `trim_blocks`, plus signs enable it: http://jinja.pocoo.org/docs/2.10/templates/#whitespace-control

PR #974, opened 2017-11-17
Implement etcd monitoring by creating an internal loadbalancer to

### LINKED ISSUE #825: Etcd nodes need fluent-bit for logging
Since etcd is not hosted by k8s, we will not be able to capture its logs using the [fluent-bit dameonset](https://github.com/samsung-cnct/k2-logging-fluent-bit-daemonset). One solution would be to render systemd units to deploy fluent-bit on etcd nodes.

davidewatson:
the apiservers and pointing fluent-bit at it (fixes #825). Note
that this change makes the apiserver.{{ cluster.name }}.internal
CNAME to point to the internal ELB and therefore internal IP
addresses (as the name suggests!).

### DIFF
--- a/ansible/roles/kraken.provider/kraken.provider.aws/templates/kraken.provider.aws.module.master.tf.jinja2
+++ b/ansible/roles/kraken.provider/kraken.provider.aws/templates/kraken.provider.aws.module.master.tf.jinja2
@@ -192,12 +192,55 @@ resource "aws_elb" "master_elb" {
   }
 }
 
+resource "aws_elb" "master_internal_elb" {
+  name = "${var.master_name}-master-internal-elb"
+
+  internal = true
+
+  cross_zone_load_balancing = true
+
+  {% if master.apiServerConfig.loadBalancerConfigs is defined %}
+  subnets = [{% set comma = joiner(",") %}{% for subnet in master.apiServerConfig.loadBalancerConfigs.subnet %}{{ comma() }}"${var.{{subnet}}_subnet_id}"{% endfor %}]
+  {% else %}
+  subnets = [{% set comma = joiner(",") %}{% for subnet in master.nodeConfig.providerConfig.subnet %}{{ comma() }}"${var.{{subnet}}_subnet_id}"{% endfor %}]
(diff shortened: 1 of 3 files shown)


leahnp:
Does it make sense to try and make this more general ie: "logging daemon" and let people configure it with whatever logging service they want? That might be overkill for the initial pass at etcd logging...just curious your viewpoint on that.

Otherwise this is looking good, is it ready to be merged or are we waiting on something?

davidewatson:
I think this PR is almost, but not quite, ready to be merged.

Regarding a "logging daemon" configuration. You make a good point and it's one that I considered. I think a "logging daemon" configuration would be too specific though. What about the requirements for other services?

Hard coding the open ports on the internal LB is not great. Doing so makes assumptions about the logging service which may not be valid. I think what we really want is a general open port list for the internal LB. We already have something like that for the external LB (see the `ingressAcl:` and `ingressSecurity:` stanzas).

My main concern is that the current configuration file has the (external) port lists as top-level entries, so I do not think it is possible to extend the schema in an orthogonal way. I think the right thing to do is extend the schema to allow arbitrary LBs to be defined - a pretty big change. I propose we handle this as an additional issue.

Thoughts?

ansible/roles/kraken.provider/kraken.provider.aws/templates/kraken.provider.aws.module.master.tf.jinja2
```diff
@@ -192,12 +192,55 @@ resource "aws_elb" "master_elb" {
   }
 }
 
+resource "aws_elb" "master_internal_elb" {
```
leahnp: Does it make sense to indicate somewhere this is for etcd communication? Will it potentially be used for anything else?

davidewatson:
I don't think we've discussed where the fluent-bit logs will be stored for CI. We should.

Note that this PR will likely be superseded by https://github.com/samsung-cnct/kraken-lib/pull/1015 I'm not super keen on encoding LB configuration in the `kraken` config, nor requiring an additional LB for logging...

PR #977, opened 2017-11-17
Updated versions for weave, flannel, and calico

oneilcin:
Did a quick cluster up and get pods for both aws and gke.  Will let CI do the rest

### DIFF
--- a/ansible/roles/kraken.config/files/config.yaml
+++ b/ansible/roles/kraken.config/files/config.yaml
@@ -41,7 +41,7 @@ definitions:
             version: v2.6.1
             location: quay.io/calico/node
           flannel:
-            version: v0.8.0
+            version: v0.9.1
             location: quay.io/coreos/flannel
         network:
           network: 10.128.0.0/10
@@ -60,10 +60,10 @@ definitions:
             version: v1.10.0
             location: quay.io/calico/cni
           calicoNode:
-            version: v2.4.1
+            version: v2.6.1
             location: quay.io/calico/node
           flannel:
-            version: v0.8.0
+            version: v0.9.1
             location: quay.io/coreos/flannel
         network:
           network: 10.128.0.0/10
@@ -87,10 +87,10 @@ definitions:
       options:
         containers:
           weave:
-            version: 1.9.8
+            version: 2.1.1
             location: weaveworks/weave-kube
           weave_npc:
-            version: 1.9.8
+            version: 2.1.1
             location: weaveworks/weave-npc
         network:
           network: 10.128.0.0/10
--- a/ansible/roles/kraken.config/files/config.yaml
+++ b/ansible/roles/kraken.config/files/config.yaml
@@ -60,10 +60,10 @@ definitions:
             version: v1.10.0
             location: quay.io/calico/cni
           calicoNode:
-            version: v2.6.2
+            version: v2.4.1
             location: quay.io/calico/node
           flannel:
-            version: v0.9.1
+            version: v0.8.0
             location: quay.io/coreos/flannel
         network:
           network: 10.128.0.0/10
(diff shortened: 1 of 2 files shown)


ansible/roles/kraken.config/files/config.yaml
```diff
@@ -60,10 +60,10 @@ definitions:
             version: v1.10.0
             location: quay.io/calico/cni
           calicoNode:
-            version: v2.4.1
+            version: v2.6.2
```
guineveresaenger: this should remain at v2.4.1 (the previous version)

ansible/roles/kraken.config/files/config.yaml
```diff
@@ -60,10 +60,10 @@ definitions:
...
             location: quay.io/calico/node
           flannel:
-            version: v0.8.0
+            version: v0.9.1
```
guineveresaenger: also leave as it was

ansible/roles/kraken.config/files/gke-config.yaml
```diff
@@ -59,10 +59,10 @@ definitions:
             version: v1.10.0
             location: quay.io/calico/cni
           calicoNode:
-            version: v2.4.1
+            version: v2.6.2
```
guineveresaenger: leave as it was

ansible/roles/kraken.config/files/gke-config.yaml
```diff
@@ -59,10 +59,10 @@ definitions:
...
             location: quay.io/calico/node
           flannel:
-            version: v0.8.0
+            version: v0.9.1
```
guineveresaenger: leave as it was

ansible/roles/kraken.config/files/config.yaml
oneilcin: How do you know to leave just this one?

ansible/roles/kraken.config/files/gke-config.yaml
```diff
@@ -102,10 +102,10 @@ definitions:
       options:
         containers:
           weave:
-            version: 1.9.8
+            version: 2.1.1
             location: weaveworks/weave-kube
           weave_npc:
-            version: 1.9.8
+            version: 2.1.1
```
guineveresaenger: leave as it was

ansible/roles/kraken.config/files/gke-config.yaml
```diff
@@ -102,10 +102,10 @@ definitions:
       options:
         containers:
           weave:
-            version: 1.9.8
+            version: 2.1.1
```
guineveresaenger: leave as it was

guineveresaenger: actually, this one's fine to update. For cleanup, this stanza should be removed, as it's the same as the default.

ansible/roles/kraken.config/files/gke-config.yaml
guineveresaenger: also fine to update.

ansible/roles/kraken.config/files/config.yaml
```diff
@@ -87,10 +87,10 @@ definitions:
       options:
         containers:
           weave:
```
joejulian: We have no automated testing for weave. Have you at least manually tested this version?

I'd rather not change it if we don't test it. At least if it's broken and we didn't touch it it's obvious why it's broken. If we change it it should imply it worked at the time we changed it.

guineveresaenger: Do we have automated testing for calico?

oneilcin: @joejulian Best way to manually test weave is working?
Note: if we are not going to test this, it shouldn't be in our on-call version check list

oneilcin:
Tested with Weave in the config for both AWS and GKE.  All services and pods were healthy.  Created an nginx pod, and then curled to it from a node other than the one it is running on.  All good!

ansible/roles/kraken.config/files/config.yaml
oneilcin: tested

PR #983, opened 2017-11-21
Fix namespace deletion.

davidewatson:


leahnp:
Are all k8s "namespaces" "service namespaces"?

davidewatson:
Note that the change to the `when` clause is a change in behavior. We used to unconditionally attempt to delete namespaces and then ignored any errors. Now we only delete namespaces on down. This might leak namespaces, but I think that is still possible if a user removes a chart from the kraken config and there are no other charts with that namespace.

ansible/roles/kraken.services/tasks/kill-services.yaml
```diff
@@ -90,9 +90,9 @@
 
 - name: Delete all service namespaces
   command: >
-    {{ kubectl }} --kubeconfig={{ kubeconfig }} delete namespace {{ item }}
+    {{ kubectl }} --kubeconfig={{ kubeconfig }} delete namespace {{ item.namespace }}
   with_items: "{{ cluster_namespaces }}"
-  when: cluster_namespaces is defined
+  when: kraken_action == 'down'
```
joejulian: Why bother? If the cluster is being brought down, there will be no etcd data left anyway.

davidewatson: "The only way to be sure."

davidewatson: Ack.

PR #990, opened 2017-11-29
Updates to k8s and dependencies

guineveresaenger:


philoserf:
or rather than a pr per item a commit per item. lets us revert a single commit the gittish way

leahnp:
@philoserf will per commit work though, considering we squash and merge?

PR #991, opened 2017-11-29
Pre provider

coffeepac:
adds a new tag to kraken-lib 'pre_provider' that runs all roles before the provider selector.  This is useful for creating the cloud-config artifacts and stopping.  possibly only useful as a development tool.


leahnp:
Passing CI - E2E "Networking Granular Checks" is the only test failing, going to merge

PR #996, opened 2017-12-05
Fail earlier if gke cluster fails to deploy

gochist:
Fail earlier if gke cluster fails to deploy (fixes https://github.com/samsung-cnct/kraken-lib/issues/548)

### DIFF
--- a/ansible/roles/kraken.provider/kraken.provider.gke/tasks/create-cluster.yaml
+++ b/ansible/roles/kraken.provider/kraken.provider.gke/tasks/create-cluster.yaml
@@ -9,6 +9,10 @@
   register: deployment_result
   ignore_errors: yes
 
+- name: Fail if RESOURCE_ERROR occurs when deploy cluster
+  fail: "msg='Deployment {{ cluster.name }}-k2-deployment failed with: {{ deployment_result.stderr }}'"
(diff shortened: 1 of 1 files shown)


ansible/roles/kraken.provider/kraken.provider.gke/tasks/create-cluster.yaml
```diff
@@ -9,6 +9,10 @@
   register: deployment_result
   ignore_errors: yes
 
+- name: Fail if RESOURCE_ERROR occurs when deploy cluster
+  fail: "msg='Deployment {{ cluster.name }}-k2-deployment failed with: {{ deployment_result.stderr }}'"
```
DStorck: we aren't using "k2" anymore, perhaps just replace with "kraken" or "kraken-lib"

gochist: @DStorck I think it is needed to open another issue to change name of its deployment. Currently it forces to use 'k2' in its cluster name.

DStorck: oh yeah, we should file an issue for that. we have been slowly working to find all the references and change them.

PR #997, opened 2017-12-05
Gitlab ci config 

leahnp:
Fixes: https://github.com/samsung-cnct/kraken-lib/issues/989


.gitlab-ci.yml
```diff
@@ -0,0 +1,59 @@
+---
+# This pipeline requires 5 variables to be set in the CI/CD Settings
+# hint for creating the base64 encoded content: `cat .ssh/id_rsa|base64|pbcopy`
+
+# SSH_KEY - base64 encoded ssh key (think ~/.ssh/id_rsa)
+# SSH_PUBLIC_KEY - base64 encoded ssh public key (think ~/.ssh/id_rsa.pub)
+# AWS_CREDENTIALS - base64 encoded credentials file (think ~/.aws/credentials)
+# AWS_CONFIG - base64 encoded config file (think ~/.aws/config)
+# GKE_CONFIG - base64 encoded json file (think ~/.config/gcloud/someconfig.json)
+image: docker:latest
```
philoserf: just a style point. I like a little whitespace between the package level comment and the code.

.gitlab-ci.yml
```diff
@@ -0,0 +1,59 @@
...
+
+services:
+  - docker:dind
+
+variables:
+  DOCKER_HOST: tcp://localhost:2375
+  DOCKER_DRIVER: overlay2
+# images
```
philoserf: is the comment needed? i do not think so. better no comment except when we really need 'em.

> this is my ruby/golang worldview and python/rust person might argue with me

.gitlab-ci.yml
```diff
@@ -0,0 +1,59 @@
...
+#     paths:
+#       - $kraken-lib
+#   script:
+#     - 
+
+prep:ansible_lint:
+  stage: prep
+  image: $ANSIBLE_LINT
+  script:
+    - ansible-lint ansible/*.yaml
```
philoserf: I do not think this will get all the Ansible YAML files. You'll need something like this `ansible-lint ansible/*.yaml ansible/**/*.yaml`

As is it'll only get one level deep. I'll look at `ansible-lint` for a better way.

philoserf: this hits every ansible YAML file one by one and suggestions we have some _de_linting work to do.

`ls ansible/*.yaml ansible/**/*.yaml | xargs -I {} ansible-lint {}`

philoserf: perhaps we should yamllint them too/first

`ls ansible/*.yaml ansible/**/*.yaml | xargs -I {} yamllint {}`

philoserf:
Keep in mind please.

Not all comments are about conversion from one Ci/CD system to another.

leahnp:
@philoserf thanks for the review - this is just a WIP though to test the jobs in gitlab, I will let you know when it's ready for review!

.gitlab-ci.yml
```diff
@@ -0,0 +1,221 @@
...
+    BUILD_ID: $CI_PIPELINE_ID
+  image: $KRAKEN_TOOLS
+  script:
+    - mkdir -p /root/.ssh /root/.aws/
+    - echo $SSH_KEY | base64 -d > /root/.ssh/id_rsa
+    - chmod og-wrx /root/.ssh/id_rsa
+    - echo $SSH_PUBLIC_KEY | base64 -d > /root/.ssh/id_rsa.pub
+    - echo $AWS_CREDENTIALS | base64 -d > /root/.aws/credentials
+    - echo $AWS_CONFIG | base64 -d > /root/.aws/config
+    - PWD=`pwd` ./bin/up.sh --config $CI_PROJECT_DIR/cluster/aws/config.yaml --output $PWD/cluster/aws/ -t dryrun
```
davidewatson: I think we should probably pass `--verbose` here as well as anywhere else we call kraken-lib.

.gitlab-ci.yml
```diff
@@ -0,0 +1,231 @@
...
+  image: $KRAKEN_TOOLS
+  when: always
+  script: 
+    - mkdir -p /root/.ssh /root/.aws/
+    - echo $SSH_KEY | base64 -d > /root/.ssh/id_rsa
+    - chmod og-wrx /root/.ssh/id_rsa
+    - echo $SSH_PUBLIC_KEY | base64 -d > /root/.ssh/id_rsa.pub
+    - echo $AWS_CREDENTIALS | base64 -d > /root/.aws/credentials
+    - echo $AWS_CONFIG | base64 -d > /root/.aws/config
+    - ./bin/down.sh --config $CI_PROJECT_DIR/cluster/aws/config.yaml --output $CI_PROJECT_DIR/cluster/aws/
```
davidewatson: This also needs a `--verbose`.

.gitlab-ci.yml
```diff
@@ -0,0 +1,231 @@
...
+  stage: cleanup
+  image: $KRAKEN_TOOLS
+  when: always
+  script:
+    - mkdir -p /root/.ssh /root/.config/gcloud/
+    - echo $SSH_KEY | base64 -d > /root/.ssh/id_rsa
+    - chmod og-wrx /root/.ssh/id_rsa
+    - echo $SSH_PUBLIC_KEY | base64 -d > /root/.ssh/id_rsa.pub
+    - echo $GKE_CONFIG | base64 -d > /root/.config/gcloud/patrickRobot.json
+    - ./bin/down.sh --config $CI_PROJECT_DIR/cluster/gke/config.yaml --output $CI_PROJECT_DIR/cluster/gke/
```
davidewatson: `--verbose`

.gitlab-ci.yml
```diff
@@ -0,0 +1,231 @@
...
+      - $CI_PROJECT_DIR/cluster/aws/**
+    expire_in: 2 hour
+  script: 
+    - mkdir -p /root/.ssh /root/.aws/
+    - echo $SSH_KEY | base64 -d > /root/.ssh/id_rsa
+    - chmod og-wrx /root/.ssh/id_rsa
+    - echo $SSH_PUBLIC_KEY | base64 -d > /root/.ssh/id_rsa.pub
+    - echo $AWS_CREDENTIALS | base64 -d > /root/.aws/credentials
+    - echo $AWS_CONFIG | base64 -d > /root/.aws/config
+    - ./bin/up.sh --config $CI_PROJECT_DIR/cluster/aws/config.yaml --output $CI_PROJECT_DIR/cluster/aws/
```
davidewatson: `--verbose`

.gitlab-ci.yml
```diff
@@ -0,0 +1,231 @@
...
+    paths:
+      - $CI_PROJECT_DIR/cluster/gke/**
+    expire_in: 2 hour
+  script:
+    - mkdir -p /root/.ssh /root/.config/gcloud/
+    - echo $SSH_KEY | base64 -d > /root/.ssh/id_rsa
+    - chmod og-wrx /root/.ssh/id_rsa
+    - echo $SSH_PUBLIC_KEY | base64 -d > /root/.ssh/id_rsa.pub
+    - echo $GKE_CONFIG | base64 -d > /root/.config/gcloud/patrickRobot.json
+    - ./bin/up.sh --config $CI_PROJECT_DIR/cluster/gke/config.yaml --output $CI_PROJECT_DIR/cluster/gke/
```
davidewatson: `--verbose`

.gitlab-ci.yml
```diff
@@ -0,0 +1,231 @@
...
+  only:
+    - tags
+    - branches
+  image: $KRAKEN_TOOLS
+  artifacts: 
+    paths:
+      - $CI_PROJECT_DIR/cluster/aws/config.yaml 
+    expire_in: 2 hour
+  script:
+    - ./bin/up.sh --generate --config $CI_PROJECT_DIR/cluster/aws/config.yaml
```
davidewatson: Will the config used be saved as an artifact, or should we `cat` it out here? One wrinkle wit catting the config is that we do not want to include any sensitive data in the logs. For example the AWS credentials stanza:

```
      authentication:
        accessKey:
        accessSecret:
```

.gitlab-ci.yml
```diff
@@ -0,0 +1,231 @@
...
+    BUILD_ID: $CI_PIPELINE_ID
+  image: $KRAKEN_TOOLS
+  script:
+    - mkdir -p /root/.ssh /root/.aws/
+    - echo $SSH_KEY | base64 -d > /root/.ssh/id_rsa
+    - chmod og-wrx /root/.ssh/id_rsa
+    - echo $SSH_PUBLIC_KEY | base64 -d > /root/.ssh/id_rsa.pub
+    - echo $AWS_CREDENTIALS | base64 -d > /root/.aws/credentials
+    - echo $AWS_CONFIG | base64 -d > /root/.aws/config
+    - PWD=`pwd` ./bin/up.sh --config $CI_PROJECT_DIR/cluster/aws/config.yaml --output $PWD/cluster/aws/ -t dryrun --verbose true
```
DStorck: CI is choking on this right now for verbose, try passing in -vvv instead of --verbose

.gitlab-ci.yml
leahnp: @davidewatson the config is saved as an artifact for the life of the pipeline. To my knowledge all the secrets are secured by using the gitlab provided "secret variables" - https://git.cnct.io/help/ci/variables/README#secret-variables

Does that address your concern?

.gitlab-ci.yml
leahnp: Removed --verbose as it appears to not be functioning correctly

.gitlab-ci.yml
leahnp: Removed --verbose as it appears to not be functioning correctly

.gitlab-ci.yml
leahnp: Removed --verbose as it appears to not be functioning correctly

.gitlab-ci.yml
leahnp: Removed --verbose as it appears to not be functioning correctly

.gitlab-ci.yml
davidewatson: Yes, thank you!

philoserf: my hope for the humans over the bots and the ci systems is this

have everything that can be done independently of automation in a human executable place
so here `./bin/up.sh ...` and others like it would be in a `gen-test-config` script or a gen-test-config make task.
the CI/CD systems config would have `script:...- some command` items that were unique to the CI/CD system.

I will find a way to better articulate the nuance. A too simple example is here: <https://github.com/philoserf/my_training/blob/master/.circleci/config.yml>
A human could do any of these tasks without waiting for Gitlab to run. This is especially useful during CI/CD code/config development time.

.gitlab-ci.yml
```diff
@@ -0,0 +1,231 @@
+---
```
philoserf: the file has a few trailing whitespaces
usually after an `item: \n` YAML object

.gitlab-ci.yml
```diff
@@ -0,0 +1,231 @@
...
+    BUILD_ID: $CI_PIPELINE_ID
+  image: $KRAKEN_TOOLS
+  script:
+    - mkdir -p /root/.ssh /root/.aws/
+    - echo $SSH_KEY | base64 -d > /root/.ssh/id_rsa
+    - chmod og-wrx /root/.ssh/id_rsa
+    - echo $SSH_PUBLIC_KEY | base64 -d > /root/.ssh/id_rsa.pub
+    - echo $AWS_CREDENTIALS | base64 -d > /root/.aws/credentials
+    - echo $AWS_CONFIG | base64 -d > /root/.aws/config
+    - PWD=`pwd` ./bin/up.sh --config $CI_PROJECT_DIR/cluster/aws/config.yaml --output $PWD/cluster/aws/ -t dryrun
```
philoserf: in YAML long lines can be made shorter like this:

```yaml
    - >-
      PWD=`pwd` ./bin/up.sh 
      --config $CI_PROJECT_DIR/cluster/aws/config.yaml
      --output $PWD/cluster/aws/
      -t dryrun
```

philoserf: prefer long form for flags `--tags dryrun` vs `-t dryrun`

build-scripts/conformance-tests.sh
```diff
@@ -11,14 +12,19 @@ gsutil -mq cp "gs://kubernetes-release/release/${KUBERNETES_RELEASE_VERSION}/kub
 gsutil -mq cp "gs://kubernetes-release/release/${KUBERNETES_RELEASE_VERSION}/kubernetes-test.tar.gz" ${cache_dir}
 gsutil -mq cp "gs://kubernetes-release/release/${KUBERNETES_RELEASE_VERSION}/kubernetes-client-${platform}-${arch}.tar.gz" ${cache_dir}
 
+#  decompress test files
+gunzip ${cache_dir}/kubernetes.tar.gz
```
philoserf: can we not untar and unzip together in our container

`tar -xzf ...` as we were doing?

build-scripts/conformance-tests.sh
```diff
@@ -1,4 +1,5 @@
 set -x
```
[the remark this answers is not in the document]
philoserf: BTW: We keep the `.sh` extension for shell libraries that get sourced into other scripts. Then the extension acts as a queue. This is a library, this is an executable.

.gitlab-ci.yml
leahnp: Great comment, I think I adjusted it how you described.

.gitlab-ci.yml
leahnp: Done - thanks!

leahnp: Agreed.

build-scripts/conformance-tests.sh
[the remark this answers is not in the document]
leahnp: Great information, all points addressed.

build-scripts/conformance-tests.sh
leahnp: It puts too much load on memory - I agree together would be cleaner though.

build-scripts/conformance-tests.sh
```diff
@@ -28,11 +34,12 @@ mkdir -p "${GOPATH}"
 
 ## run
 K2_CLUSTER_NAME=`echo $2 | tr -cd '[[:alnum:]]-' | tr '[:upper:]' '[:lower:]'`
-export KUBE_CONFORMANCE_KUBECONFIG=${PWD}/cluster/aws/${K2_CLUSTER_NAME}/admin.kubeconfig
+# export KUBE_CONFORMANCE_KUBECONFIG=${PWD}/cluster/aws/${K2_CLUSTER_NAME}/admin.kubeconfig
```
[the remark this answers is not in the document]
leahnp: I understand the point of this comment, but I don't know how to address it.

philoserf:
Do you want to squash some of this together of let GitHub squash it?

joejulian:
Our policy is to do squash merges here.

build-scripts/conformance-tests.sh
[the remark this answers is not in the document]
joejulian: Delete commented code.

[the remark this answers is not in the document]
leahnp: Got it - thanks

.gitlab-ci.yml
```diff
@@ -0,0 +1,233 @@
+---
+# This pipeline requires 7 variables to be set in the CI/CD Settings
+# hint for creating the base64 encoded content: `cat .ssh/id_rsa|base64|pbcopy`
+
+# SSH_KEY - base64 encoded ssh key (think ~/.ssh/id_rsa)
+# SSH_PUBLIC_KEY - base64 encoded ssh public key (think ~/.ssh/id_rsa.pub)
```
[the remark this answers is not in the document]
philoserf: They can be copied from one gitlab config to another.  I included the generation method in a comment in mind for our future selves if/when we add more base64 ENVs.

.gitlab-ci.yml
```diff
@@ -0,0 +1,233 @@
...
+    paths:
+      - $CI_PROJECT_DIR/cluster/gke/**
+    expire_in: 2 hour
+  script:
+    - mkdir -p /root/.ssh /root/.config/gcloud/
+    - echo $SSH_KEY | base64 -d > /root/.ssh/id_rsa
+    - chmod og-wrx /root/.ssh/id_rsa
+    - echo $SSH_PUBLIC_KEY | base64 -d > /root/.ssh/id_rsa.pub
+    - echo $GKE_CONFIG | base64 -d > /root/.config/gcloud/patrickRobot.json
+    - ./bin/up.sh --config $CI_PROJECT_DIR/cluster/gke/config.yaml --output $CI_PROJECT_DIR/cluster/gke/ --verbose "-vvv"
```
davidewatson: Since we don't set `allow_failure: True` here, will failures on `up.sh` cause `down.sh` to be skipped and leak AWS resources?

If we do set `allow_failure: True`, then will the job appear to succeed even it failed? Maybe this stage should conditionally call `down.sh` when `up.sh` fails?

.gitlab-ci.yml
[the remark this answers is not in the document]
joejulian: But ssh keys can be created (`ssh-keygen`), used and discarded. I don't think we need any long-lived ssh keys for CI (do we?).

.gitlab-ci.yml
joejulian: No. The `when: always` will make the cleanup jobs run even if the test stage fails.

.gitlab-ci.yml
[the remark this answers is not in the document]
leahnp: Is there an action item on this?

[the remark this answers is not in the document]
joejulian: Either defend putting long-lived ssh keys in gitlab variables or remove them from variables and instead create them as part of the CI process.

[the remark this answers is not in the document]
leahnp: I don't have enough context to feel strongly about one or the other - this model was presented by @venezia - Mike do you have an opinion?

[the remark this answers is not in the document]
leahnp: It appears that this method is what gitlab suggests: https://docs.gitlab.com/ee/ci/ssh_keys/README.html#how-it-works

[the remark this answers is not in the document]
joejulian: Sure, but that's assuming interaction with a pre-existing machine. Since the duration of the need for this key does not extend beyond the lifecycle of the CI run, there's no need for a static key.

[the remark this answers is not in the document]
leahnp: What is the down side of a long-lived ssh key for this use case?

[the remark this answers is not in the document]
leahnp: Been reading/watching vids on the subject, I get it now - will use ssh-keygen. A few other ci pipelines use the above method of generating keys, should we make tickets to update those pipelines?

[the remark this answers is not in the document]
venezia: The current kraken system assumes that there is a key that you want to use in conjunction with the system.  Kraken does not support generating a key for you, etc.

Earlier, in Jenkins, a ssh key was provided through S3 to use for all of this.  This environmental variable stuff was done to bring parity with that process.

If you want the build system to build a random key and use it, OK.  That's an extra step and needs to be passed in through artifacts for the job.

joejulian:
This claims that "Some checks haven't completed yet" but I cannot find any that are running that would cause that, and no artifacts in aws that would support that claim. I'm going merge this based on the success in gitlab.

PR #998, opened 2017-12-05
Fix e2e failure on udp communication from node to pod

joejulian:
A kernel bug introduced somewhere in the coreos 1520 releases prevents udp communication from a node to a pod due to a bad checksum (which is expected) being checked for validation (which is unexpected) on a veth interface. CoreOs release 1520.9.0 does not exhibit this failure.


joejulian:
Do not merge this PR unless e2e passes.

coffeepac:
resolved by #1057

PR #1009, opened 2017-12-13
Add certified kubernetes badges

spiffxp:
We're mentioned as certified here: https://www.cncf.io/announcement/2017/11/13/cloud-native-computing-foundation-launches-certified-kubernetes-program-32-conformant-distributions-platforms/

Add the badges we have earned, and link to the instructions people can use to reproduce the conformance results on their own with kraken

ref: samsung-cnct/kraken#223


davidewatson:
CI failed due to etcd discovery failure. Restarting.

philoserf:
Should we
Can we
Drop the label on the Kraken repo README too?
It is my face of the Kraken.

PR #1011, opened 2017-12-14
Drain before delete

joejulian:
Drain nodes before deleting them to allow kubernetes to honor PodDisruptionBudgets


leahnp:
What is the current use case for this functionality? Will it interfere with cluster-autoscaling in clusters where that is enabled? Cluster autoscaling already drains nodes and respects PodDisruptionBudgets to my knowledge.

ansible/library/kubernetes_node_wait.py
```diff
@@ -0,0 +1,206 @@
+#!/usr/bin/python
+
+from datetime import timedelta, datetime
+from ansible.module_utils.basic import AnsibleModule
+
+try:
```
leahnp: Is this just for best practices - or is there a real concern the k8s library would fail to import?

ansible/library/kubernetes_node_wait.py
```diff
@@ -0,0 +1,206 @@
...
+except ImportError as err:
+    HAS_LIB = False
+
+ANSIBLE_METADATA = {
+    'metadata_version': '1.1',
+    'status': 'preview',
+    'supported_by': 'community'
+}
+
+DOCUMENTATION = '''
```
leahnp: Should we mentioned this documentation in the primary README? Or link it somehow.

ansible/library/kubernetes_node_wait.py
```diff
@@ -0,0 +1,206 @@
...
+            label_selector=",".join(module.params['labels']),
+            _request_timeout=convert_from_timedelta(stop_at - datetime.utcnow())):
+        if is_ready(event['object'], module.params['state']
+                   ) and event['object'].metadata.name not in result['nodes']:
+            if event['type'] == 'ADDED':
+                result['initial_nodes'] += 1
+            result['nodes'].append(event['object'].metadata.name)
+        if len(result['nodes']) >= module.params['count']:
+            watchobj.stop()
+        if datetime.utcnow() > stop_at:
```
leahnp: To be sure I understand this - if it times out it stops trying to take any action on the nodes and leaves the cluster in the same state?

ansible/library/kubernetes_node_wait.py
joejulian: Best practice. We don't have to worry in our container because we control the environment, but if you were to run ansible on your laptop, for instance, and didn't have the kubernetes python library installed, we need to fail.

This is also a requirement for getting this module contributed into ansible, something I plan to work on.

ansible/library/kubernetes_node_wait.py
joejulian: This is used upstream for generating the online docs pages.

ansible/library/kubernetes_node_wait.py
joejulian: If it times out, it does fail. The actions taken when the module fails are up to the role, but in this implementation, yes. If it exceeds 10 minutes something's wrong and the playbook fails.

ansible/library/kubernetes_node_wait.py
leahnp: Cool

joejulian:
> What is the current use case for this functionality? 

This is simply a reimplementation of the filter that was doing the same for a whole group of nodes so the only change in the previous update process is the addition of the drain and delete of the node from kubernetes before deleting the node from aws.

> Will it interfere with cluster-autoscaling in clusters where that is enabled? Cluster autoscaling already drains nodes and respects PodDisruptionBudgets to my knowledge.

I don't think this has been looked at wrt update yet and is out of scope for this PR.

ansible/library/kubernetes_node_wait.py
leahnp: I see - should we mention this functionality in the kraken-lib README for our users?

joejulian: Not in my opinion. This is not something a user will interface with, and a developer can look at the source code documentation.

ansible/library/kubernetes_node_wait.py
davidewatson: Don't you need to call `watchobj.stop()` in this case as well?

joejulian: No because this returns and deletes the object, closing the socket as well.

PR #1014, opened 2017-12-15
Not sure if this was a typo or not, but possibly should be 'tagged' i…

kmansel:
…nstead of 'targed'?  If not, then I guess i'm not familiar with the term and disregard.


samba:
@kmansel Thanks for raising this. The original intent was actually `targeted`.

kmansel:
@samba this has been updated and ready to go . 👍

samba:
Looks good to me. 🚀

PR #1015, opened 2017-12-15
Use kubelet to run etcd 

davidewatson:


davidewatson:
Since this is a significant change to how kraken deploys etcd, I was going to switch from using service discovery to using static IP addresses. We've wanted to do this or while as it would make etcd cluster formation more reliable. Unfortunately, this might not be so easy. To create a static pod manifest with static IP addresses we need to template each etcd node separately. `ansible/cloud_config/write_files.yaml` makes this difficult because it assumes that templates are per nodepool.

philoserf:
can we test the impact of the change on upgrade, please?

alejandroEsc:
@davidewatson are you suggesting here that not only will we move over to have kubelet manage etcd (for which then i would ask why dont we use operator also or perhaps cyklps-etcd) but that we also use static ips? I thought these were mutually exclusive concepts, are we determining that we can/should use static ips now?

alejandroEsc:
@philoserf include also the recovery story, as this will become important also.

davidewatson:
@philoserf: We will. Note that kraken created etcd clusters are not changed by kraken after successful creation. Upgrades are a manual process, so I think we only need to test that old kraken configs can still be used to spin up clusters using the existing etcd configuration.

@alejandroEsc: Good point about the recovery playbook needing to be updated. Static pods and static IP addresses are orthogonal concepts. You can use both at the same time if you are able to generate the manifest.

One reason for not using `etcd-operator` is that it requires self-hosted etcd [1] and we believe circular dependencies are problematic. Perhaps more importantly, due to bugs, I've been told that self-hosted etcd is not highly available right now.

[1] This is partially false. If the self-hosted etcd were in a different k8s cluster then we might be okay.

davidewatson:
This was an alternative fix for https://github.com/samsung-cnct/kraken-lib/issues/825, in addition to https://github.com/samsung-cnct/kraken-lib/pull/974

PR #1023, opened 2017-12-20
kubeNodeTaint is an object in the schema. Fixes colonybead-state/1.

davidewatson:


DStorck:
kubeNodeTaint is also referenced here kraken-lib/schemas/config/v1/schedulingConfig.json, should you also change the type there?

joejulian:
No. In those places we're saying that the `taints` key is an _array_ of taint objects.

DStorck:
Gotcha, thanks.

PR #1031, opened 2017-12-21
update calico versions

coffeepac:
patch release updates for calicoCni and calicoNode

part of samsung-cnct/issues#97


leahnp:
@coffeepac I think this needs a rebase - can we merge if E2E fails?

coffeepac:
happy to rebase it, will do shortly.

if the e2e test is the UDP test case then yes, it can be merged.  if other e2e test failures then no.

PR #1033, opened 2017-12-22
Channel option added for helm registries

DStorck:


Documentation/kraken-configs/helmconfigs.md
```diff
@@ -57,6 +56,10 @@ helmConfigs:
         chart: samsung_cnct/heapster
         version: 0.1.0
         namespace: kube-system
+      - name: curator
+        registry: quay.io
+        chart: samsung_cnct/curator
+        channel: stable 
       - name: central-logging
```
davidewatson: Is this still a (working) thing?

ansible/roles/kraken.services/tasks/run-services.yaml
```diff
@@ -197,7 +197,24 @@
...
+      --values {{ config_base }}/{{ cluster.name }}/{{ item.registry }}-{{ item.name | replace('/','_') }}.helmvalues
+  environment:
+    KUBECONFIG: "{{ kubeconfig }}"
+    HELM_HOME: "{{ helm_home }}"
+    PATH: "{{ helm_path }}:{{ ansible_env.PATH }}"
+  with_items: "{{ cluster_services }}"
+  when:
+    - cluster_services is defined
+    - helm_command is defined
+    - item.registry is defined and item.channel is defined
```
davidewatson: Shouldn't we be able to specify a `channel` and a `version`?

Documentation/kraken-configs/helmconfigs.md
DStorck: curator or central-logging?

ansible/roles/kraken.services/tasks/run-services.yaml
DStorck: so if a user specified both, which would you want to win?  a channel has a version associated with it, not necessarily the same version the user is specifiying.  so it could get weird.

Documentation/kraken-configs/helmconfigs.md
davidewatson: `central-logging`. Afaik, only `logging` has been ported over to GitLab...

DStorck: that can probably get ripped out as an example.   I hadn't been looking at that, i just wanted to add in an example of using channel.

DStorck: actually atlas is still alive and well in our main config.yaml, we should think about removing that as well

DStorck:
Don't merge yet, I'm going to clean up atlas references as well.

DStorck:
Okay, all set. Took out atlas references from configs and documentation.

ansible/roles/kraken.services/tasks/run-services.yaml
davidewatson: You are right. `quay.io` does not support channels and versions at the same time. That's a shame.

PR #1037, opened 2017-12-28
First step for cleaning up a_cluster vars

gochist:
This is a first step for https://github.com/samsung-cnct/kraken-lib/issues/874

### DIFF
--- a/ansible/roles/kraken.access/tasks/main.yaml
+++ b/ansible/roles/kraken.access/tasks/main.yaml
@@ -3,5 +3,5 @@
   with_items:
     - "{{ kraken_config.clusters }}"
   loop_control:
-    loop_var: a_cluster
-  when: a_cluster.providerConfig.type == 'cloudinit'
+    loop_var: cluster
+  when: cluster.providerConfig.type == 'cloudinit'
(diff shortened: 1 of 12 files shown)


coffeepac:
I haven't checked this out to try but the original reason for using 'a_cluster' instead of 'cluster' was a warning that gets thrown from the ansible about a possible variable occlusion.  ansible lets variables into the global namespace easily so I added the layer of indirection to make sure we were getting the expected value for `cluster`.

are you seeing this warning?  It occurs during the start of the run.

coffeepac:
this is from the CI run for this PR:

`TASK [roles/kraken.access : include] *******************************************
 [WARNING]: The loop variable 'cluster' is already in use. You should set the
`loop_var` value in the `loop_control` option for the task to something else to
avoid variable collisions and unexpected behavior.
included: /common-tools/samsung-cnct_kraken-lib/ansible/roles/kraken.access/tasks/setup-certs.yaml for localhost`

That warning makes me concerned we are going to be getting unexpected behavior.  Maybe it doesn't matter!

gochist:
Oh, I just passed by the warning messages.

PR #1041, opened 2018-01-03
Adds cleanup script for leaked resources to CI [BAC-68]

guineveresaenger:
This is a patch fix for cleaning up leaked kraken resources while we look into the terraform issues that may be causing these leaks.


.gitlab-ci.yml
```diff
@@ -149,6 +150,21 @@ cleanup:cloud-gke:
     - echo $GKE_CONFIG | base64 -d > /root/.config/gcloud/patrickRobot.json
     - ./bin/down.sh --config $CI_PROJECT_DIR/cluster/gke/config.yaml --output $CI_PROJECT_DIR/cluster/gke/ --verbose "-vvv"
 
+#cleanup of possible leaked resources 
+cleanup:teardown-aws:
+  stage: cleanup
+  image: $KRAKEN_TOOLS
+  when: always
+  script:
+    - ./hack/kraken-teardown-aws.sh -c $CI_PROJECT_DIR/cluster/aws/config.yaml | bash
```
leahnp: I thought I was told be someone we were not using the `.sh` on scripts as a team best practice

guineveresaenger: I followed the practice in the ci file - see also line 151
but I'd like to learn more about this?

joejulian: https://google.github.io/styleguide/shell.xml#File_Extensions

guineveresaenger:
Huh yikes. Looking at the stages, it appears I don't actually want the teardown script to run in parallel with the regular cleanup script. They should be in their own stage. Gah.
@leahnp @samba

.gitlab-ci.yml
samba: Yes, the style guide advises to *not* use the shell extension for executables. 

The use-case of this particular script is *a little* unique though. 
Consistent with my usual preference for explicit declaration, I'd advocate for a slightly varied call...

```
bash ./hack/kraken-teardown-aws.sh ... | bash
```

samba: Bigger problem though:

- That argument to the script requires the _cluster name_, not the config file. 

In this case, cluster names look something like `krakenlib-ci-1234`. (Based on the pipeline number.)

.gitlab-ci.yml
```diff
@@ -149,6 +150,22 @@ cleanup:cloud-gke:
...
+
+cleanup:teardown-gke:
+  stage: cleanup
+  image: $KRAKEN_TOOLS
+  when: always
+  script:
+    - ./hack/kraken-teardown-gke.sh -c $CI_PROJECT_DIR/cluster/gke/config.yaml | bash
```
samba: Please review the required arguments for this script too; GCP requires the zone and project ID specified. Also the `-c` argument here relates to the _deployment_ name, not the cluster name.

guineveresaenger:
Please do not merge, as the teardown scripts are not behaving properly yet.

PR #1046, opened 2018-01-06
always save artifacts from 'cluster up'

coffeepac:
currently we only save the state from 'up' commands when the command succeeds.  however, if the etcd cluster fails to form, the cluster up fails.  

we still need the artifact here so that the down command will work.


davidewatson:
This passed, but the GitHub status was never updated from pending.

PR #1047, opened 2018-01-08
add a space to the CLI args to make them work again

coffeepac:
At some point during the CLI arg refactoring, someone removed a
trailing space that this script was reyling to function.  I have
added it here as a leading space so that I don't have to test
anything else

currently, this means that `kraken tool` doesn't work

### DIFF
--- a/bin/max_k8s_version.sh
+++ b/bin/max_k8s_version.sh
@@ -23,7 +23,7 @@ source "${my_dir}/../lib/kraken_arguments.sh"
 
 # setup a sigint trap
 trap control_c SIGINT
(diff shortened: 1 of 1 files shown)


davidewatson:
This bug was introduced here: https://github.com/samsung-cnct/kraken-lib/commit/fa64ecb8c59d4dc5abd57edc70cbeb833f8700d4

But the real problem is that we should not depend on the space from the initial checkin: https://github.com/samsung-cnct/kraken-lib/commit/3b986c780be560eea6e550bf9a78186fe3394ab3

PR #1051, opened 2018-01-10
[BAC-105] - add support for multiple versions of k8s in CI

coffeepac:
explicitly add stanzas that only run against master for minus one
and minus two where the minus is coming off of the minor version.


coffeepac:
closing this.  will re-open with a different branch name so we have some way of triggering all versions from a PR.  requires some branch naming conventioning.

PR #1053, opened 2018-01-12
[BAC-147] fix kraken-lib release process

coffeepac:
need to sort out why the command to re-write the dockerfile is failing.

EDIT:  the problem was that the dockerfile re-writing script was attempting to use `/bin/bash` which isn't present in docker image `docker:latest`.  We don't need anything in bash so I switched the shebang to sh.  script works now.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -142,8 +142,8 @@ test:validate-config:
   script:
     - echo "When this job fails, something's wrong in your CI config."
     - test -f /root/.ssh/id_rsa || exit $?
-    - test -f $CI_PROJECT_DIR/cluster/gke/config.yaml || exit $?
-    - test -f $CI_PROJECT_DIR/cluster/aws/config.yaml || exit $?
+    - test -f $CI_PROJECT_DIR/cluster/gke-$KUBERNETES_E2E_TEST_VERSION/config.yaml || exit $?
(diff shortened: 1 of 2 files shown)


coffeepac:
ready for merge

philoserf:
as `#!/bin/bash` comes from our bash style guide. Perhaps variance from the style guide in a script should contain a justifying comment. That way our future selves will have the required context we built in this PR commentary.

samba:
Agree w/ @philoserf ; however in CI containers, I'd also approve `apk add bash` as needed.

coffeepac:
This isnt a container we build, we are using upstream docker.  

I can add a comment.

samba:
Sorry, I meant to add that `apk add bash` command as part of the CI script, since we control which container it's running in, and that installation is pretty lightweight.

coffeepac:
I would rather not mutate the container we are running in right now.  This is a break-fix and what we have in this PR right now works.  I would rather not sort out what changes may happen.

I can file an issue to handle making that change for later.

coffeepac:
@samba or @philoserf I believe I have addressed all concerns and the latest commit has passed CI.  Can I get this merged?

samba:
LGTM. Merging, even though I'm not actually listed as a reviewer etc...

PR #1059, opened 2018-03-21
add system log mount to master kubelets

coffeepac:
the kubelet is responsible for making symlinks from where docker stores
log files (/var/lib/docker/) to where kubernetes says the log file
will be available (/var/lib/pods or /var/lib/containers).  to let
kubelet make these changes on the host, we need to mount the hosts
/var/log or we're just farting around in the kubelet's running
imagefs


coffeepac:
there is something busted with the 'minus-one' CI job.  the ci job is repeatedly making several (I think 4) vpcs.

PR #1060, opened 2018-03-28
Add untested support for cluster autoscaler with newer k8s versions.

davidewatson:


davidewatson:
https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler#releases

and

https://github.com/kubernetes/autoscaler/releases

were used as input.

